PR #9, opened 2024-02-19
Add blurb at the bottom 

wainwmr:
Add some text at bottom of player
Use latest spem.json file


wainwmr (commented):
Using autoprefixer
Changed play/pause button to be span instead of button
Detect whether running on SV automatically.

PR #53, opened 2026-04-30
[fix] Fix event listener leak causing UI hang on mobile

wainwright1000:
Extracted touchmove/touchend handlers to stable instance properties to prevent exponential event listener accumulation on rapid mobile taps.


wainwright1000:
Closing — this fix is already present on the dev branch (commit adbd1a2).

PR #186, opened 2026-05-03
feat: dark mode by default, title font update, and workflow docs

wainwmr:
## Changes

- **#90** — Better font/graphic for Spem Player title
- Use dark mode by default (removes OS-level colour-scheme detection)
- Workflow docs update (PR #185)
- Prettier compliance fix


wainwmr (commented):
All good

PR #206, opened 2026-05-04
fix: reject bars 140 and 141 in setBar

### LINKED ISSUE #177: setBar accepts invalid bars 140 and 141
## Description

When navigating to the final bar of the piece and pressing the Right Arrow key, the bar indicator in the controls input shows 140 instead of wrapping back to 0. The score scrolls past its right edge and the bar highlight displays at an incorrect position. The same occurs if the user manually types 140 into the bar input or presses Ctrl+Right Arrow from bar 139.

## Steps to Reproduce

1. Load the Spem Player page.
2. Navigate to bar 139 (the final bar) using the bar input or arrow keys.
3. Press the Right Arrow key once more.
4. Observe the bar input field and the score position.

## Expected result

The bar input should show 0 (wrapping to the beginning of the piece), the score should scroll to the start, and the highlight should indicate bar 0.

## Actual result

The bar input shows 140, the score scrolls beyond the final bar boundary, and the highlight is positioned off the score.

## Current State

- **File:** `index.ts`
- **Function:** `setBar(b)` (line 122)
- **Current guard:** `if (b > 141)` (line 124)

```typescript
function setBar(b: number) {
  b = toNum(b, false);
  if (b > 141) {
    controls.pause();
    b = 0;
  }
  else if (b < 0) {
    b = 139;
  }
  current.bar = b;
  controls.setAttribute("bar", String(b));
  score.setAttribute("bar", String(b));
  canvas.setAttribute("bar", String(b));
}
```

Because the upper bound check is `b > 141`, values 140 and 141 pass through unchanged. These values exceed the valid range (0-139) and are passed to `music-score`, `music-controls`, and `music-canvas` attributes.

## Root Cause

The upper bound guard uses the wrong comparison operator and the wrong threshold. It should reject any bar `>= 140`, not only `> 141`.

## Scope

- **In scope:** fix the upper bound guard in `setBar` so that bars 140 and above wrap to 0, consistent with the intended clamping behaviour.
- **Out of scope:** refactoring the bar input widgets or changing the wrap-to-0 behaviour for other boundary cases.

## Recommended Fix

Change line 124 of `index.ts` from:

```typescript
if (b > 141) {
```

to:

```typescript
if (b >= 140) {
```

This ensures any bar at or beyond the maximum valid value (139) wraps to 0, matching the existing lower-bound wrap (`b < 0` wraps to 139).

## Test Plan

The following probe tests confirm the failure:

```typescript
it("setBar should reject bar 140 and wrap to 0", () => {
  // dispatch music-controls-changed with bar 140
  // assert score.getAttribute("bar") === "0"
});

it("ArrowRight from bar 139 should wrap to bar 0, not stop at 140", () => {
  // set bar to 139, simulate ArrowRight
  // assert score.getAttribute("bar") === "0"
});
```

Both currently fail: the score attribute is `"140"` instead of `"0"`.

After the fix, these tests should pass. Also run the full test suite (`npx vitest run`) to ensure no regressions in bar navigation.

## Dependencies

- None.

## Open Questions

- None.

wainwright1000:
Fixes #177

### DIFF
--- a/e2e/bar-navigation.spec.ts
+++ b/e2e/bar-navigation.spec.ts
@@ -1,0 +1,14 @@
+import { test, expect } from "@playwright/test";
+
+test.describe("Bar navigation boundary wrapping", () => {
+  test("ArrowRight from bar 139 wraps to bar 0", async ({ page }) => {
+    await page.goto("/?bar=139");
+
+    const controls = page.locator("music-controls");
+    await expect(controls).toHaveAttribute("bar", "139");
+
+    await page.keyboard.press("ArrowRight");
+
+    await expect(controls).toHaveAttribute("bar", "0");
+  });
+});
--- a/index.ts
+++ b/index.ts
@@ -1,6 +1,6 @@
-import './src/scss/style.scss';
+import "./src/scss/style.scss";
 
-import config from "./src/ts/config"
+import config from "./src/ts/config";
 
 import { PartType, State, colors, toNum, HDSQTIME } from "./src/ts/common";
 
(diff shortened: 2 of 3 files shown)


wainwmr (approved):
All good

PR #209, opened 2026-05-04
fix: Harden build scripts (#191/#192)

### LINKED ISSUE #191: Build scripts have CRLF line endings and no error handling
## Problem

`build/buildScore.sh` and `build/buildAllScores.sh` were introduced in ticket #187 (2026-05-03). Both files have the following defects:

1. **CRLF line endings.** Both files contain Windows-style `\r\n` line endings. On Unix, a shell invoked via shebang (`#!/bin/bash`) reads `bash\r` as the interpreter path and fails with "bad interpreter: No such file or directory." The `package.json` script runs `bash ./build/buildAllScores.sh`, which bypasses the shebang and may tolerate CRLF on some Bash versions, but this is not guaranteed.

2. **No error handling.** Neither script uses `set -e`, `set -u`, or `set -o pipefail`. If `lilypond` fails or the output directory is missing, `buildScore.sh` continues to the `sed` step. If one file in the loop fails, `buildAllScores.sh` continues silently to the next file.

3. **No argument validation.** `buildScore.sh` accepts a single argument (the `.ly` file path) but does not verify that the argument is non-empty or that the file exists.

4. **No dependency checks.** Neither script verifies that `lilypond` is installed or on PATH.

## Recommended fix

- Convert both files to LF-only line endings (add `*.sh text eol=lf` to `.gitattributes` if needed).
- Add trailing newlines.
- Add `set -euo pipefail` to both scripts.
- Validate `$1` in `buildScore.sh`.
- Verify `lilypond` is available before invoking.

## Context

- Created as part of ticket #187.
- Platform: macOS (based on `sed -i ''` syntax), but the CRLF endings suggest the files were created on Windows.
- Files are referenced in `package.json` as `npm run build:scores`.

## Test plan

1. Run `bash build/buildAllScores.sh` on a Unix system and verify it does not fail with "bad interpreter."
2. Run `build/buildScore.sh` with no arguments and verify it exits with a usage message.
3. Run `build/buildScore.sh` with a non-existent file and verify it exits with an error.
4. Temporarily rename `lilypond` and verify the script fails early with a clear message.

### LINKED ISSUE #192: Build script sed command uses macOS-specific syntax
## Problem

`build/buildScore.sh` uses macOS-specific `sed` syntax:

```bash
sed -i '' -E '1,1s/ height="[0-9.]+[a-zA-Z]*"//g; 1,1s/ width="[0-9.]+[a-zA-Z]*"//g' "$svg"
```

On Linux, `sed -i ''` creates a backup file literally named `''` instead of editing in place. On Windows (Git Bash), the behaviour depends on the `sed` implementation. This makes the build script non-portable.

## Recommended fix

Replace the macOS-specific `sed -i ''` with a cross-platform pattern. Options:

1. **Portable in-place edit:**
   ```bash
   sed -i.bak -E '1s/ height="[0-9.]+[a-zA-Z]*"//g; 1s/ width="[0-9.]+[a-zA-Z]*"//g' "$svg" && rm -f "$svg.bak"
   ```

2. **Redirect to temp file:**
   ```bash
   sed -E '1s/ height="[0-9.]+[a-zA-Z]*"//g; 1s/ width="[0-9.]+[a-zA-Z]*"//g' "$svg" > "$svg.tmp" && mv "$svg.tmp" "$svg"
   ```

Either approach works on macOS and Linux.

## Context

- Introduced in ticket #187 (2026-05-03).
- The `sed` step strips `height` and `width` attributes from the first line of generated SVGs so that the browser scales them correctly.
- The post-processing has already been applied to the committed SVGs in `src/scores/`, so this fix only affects future builds.

## Test plan

1. Run the build script on macOS and verify SVGs are generated without `height`/`width` on the first line.
2. Run the build script on Linux and verify the same result (no `''` backup file created).
3. Verify the generated SVG still displays correctly in the application.

wainwright1000:
- Add set -euo pipefail to both scripts
- Validate arguments and check lilypond dependency in buildScore.sh
- Replace macOS sed -i '' with portable temp-file redirect
- Add .gitattributes to enforce LF line endings for *.sh

Related to #191, #192


src/test/build-scripts.test.ts
wainwmr, line 1: I don't like this file.  Seems very strange/wrong to have a test that checks that a build file has the right kind of content.  Tests should check outcomes, not look inside code!  Suggest we just delete this one.

wainwmr (changes requested):

wainwright1000:
I understand.  I don't know enough to present the other side (and I'm not going to start using Bob against you).

I think the root cause here is that I ran the bug hunt against, well pretty much everything.  And I'm not sure that's right.  Could you have a look at File Mappings, which is where we define which files/folders get scanned for bugs and tech debt?  Maybe what we do is not look for bugs in "Configuration and Build Tools".  While you're there, please have a look at that list.  Is there anything not covered by the bug hunt mechanism that should be?

Let me know, and I'll change that wiki entry (which is authoritative) and baby sit the process that runs as a result.

wainwright1000:
Accepted your feedback on the build-scripts test file.

I have removed `src/test/build-scripts.test.ts`. You are right — the tests were checking file content (presence of `set -euo pipefail`, absence of `\r\n`, argument validation patterns) rather than observable outcomes. That is linting, not testing, and belongs in tools like shellcheck or `.editorconfig` rather than in Vitest.

All other changes from the original PR remain intact:
- `set -euo pipefail` added to both build scripts
- Argument validation and lilypond dependency check in `buildScore.sh`
- Portable sed replacement (temp file instead of macOS `-i ''`)
- `.gitattributes` enforcing LF line endings for shell scripts

Kimi.

wainwmr (approved):
All good

PR #226, opened 2026-05-08
fix: Prevent duplicate rAF loop and reset spinner on autoplay block (#179, #180)

### LINKED ISSUE #179: Changing choir during playback doubles bar update rate
## Description

Changing choir, part, or recording while audio is playing causes the playback loop to run twice, doubling the rate at which bar updates and `music-controls-changed` events fire. The bar input field flickers and the canvas receives twice as many position updates as it should.

## Steps to Reproduce

1. Load the Spem Player page.
2. Press Space or click the play button to start playback.
3. While audio is playing, press a digit key (1-8) to change choir, or press V to toggle recording.
4. Observe the bar input field and canvas position indicator.

## Expected result

Bar updates should continue at the same rate as before the change. The canvas and score should receive one position update per animation frame.

## Actual result

Bar updates fire at roughly double the previous rate. The canvas position indicator may appear jerky because two concurrent loops are competing to set the bar value.

## Current State

- **File:** `src/ts/MusicControls.ts`
- **Function:** `play()` (line 170)

`play()` starts a `requestAnimationFrame` loop on every call without checking whether one is already running:

```typescript
function loop() {
  self.bar = getBarFromTime(self.audio.currentTime, self.recording);
  // ... update barinput ...
  self.fireEvent("music-controls-changed");
  if (self.isPlaying()) {
    window.requestAnimationFrame(loop);
  }
}
window.requestAnimationFrame(loop);
```

`setChoir()`, `setRecording()`, and `setPart()` all call `this.play()` when `isPlaying()` is true (lines 241, 247, 256).

## Root Cause

No guard prevents `play()` from starting a new loop while another is active. The loop terminates only when `isPlaying()` becomes false, but a second call to `play()` creates a second loop before the first one reaches its termination check.

## Scope

- **In scope:** add a guard in `play()` to prevent duplicate rAF loops, or track and cancel the existing frame ID before starting a new one.
- **Out of scope:** refactoring the loop logic or changing how events are dispatched.

## Recommended Fix

Track the animation frame ID in a private field and return early if a loop is already active:

```typescript
#animationFrameId: number | null = null;

async play() {
  // ... existing loading logic ...
  if (this.#animationFrameId !== null) {
    return; // loop already running
  }
  await this.audio.play();
  // ... state updates ...
  const loop = () => {
    this.bar = getBarFromTime(this.audio.currentTime, this.recording);
    // ...
    if (this.isPlaying()) {
      this.#animationFrameId = window.requestAnimationFrame(loop);
    } else {
      this.#animationFrameId = null;
    }
  };
  this.#animationFrameId = window.requestAnimationFrame(loop);
}
```

## Test Plan

Probe test confirming the failure:

```typescript
it("changing choir during playback should not double-fire change events", async () => {
  const elem = document.querySelector("music-controls") as MusicControls;
  elem.audio.play = vi.fn(() => Promise.resolve());
  elem.audio.pause = vi.fn();

  let eventCount = 0;
  elem.addEventListener("music-controls-changed", () => eventCount++);

  await elem.play();
  await new Promise(r => setTimeout(r, 100));
  const countAfterNormal = eventCount;

  eventCount = 0;
  elem.setChoir(1);
  await new Promise(r => setTimeout(r, 100));
  const countAfterChoirChange = eventCount;

  expect(countAfterChoirChange).toBeLessThanOrEqual(countAfterNormal * 1.5);
});
```

Currently fails: `expected 9 to be less than or equal to 7.5`.

## Dependencies

- None.

## Open Questions

- None.

### LINKED ISSUE #180: Play button stuck on loading spinner when autoplay is blocked
## Description

When the browser blocks audio autoplay or the network fails to load the MP3, the play button remains stuck on the loading spinner indefinitely. The user cannot click play again because the loading icon never returns to the play icon.

## Steps to Reproduce

1. Load the Spem Player page in a fresh browser session (no prior interaction).
2. Without clicking anything else first, press Space or click the play button.
3. If the browser blocks autoplay (common in Safari and mobile browsers), observe the play button.

## Expected result

If playback cannot start, the loading spinner should disappear and the play button should return to its normal state so the user can try again.

## Actual result

The loading spinner stays visible indefinitely. The user must refresh the page to recover.

## Current State

- **File:** `src/ts/MusicControls.ts`
- **Function:** `play()` (line 170)

```typescript
async play() {
  const newfile = this.getMP3filename();
  if (!this.isSameAudio(newfile)) {
    // ... set loading spinner visible ...
    this.audio.src = newfile;
    this.audio.load();
  }

  await this.audio.play();  // line 190: unhandled rejection

  this.playing = true;
  // ... set pause icon visible ...
}
```

`await this.audio.play()` can throw (browser autoplay policy, network error, corrupted file). There is no `try/catch`, so the rejection propagates uncaught and the loading state is never reset.

## Root Cause

Missing error handling around `audio.play()`.

## Scope

- **In scope:** wrap `audio.play()` in try/catch and reset the button icons on rejection.
- **Out of scope:** adding a user-visible error message or retry logic.

## Recommended Fix

Wrap `audio.play()` in try/catch:

```typescript
async play() {
  // ... existing loading logic ...

  try {
    await this.audio.play();
  } catch (e) {
    console.error("Audio playback failed:", e);
    this.playing = false;
    if (this.svgPlay && this.svgLoading && this.svgPause) {
      this.svgPlay.style.display = "block";
      this.svgPause.style.display = "none";
      this.svgLoading.style.display = "none";
    }
    return;
  }

  this.playing = true;
  // ... existing state updates ...
}
```

## Test Plan

Probe test confirming the failure:

```typescript
it("play() should handle audio.play() rejection gracefully", async () => {
  const elem = document.querySelector("music-controls") as MusicControls;
  elem.audio.play = vi.fn(() => Promise.reject(new Error("Autoplay blocked")));

  elem.svgLoading!.style.display = "block";
  elem.svgPlay!.style.display = "none";
  elem.svgPause!.style.display = "none";

  await expect(elem.play()).rejects.toThrow("Autoplay blocked");

  expect(elem.svgLoading!.style.display).toBe("none");
  expect(elem.svgPlay!.style.display).toBe("block");
});
```

Currently fails: `expected 'block' to be 'none'` (spinner stays visible).

## Dependencies

- None.

## Open Questions

- None.

wainwright1000:
This PR bundles fixes for both #179 and #180, as both touch MusicControls.ts and controls.test.ts.

- #179: Prevent duplicate rAF loop when changing choir during playback. Adds a #isLooping guard so that calling play() while already playing does not schedule a second animation frame loop.
- #180: Reset loading spinner when udio.play() is rejected (autoplay blocked). Catches the rejection, restores the play icon, and sets playing = false.

See #179
See #180

### DIFF
--- a/src/test/controls.test.ts
+++ b/src/test/controls.test.ts
@@ -448,4 +448,116 @@ describe("MusicControls custom element", () => {
     expect(elem.bar).toBe(0);
     expect(bar.value).toBe("0");
   });
+
+  it("calling play() while already playing does not start a duplicate rAF loop", async () => {
+    const elem = document.querySelector("music-controls") as MusicControls;
+
+    // Capture all requestAnimationFrame callbacks
+    const rafCallbacks: Array<(time: number) => void> = [];
+    const originalRAF = window.requestAnimationFrame;
+    window.requestAnimationFrame = (callback: (time: number) => void) => {
+      rafCallbacks.push(callback);
+      return rafCallbacks.length;
+    };
+
+    // Start playback (schedules the first loop)
+    const waitingForPlay = waitForEvent(
+      elem,
+      "music-controls-playing",
+      handleAudioStarted
+    );
+    elem.setAttribute("playing", "true");
+    await waitingForPlay;
+
+    // Call play() again, as setChoir/setPart/setRecording would
+    await elem.play();
+
+    // Count music-controls-changed events when running captured callbacks
+    let _eventCount = 0;
+    const countListener = () => {
+      _eventCount++;
+    };
+    elem.addEventListener("music-controls-changed", countListener);
+
+    // Run each captured callback once
+    rafCallbacks.forEach((cb) => cb(0));
+
+    // Before the fix: two loops fire, so two events
+    // After the fix: one loop fires, so one event
+    expect(_eventCount).toBe(1);
+
+    // Cleanup
+    elem.removeEventListener("music-controls-changed", countListener);
+    window.requestAnimationFrame = originalRAF;
+  });
+
+  it("pausing after duplicate play() clears all loops without zombies", async () => {
+    const elem = document.querySelector("music-controls") as MusicControls;
+
+    const rafCallbacks: Array<(time: number) => void> = [];
+    const originalRAF = window.requestAnimationFrame;
+    window.requestAnimationFrame = (callback: (time: number) => void) => {
+      rafCallbacks.push(callback);
+      return rafCallbacks.length;
+    };
+
+    // Start playback and duplicate the loop
+    const waitingForPlay = waitForEvent(
+      elem,
+      "music-controls-playing",
+      handleAudioStarted
+    );
+    elem.setAttribute("playing", "true");
+    await waitingForPlay;
+    await elem.play();
+
+    // Pause
+    const waitingForPause = waitForEvent(
+      elem,
+      "music-controls-paused",
+      handleAudioStarted
+    );
+    elem.pause();
+    await waitingForPause;
+
+    // Run all captured callbacks that were scheduled before or during pause
+    const callbacksBeforeRun = rafCallbacks.length;
+    let _eventCount = 0;
+    const countListener = () => {
+      _eventCount++;
+    };
+    elem.addEventListener("music-controls-changed", countListener);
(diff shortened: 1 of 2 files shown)


wainwmr (approved):
All good

PR #230, opened 2026-05-09
fix: correct double-dot duration formula (#229)

### LINKED ISSUE #229: Duration double-dot formula uses incorrect exponent arithmetic
## Summary

The `Duration` class in `src/ts/music-classes.ts` calculates dotted note durations using `1.5 ** n` instead of the correct `2 - 0.5 ** n`. This is a latent defect: the current Tallis source contains no double-dotted notes, so the bug is not triggerable today. However, any future addition of double-dotted notation would produce incorrect audio timing silently.

## Classification Note

This ticket was initially classified as a refactor (no current user impact) during a bug hunt session. Bob (methodology review) changed the classification to bug on the grounds that:
- The formula is unambiguously wrong, not merely suboptimal.
- The risk of silent failure on future notation changes is real.
- The fix is trivial (one line) and the test cost is minimal.

This is a borderline case. We would appreciate Mark's opinion on where the line between latent defect and refactor should sit for future bug hunts.

## Steps to Reproduce

1. Examine `src/ts/music-classes.ts` line 47.
2. Observe `this.sfths = this.sfths * 1.5 ** this.dotted.length`.
3. For a double-dotted note (`n = 2`), this yields `1.5 ** 2 = 2.25`.
4. The correct multiplier for two dots is `2 - 0.5 ** 2 = 1.75`.

## Expected Result

Double-dotted notes should use the formula `2 - 0.5 ** n`.

## Actual Result

The code uses `1.5 ** n`, which overstates the duration.

## Current State

- File: `src/ts/music-classes.ts`, `Duration` constructor, line 47.
- Test file: `src/test/music-classes.test.ts` expects the old (wrong) value.

## Recommended Fix

Change:
```typescript
this.sfths = this.sfths * 1.5 ** this.dotted.length;
```
To:
```typescript
this.sfths = this.sfths * (2 - 0.5 ** this.dotted.length);
```

Add guards:
```typescript
if (dotted != undefined && this.dotted.length > 0) {
```
```typescript
if (multiplier != undefined && multiplier !== 1) {
```

Update test expectation from `36` to `28`.

## Test Plan

- `src/test/music-classes.test.ts`: verify single-dot and double-dot durations.

## Dependencies

None.

wainwright1000:
Fixes #229.

Changes the dotted-note duration calculation from `1.5 ** n` to `2 - 0.5 ** n`.

Also adds guards:
- Skip dot calculation when `dotted.length === 0`
- Skip multiplier application when `multiplier === 1`

Tests updated: double-dot expectation changed from 36 to 28.


wainwmr (approved):
All good

PR #248, opened 2026-05-12
build: enforce LF line endings for source files in .gitattributes

### LINKED ISSUE #247: Enforce LF line endings for TypeScript and SCSS in .gitattributes
Add TypeScript, SCSS, and other source file types to .gitattributes with 	ext eol=lf to ensure consistent line endings across platforms.

**Problem:**
Windows contributors with core.autocrlf=true get CRLF line endings in their working tree for TypeScript, SCSS, JS, JSON, CSS, and HTML files. Prettier expects LF. This causes 
pm run format:check to report false positives on Windows — 25 files appear to fail formatting when the only difference is line endings. Mark's CI (Ubuntu) passes because it has LF line endings.

**Impact:**
- Windows contributors see spurious Prettier failures
- Local ormat:check and CI produce different results
- Risk of Windows contributors submitting PRs with unnecessary formatting noise

**Fix:**
Add *.ts, *.tsx, *.js, *.jsx, *.json, *.scss, *.css, *.html to .gitattributes with 	ext eol=lf, matching the existing *.md and *.sh entries.

**Risk reduction:**
- Eliminates platform-specific formatting discrepancies
- Ensures Prettier behaves identically for all contributors
- Reduces support burden from Windows contributors confused by local failures

wainwright1000:
See #247

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,2 +1,10 @@
 *.sh text eol=lf
 *.md text eol=lf
+*.ts text eol=lf
+*.tsx text eol=lf
+*.js text eol=lf
+*.jsx text eol=lf
+*.json text eol=lf
+*.scss text eol=lf
+*.css text eol=lf
+*.html text eol=lf


wainwmr (approved):
all good

PR #272, opened 2026-05-15
fix: stop animation loops when audio ends naturally (#98)

### LINKED ISSUE #98: loop() never finishes after playing to the end of spem
**Type:** BUG
**Area:** CANVAS
**Status:** specified
**Priority:** P1
**Difficulty:** S

**Source:** `index.ts`:55
**First seen:** 2025-09-11 (e7ec6425)

## Description
The MusicCanvas animation loop (play() -> requestAnimationFrame(loop)) continues firing after audio playback reaches the end of the piece. This suggests the audio ended event either does not set playing=false promptly, or there is a race condition where draw() or another callback re-triggers the loop.

## Recommended fix
Add an `ended` event listener to the Audio element in MusicControls that sets `playing = false`, updates UI icons, and fires `music-controls-paused`. In MusicCanvas, listen for `music-controls-paused` to cancel any pending `requestAnimationFrame`. Store the rAF request ID in both components so it can be cancelled explicitly.

## Test plan
Mock the Audio element, dispatch an `ended` event, and verify that (a) `playing` becomes false, (b) `music-controls-paused` is fired, and (c) no further `requestAnimationFrame` callbacks occur for either MusicControls or MusicCanvas.

## Notes
The fix addresses both the controls loop and the canvas loop. MusicControls currently has no `ended` listener at all.

wainwright1000:
Fixes #98

When audio playback reaches the end of the piece naturally, the `ended` event on the HTMLAudioElement was not being handled. This left `MusicControls.playing` set to `true`, causing both the controls rAF loop and the canvas rAF loop to keep firing indefinitely.

**Changes:**

- `MusicControls.ts`: add `ended` event listener that calls `pause()`, which fires `music-controls-paused` and updates UI state. Store the rAF loop ID (`#loopId`) and explicitly cancel it in `pause()`.
- `MusicCanvas.ts`: store the playback rAF loop ID (`playbackLoopId`) and explicitly cancel it in `setPlaying(false)`.
- `controls.test.ts`: add test that dispatches `ended` on the audio element and verifies `music-controls-paused` is fired, `playing` becomes false, and no further rAF callbacks are scheduled.
- `canvas.test.ts`: add test that verifies `setPlaying(false)` explicitly cancels the pending playback animation frame.

### DIFF
--- a/src/test/canvas.test.ts
+++ b/src/test/canvas.test.ts
@@ -452,4 +452,25 @@ describe("MusicCanvas custom element", () => {
     expect(pos.choir).toBeGreaterThanOrEqual(0);
     expect(pos.choir).toBeLessThan(config.choirs[0].length);
   });
+
+  it("setPlaying(false) cancels pending playback animation frame", () => {
+    expect(canvas).not.toBeNull();
+
+    const originalCancel = window.cancelAnimationFrame;
+    const cancelledIds: number[] = [];
+    window.cancelAnimationFrame = (id: number) => {
+      cancelledIds.push(id);
+    };
+
+    // Start playback
+    canvas!.setPlaying(true);
+
+    // Stop playback
+    canvas!.setPlaying(false);
+
+    // Should have explicitly cancelled the animation frame
+    expect(cancelledIds.length).toBeGreaterThan(0);
+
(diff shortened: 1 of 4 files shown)


wainwmr (approved):
All good

PR #275, opened 2026-05-15
fix: parse fraction denominator in LilyPond parser

### LINKED ISSUE #100: we're ignoring the denominator altogether.  Let's hope it's not there
**Type:** HACK
**Area:** LILY
**Status:** specified
**Priority:** P2
**Difficulty:** XS

**Source:** `src/ts/lily.ts`:123
**First seen:** 2024-04-23 (48d1ecfa)

## Description
The fraction parser in the Ohm grammar only extracts the numerator, ignoring any denominator. This works for current LilyPond input but will silently produce wrong durations if a true fraction like "3/4" appears.

## Recommended fix
Update the Ohm grammar `fraction` rule to capture both numerator and denominator (`fraction = digit+ ("/" digit+)?`). Update the `fraction` semantic action in `lily.ts` to return a ratio object `{num, den}` or a float. Update `Duration` to accept a multiplier ratio instead of a single integer.

## Test plan
Add parser tests for "3/4", "1/2", and plain "2" (no denominator). Assert resulting Duration multiplier is 0.75, 0.5, and 2 respectively.

## Notes
Requires updating both the Ohm grammar (ly-grammar.ohm) and the fraction semantic action in lily.ts.

wainwright1000:
fix: parse fraction denominator in LilyPond parser

The `fraction` semantic action in `lily.ts` was ignoring the denominator, returning only the numerator. This meant durations like `4*3/4` were parsed as `4*3` instead of `4*0.75`.

**Fix:** Use the full matched source string of the fraction rule to parse both numerator and denominator. If a `/` is present, split and divide; otherwise return the integer value.

**Tests added:**
- `parses fraction with denominator 3/4` → 0.75
- `parses fraction with denominator 1/2` → 0.5
- `parses fraction without denominator` → 2
- `parses durationScaled with fraction multiplier` → Duration with sfths = 12

Fixes #100

### DIFF
--- a/src/test/lily.test.ts
+++ b/src/test/lily.test.ts
@@ -164,4 +164,38 @@ describe("lilypond parsing tests", () => {
     expect(frLocations[1].from).toBe(1.0);
     expect(frLocations[1].to).toBe(1.125);
   });
+
+  it("parses fraction with denominator 3/4", () => {
+    const s = setupLilypondParser();
+    const result = lyGrammar.match("3/4", "fraction");
+    expect(result.succeeded()).toBe(true);
+    const value = s(result).parse();
+    expect(value).toBe(0.75);
+  });
+
+  it("parses fraction with denominator 1/2", () => {
+    const s = setupLilypondParser();
+    const result = lyGrammar.match("1/2", "fraction");
+    expect(result.succeeded()).toBe(true);
+    const value = s(result).parse();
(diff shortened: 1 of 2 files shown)


wainwright1000:
Closing this PR to rename the branch from fix/100-fraction-denominator to andrew/100-fraction-denominator, matching the project naming convention. The replacement PR is #282.

PR #278, opened 2026-05-15
fix: cap Vitest workers at 4 and raise test timeout to 10s

### LINKED ISSUE #215: canvas.test.ts and canvaswatcher.test.ts time out when run in parallel
## Specification

**Title**
canvas.test.ts and canvaswatcher.test.ts time out when run in parallel

**Description**
Two Vitest test suites intermittently fail with hook timeouts when executed as part of the full test suite. When run in isolation, both suites pass reliably. This wastes CI time and produces false-negative results that obscure genuine regressions.

**Current state**
`src/test/canvas.test.ts` uses a `beforeAll` hook at line 10 to instantiate `music-canvas`. `src/test/canvaswatcher.test.ts` uses a `beforeEach` hook at line 10 to instantiate `music-canvas-watcher`. Both hooks time out after 10000ms when the full suite runs in parallel.

The failure is non-deterministic. Re-running the suite usually passes.

**Root cause**
Unknown. Suspected causes: shared DOM state pollution between parallel tests, or expensive custom element initialisation in `connectedCallback` that exceeds the default hook timeout under load.

**Scope**
In scope: identifying the root cause and applying a fix (increased timeout, serial execution tag, or DOM isolation).
Out of scope: refactoring the tests themselves or the components under test.

**Recommended fix**
Investigate three hypotheses in order:

1. **Parallel DOM pollution.** Add `vi.clearAllMocks()` and explicit `document.body.innerHTML = ''` cleanup in `afterEach` / `afterAll`. If the DOM retains event listeners or custom elements from prior tests, later tests may hang waiting for stale state.
2. **Serial execution.** Add `describe.sequential` or configure Vitest to run these suites serially if they manipulate shared DOM or global state that is not thread-safe under jsdom.
3. **Hook timeout increase.** Increase the hook timeout from the default 10000ms to 30000ms as a diagnostic step. If this alone fixes the flakiness, the root cause is slow initialisation, not state pollution.

**Test plan**
Run the full Vitest suite ten times consecutively. Assert zero timeouts in `canvas.test.ts` and `canvaswatcher.test.ts` across all runs.

**Dependencies**
None.

**Open questions**
None.

wainwright1000:
## Problem

On machines with 8 logical cores, Vitest spawns 8 fork workers. Every DOM-heavy test file loads the full application via `index.ts` in `beforeAll`, which triggers `processLilypond()`, SVG injection, and Canvas 2D rendering. Eight parallel CPU-bound jsdom processes starve each other, causing non-deterministic timeouts in `canvas.test.ts` and `canvaswatcher.test.ts`.

## Evidence

| Workers | Suite time | Result |
|---|---|---|
| 8 (default) | ~51 s | Timeouts |
| 4 | ~42 s | All 123 tests pass |
| 2 | ~56 s | All 123 tests pass |

Four workers is the sweet spot: faster than 2, reliable where 8 fails.

## Change

Add `maxWorkers: 4` and `testTimeout: 10000` to the Vitest config in `vite.config.ts`.

Closes #215

### DIFF
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -28,6 +28,8 @@ export default defineConfig({
     globals: true,
     environment: "jsdom",
     exclude: ["e2e/**", "node_modules/**"],
+    maxWorkers: 4,
+    testTimeout: 10000,
     coverage: {
       exclude: [
         "**/*.svg",


wainwright1000:
Closing this PR to rename the branch from fix/215-vitest-timeouts to andrew/215-vitest-timeouts, matching the project naming convention. The replacement PR is #281.

PR #285, opened 2026-05-16
build: add typecheck script and align ci with CI workflow

### LINKED ISSUE #277: Align npm run ci with CI workflow by adding tsc --noEmit
## Problem

Vite uses esbuild to transpile TypeScript, which deliberately skips type checking for speed. This means `npm run build` and `npm run ci` do not catch type errors — only the subset of type violations that typescript-eslint rules are written for.

A type mismatch, a property accessed on the wrong type, or a missing interface member can pass CI silently and land in production.

## Proposed fix

Add `tsc --noEmit` as a step in `npm run ci` (and optionally `npm run build`):

```json
"typecheck": "tsc --noEmit",
"ci": "npm run build && npm run typecheck && npm run test && npm run test:e2e"
```

## Why this is worth doing

1. **esbuild does not type-check.** This is by design — it is a known trade-off when using Vite. Without a separate `tsc` step, type safety is only as strong as the ESLint rules that happen to be configured.
2. **The gap is real.** Whole categories of type error (wrong-type arguments, missing properties, interface violations) are invisible to ESLint and therefore invisible to CI.
3. **Cost is low.** `tsc --noEmit` on this codebase is fast. It produces no output files — it is a pure check.
4. **Risk is bounded but real.** This may surface existing type errors in the codebase that must be fixed before CI passes. Those errors are real bugs that currently go undetected.

## Notes

Related to #276 (`@typescript-eslint/no-deprecated`), which addresses a specific lint gap. This ticket addresses the broader type-checking gap.

wainwright1000:
Align the local `npm run ci` script with the GitHub Actions CI workflow by adding a `typecheck` step.

**Changes**
- Add `"typecheck": "tsc --noEmit"` to `package.json`
- Update `"ci"` to run `typecheck` before `build`, matching CI order

Fixes #277

### DIFF
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "build": "npm run format:check && npm run lint && npm run build:ohm && vite build",
     "build:scores": "node build/buildScores.mjs",
     "build:ohm": "npx ohm generateBundles --withTypes src/ohmjs/ly-grammar.ohm",
+    "typecheck": "tsc --noEmit",
     "preview": "vite preview",
     "format": "prettier --write src/",
     "format:check": "prettier --check src/",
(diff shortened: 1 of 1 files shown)


wainwright1000:
Superseded by #291, which combines this PR with #281 and #286 into a single CI health changeset. All three changes are small, low-risk, and unrelated to application code. Reviewing them together is more efficient than three separate one-line PRs.

- Kimi

PR #319, opened 2026-05-21
fix: replace touchmove with touchend on canvas to prevent rapid state updates

### LINKED ISSUE #317: Touch drag on canvas triggers rapid state updates
## User-visible symptom
On a touch device, dragging a finger across the canvas overview causes the selected bar, choir, and part to jump erratically. The audio playback stutters as it seeks repeatedly. The user sees the score and canvas flicker through multiple positions instead of settling on the intended selection.

## Steps to reproduce
1. Open the Spem Player on a touch device (or use Chrome DevTools device emulation).
2. Start playback.
3. Touch the canvas and drag to a different position.
4. Observe the bar counter flickering and the audio stuttering.

## Expected behaviour
A single touch and drag should select one position. The state should update at most once per gesture, not on every touch move event.

## Actual behaviour
The `music-canvas-touchmove` custom event fires on every `touchmove` event during a drag. `index.ts` listens for this event and calls `handleCanvasClick`, which invokes `setChoir`, `setPart`, and `setBar` on every pixel of movement. This causes rapid state updates (potentially 60+ per second), repeated audio seeks, and flickering UI updates.

## Affected files
- `index.ts` lines 539-546
- `src/ts/MusicCanvas.ts` lines 628-632

## Root cause hypothesis
`handleCanvasClick` is bound to `music-canvas-touchmove` with no throttling or debouncing. The intended behaviour was probably to update on `touchstart` only (which is also bound to `handleCanvasClick`). The `touchmove` binding is either redundant or should use a different handler.

## Related
- Workbench item #51 (same root cause)
- #204 (Canvas click and touch return wrong bar on narrow viewports)

wainwright1000:
Fixes #317

Replaces the music-canvas-touchmove listener binding with music-canvas-touchend in index.ts. This eliminates rapid state updates (60+ per second) during touch drag gestures on the canvas overview, which caused audio stuttering and UI flicker on mobile devices.

- Touch drag now updates the canvas visual locally via MusicCanvas.ts internal handlers
- Global state (choir, part, bar) commits once per gesture on touchend instead of on every touchmove pixel
- Touchstart binding preserved for simple tap interactions

Tests: `e2e/touch-drag.spec.ts` (Playwright, chromium with hasTouch) covers tap and multi-point drag (CDP).


wainwright1000:
## Rework

Added two commits in this revision:

- **`da0fc88` test**: `e2e/touch-drag.spec.ts` (Playwright, chromium with `hasTouch`). Tap case + multi-point drag via CDP `Input.dispatchTouchEvent`. The drag asserts `barChanges < STEPS` so any future regression that re-binds `touchmove` fails the test regardless of exact count. TDD-verified locally: reverting `index.ts` to the pre-fix binding produced 23 changes (test failed as expected); restored, both tests pass.
- **`e6e854b` chore**: adds `tests-local/` to eslint ignores. Unrelated to #317; fixes local-build lint failures on clones that have the directory (yours doesn't).

Spec for #317 was [re-specified](https://github.com/wainwmr/spem-player/issues/317#issuecomment-4523322234) with the previously-missing template sections (current state, scope, recommended fix, test plan, acceptance criteria).

- Claude

wainwmr (approved):
All good

PR #328, opened 2026-05-23
build: use platform-aware Python executable in buildScores.mjs

### LINKED ISSUE #302: build: buildScores.mjs uses hardcoded python3, breaking on Windows
### 2. Description

`build/buildScores.mjs` invokes `python3` directly when running `build/postprocessSvg.py`. On Windows, `python3` is not typically on PATH; Python is invoked as `python` or `py`. This causes `npm run build:scores` to fail on Windows with `'python3' is not recognized as an internal or external command`.

### 3. Current state

`build/buildScores.mjs`, line 84 (inside `buildScore` function):

```javascript
execSync(`python3 build/postprocessSvg.py "${svg}"`);
```

The string `python3` is hardcoded. The same issue exists on line 66 in the pre-#271 version of the file.

### 4. Root cause

The script assumes a Unix-like environment where `python3` is the standard Python executable name. Windows installations of Python typically expose `python` (and optionally `py`), not `python3`. The Node.js `child_process.execSync` call fails because the shell cannot resolve `python3`.

### 5. Scope

- **In scope:** Change the Python executable name in `build/buildScores.mjs` to be platform-aware.
- **Out of scope:** Refactoring the `execSync` call to use array arguments instead of string interpolation (tracked by #261). Changing the LilyPond invocation on line 77.

### 6. Recommended fix

Add a platform-aware constant near the top of the file (after the imports):

```javascript
const PYTHON_CMD = process.platform === "win32" ? "python" : "python3";
```

Then replace line 84:

```javascript
execSync(`${PYTHON_CMD} build/postprocessSvg.py "${svg}"`);
```

This is the minimal change. `python` is the standard executable name on Windows (Python.org installer, Microsoft Store, and most CI images all provide it). On Unix, `python3` remains the correct name per PEP 394.

### 7. Test plan

1. **Manual verification (Windows):** On a Windows machine with Python installed, run:
   ```console
   npm run build:scores -- --choir="I A"
   ```
   Expected: build completes without `'python3' is not recognized` error.

2. **Unit test (tracked by #261):** When Vitest tests are added for `buildScores.mjs`, add a test case that mocks `process.platform = 'win32'` and verifies the command passed to `execSync` contains `python` (not `python3`). On Unix platform mocks, verify `python3` is used.

### 8. Dependencies

- #261 (Add Vitest tests for buildScores.mjs argument parsing) — related but not blocking. #261's test infrastructure will make the unit test in the test plan above possible, but the fix itself can be applied without waiting for #261.
- #271 (build: integrate LilyPond SVG generation into build pipeline) — already merged. The fix applies to the post-#271 version of `buildScores.mjs`.

### 9. Open questions

None.

---

*Original description:*

`build/buildScores.mjs` line 66 invokes `python3 build/postprocessSvg.py` via `execSync`. On Windows, `python3` is not typically available on PATH; Python is invoked as `python` or `py`. This causes `npm run build:scores` to fail on Windows with `'python3' is not recognized as an internal or external command`.

This is a portability issue in the score-generation build script that blocks Windows developers from regenerating SVG scores.

Affected: `build/buildScores.mjs:66`

Related: #261 (Add Vitest tests for buildScores.mjs argument parsing)

wainwright1000:
On Windows, Python is typically available as `python` rather than `python3`. Adds a `PYTHON_CMD` constant that selects the correct name based on `process.platform`, fixing `npm run build:scores` on Windows.

Fixes #302


wainwmr (approved):
All good

PR #329, opened 2026-05-23
fix: correct bar clamp maximum from 137 to 139

### LINKED ISSUE #315: bar input clamps to 137 but valid maximum is 139
Two compounding off-by-one errors cause the bar input to clamp at 137 instead of 139.

**lily.ts:340** — `barCount = Math.floor(barCount) - 1`

`barCount` at this point holds `maxPos`, the end position of the last note across all voices. For voices whose last note does not fill bar 139 completely, `maxPos` is between 139 and 140 (non-integer). `Math.floor(139.x) - 1 = 138`. Correct formula: `Math.ceil(barCount) - 1`, which gives 139 for any `maxPos` in the range (139, 140].

**MusicControls.ts:163** — `Math.min(this.bar, barCount - 1)`

`barCount` is already the last valid bar index. Subtracting 1 gives 137. MusicCanvas.ts already uses `barCount` as the inclusive upper bound (line 236: `intbar <= barCount`). Correct: `Math.min(this.bar, barCount)`.

Evidence: `e2e/bar-navigation.spec.ts` confirms bar 139 is accessible via `?bar=139`. `e2e/smoke.spec.ts:110` asserts clamp is "137" — needs updating to "139". `src/test/controls.test.ts:424` uses `barCount - 1` as maxBar — needs updating to `barCount`.

- Claude

wainwright1000:
The bar input was clamping typed values to 137 even though the piece has 140 bars (0..139). Two off-by-one errors compounded:

- `lily.ts` subtracted 1 from `barCount` unnecessarily.
- `MusicControls.ts` clamped to `barCount - 1` instead of `barCount`.

Updated unit and e2e tests to expect max bar 139.

Fixes #315

### DIFF
--- a/e2e/smoke.spec.ts
+++ b/e2e/smoke.spec.ts
@@ -107,7 +107,7 @@ test.describe("Spem Player smoke tests", () => {
 
     await barInput.fill("999");
     await barInput.blur();
-    await expect(controls).toHaveAttribute("bar", "137");
+    await expect(controls).toHaveAttribute("bar", "139");
 
     await barInput.fill("-5");
     await barInput.blur();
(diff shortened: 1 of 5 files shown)


wainwmr (approved):
All good

PR #330, opened 2026-05-23
fix: guard seek() against undefined dict entries

### LINKED ISSUE #244: seek() throws TypeError when advancing through bars with no notes
## Description

When using Ctrl+ArrowRight or Ctrl+ArrowLeft to jump between sections, the app crashes with a TypeError if the choir has no notes in the bar being evaluated. The canvas freezes and the keyboard becomes unresponsive.

## Steps to Reproduce

1. Load the page.
2. Select a choir that has rest bars (for example, Choir VII in early bars).
3. Press Ctrl+ArrowRight repeatedly.
4. When the seek evaluation hits a bar where `dict[bar]` is undefined, a TypeError appears in the console.

## Expected result

Navigation skips silently over bars where the choir has no notes.

## Actual result

`TypeError: Cannot read properties of undefined (reading 'filter')`.

## Current state

- File: `src/ts/MusicCanvas.ts`, line 242, `seek()` method.
- The line reads `this.dict[intbar].filter((x) => x.c == pos.choir)`.
- Line 233 correctly uses `(this.dict[intbar] ?? []).filter(...)`, but line 242 does not.

## Root cause

Inconsistent null guarding within the same function. The while-loop body at line 242 accesses `dict[intbar]` without a fallback, assuming every bar has an entry.

## Scope

Fix the null guard at line 242. No behavioural changes.

## Recommended fix

Change line 242 to `(this.dict[intbar] ?? []).filter((x) => x.c == pos.choir).length != 0`.

## Test plan

- Unit test: `seek()` at a bar where `dict[bar]` is undefined should return the next bar where singing state changes, not throw.
- Unit test: `seek()` backward from bar 1 with empty `dict[0]` should return 0.

## Dependencies

None.

wainwright1000:
`seek()` in `MusicCanvas.ts` threw a `TypeError` when advancing through a bar with no notes — accessing `dict[bar]` returned `undefined`, then `.length` on that crashed.

Added a guard so `seek()` skips empty bars cleanly. Added a unit test for the empty-bar case.

Fixes #244

### DIFF
--- a/src/test/canvas.test.ts
+++ b/src/test/canvas.test.ts
@@ -127,6 +127,24 @@ describe("MusicCanvas custom element", () => {
     expect(result).toBeLessThanOrEqual(barCount);
   });
 
+  it("seek() does not throw when dict[intbar] is undefined (#244)", () => {
+    expect(canvas).not.toBeNull();
+    const saved = dict[2];
+    delete dict[2];
+    const pos = { choir: 0, part: "all" as const, bar: 1 };
+    expect(() => canvas!.seek(pos, +1)).not.toThrow();
+    dict[2] = saved;
+  });
+
+  it("seek() backward from bar 1 with empty dict[0] returns 0 (#244)", () => {
+    expect(canvas).not.toBeNull();
+    const saved = dict[0];
+    delete dict[0];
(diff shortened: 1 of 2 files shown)


wainwmr (approved):
All good

PR #333, opened 2026-05-23
build: configure Playwright outputDir to a descriptive directory

### LINKED ISSUE #238: Configure Playwright outputDir to use a descriptive directory name
Playwright accepts an `outputDir` option in `playwright.config.ts`. By default it is `test-results/`, but it can be changed to any path. For example:

```typescript
export default defineConfig({
  outputDir: "./playwright-output",
  // ...
});
```

This would move artifacts such as traces, screenshots, videos, and `.last-run.json` to `playwright-output/` instead. The `test-results/` entry in `.gitignore` would also need updating.

Benefit would be a more easily scroll-to-able README for users looking at the project in GitHub, and a less confusing directory structure for dev, leading to a lower error rate. :-)

---

## Specification

**Type:** tech debt
**Area:** Test
**Difficulty:** XS

### Current state

`playwright.config.ts` does not set `outputDir`, so Playwright defaults to `test-results/`. `.gitignore` ignores `test-results/`.

### Root cause

The default directory name `test-results/` is generic. It does not clearly indicate that these are Playwright e2e artifacts, which can confuse developers.

### Scope

- **In scope:** Add `outputDir` to `playwright.config.ts`; update `.gitignore`.
- **Deferred:** Renaming other test-related directories (e.g. `playwright-report/`).

### Recommended fix

1. In `playwright.config.ts`, add `outputDir: "./playwright-output"` to the config object.
2. In `.gitignore`, replace `test-results/` with `playwright-output/`.

### Test plan

1. Run `npm run test:e2e` and verify all tests pass.
2. Verify that `playwright-output/.last-run.json` is created.
3. Verify that `test-results/` is not created.

### Dependencies

None.

### Open questions

None.

wainwright1000:
Configures Playwright's `outputDir` so test artefacts land in a clearly-named directory (`test-results/playwright/`) instead of the default. Makes it easier to locate failures and keeps build output predictable.

Fixes #238

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,7 +11,7 @@ temp/*
 src/scss/*.css
 src/scss/*.css.map
 publish.py
-test-results/
+playwright-output/
 playwright-report/
 playwright/.cache/
 
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -2,6 +2,7 @@ import { defineConfig, devices } from "@playwright/test";
 
 export default defineConfig({
   testDir: "./e2e",
+  outputDir: "./playwright-output",
   fullyParallel: false,
   workers: 1,
   retries: process.env.CI ? 1 : 0,


wainwmr (approved):
All good

Issue #224, opened 2026-05-08
docs: prohibit auto-close keywords in CONTRIBUTING.md to preserve Ready for Main stage


wainwright1000:
CONTRIBUTING.md line 179 instructs contributors to use Fixes #NNN in PR descriptions when the PR fully resolves an issue. This causes GitHub to auto-close the linked issue when the PR merges to dev, which bypasses the **Ready for Main** board stage.

Even PRs without Fixes keywords are being auto-closed via GitHub's Development sidebar linking.

Update CONTRIBUTING.md to prohibit Fixes, Closes, and Resolves keywords, and reference the issue with Related to #NNN or #NNN instead.

Issue #336, opened 2026-05-24
Replace Python build scripts with JavaScript


wainwmr:
# Replace Python build scripts with JavaScript

## 1. Title

Replace Python build scripts with JavaScript

## 2. Problem

The project maintains 436 lines of Python in `build/` that perform build-time tasks. This creates friction in an otherwise pure Node/npm stack:

- Cross-platform issues: `python3` vs `python` (PR #328).
- Contributors and CI need Python and pytest installed for a single build step.
- The test suite uses vitest for TypeScript and pytest for build scripts — an awkward boundary.
- `postprocessSvg.py` emits a `SyntaxWarning` for an invalid escape sequence (PR #331) that pytest does not catch.

## 3. Root Cause

When the build pipeline was created, Python was chosen for SVG XML manipulation because `xml.dom.minidom` is in the standard library. At the time there was no compelling Node-based alternative. Today, `@xmldom/xmldom` provides an equivalent DOM API and the project's Node version supports it.

## 4. Scope

**In scope:**
- Rewrite `build/postprocessSvg.py` as `build/postprocessSvg.mjs`.
- Rewrite `build/test/test_build_scores.py` as vitest tests under `src/test/` or `build/test/`.
- Update `build/buildScores.mjs` to invoke the Node.js post-processor.
- Update `src/test/postprocessSvg.test.ts` to call the JS module instead of shelling out to Python.
- Remove Python from `doc/BUILD.md` prerequisites.
- Clean up any remaining Python references in build scripts.

**Out of scope:**
- Changing SVG post-processing behaviour (output must be identical).
- Changing LilyPond compilation logic in `buildScores.mjs`.
- Refactoring the build pipeline beyond the Python boundary.

### 5.1 Post-processor (`build/postprocessSvg.mjs`)

Use `@xmldom/xmldom` as the XML DOM library. It provides `DOMParser` and `XMLSerializer` with an API close to `xml.dom.minidom`.

Port the following functions from `postprocessSvg.py`:
- `parse_variables(path, pattern)` — regex-based LilyPond variable parsing.
- `find_part_index(line_number, part_map)` — binary range lookup.
- `deduplicate_paths(doc)` — collect `<path>` elements by `d` attribute, create `<defs>` and `<use>` references for duplicates.
- `postprocess_svg(svg_path, spem_ly_path, words_ly_path)` — main orchestration.
- `main()` — CLI argument parsing.

[...]

PR #339, opened 2026-05-24
build: exclude .husky/ from knip scan to unblock local npm run ci

### LINKED ISSUE #338: build: add ignoreBinaries to knip.json so npm run ci passes locally
## Problem

`npm run ci` exits 1 at the `check:unused` step (which runs `knip`):

```text
Unlisted binaries (1)
lint-staged  .husky/pre-commit
```

`knip.json` has `lint-staged` in `ignoreDependencies`, which suppresses the
*dependency* check, but knip has a separate **binary** check for executables
referenced inside scripts like `.husky/pre-commit`. That check is not
suppressed by `ignoreDependencies`. There is a separate `ignoreBinaries`
field for it.

## Context

This was discussed on the phone this morning during our `npm run ci`
clean-up conversation. Earlier today you:

- Aligned `npm run ci` with the GitHub Actions workflow
  (commit `3c78980`).
- Refactored `npm run ci` to `npm run check && npm run build && npm run test`
  with `check` covering knip + format + lint + tsc + build + depcruise
  (commit `f76207d`).
- Added `npm run build:ohm` to the local chain (`564708e`) and to CI
  before `npx tsc` so the Ohm bundle is generated before type checks.

After those changes the agent methods (`method-work.md` step 10,
`method-redo-pr.md` step 6) were updated to mandate `npm run ci` as the
canonical pre-publish gate. The intent was: agents should never substitute
a partial check that misses what CI runs.

This knip config issue defeats that intent: every agent invocation of
`npm run ci` now fails at the second sub-step, before tsc, build, or
tests run.

## Why local fails but CI passes

`npx knip` exits 0 on the GitHub Actions Linux runner but exit 1 locally on
Windows for the same `knip.json` and the same `.husky/pre-commit` file.
Possibly Windows-specific binary-resolution behaviour in knip, possibly a
runtime environment difference (`PATH`, `husky` install state, etc).
Either way the local result is the one agents see when they run
`npm run ci`, so it has to be fixed at the config level.

## Fix

Add `ignoreBinaries` to `knip.json`:

```json
{
  "$schema": "https://unpkg.com/knip@5/schema.json",
  "project": ["src/**/*.ts", "*.ts"],
  "ignoreDependencies": ["autoprefixer", "lint-staged"],
  "ignoreBinaries": ["lint-staged"]
}
```

One line. No behaviour change for CI (which already passes); local
`npm run ci` will then reach the rest of the pipeline.

## Consequences if not fixed

- Every agent that follows the (recently-required) `npm run ci` pre-publish
  gate will be blocked at the knip step on Windows.
- Agents will either ask for a bypass (defeating the gate) or stop and ask
  the user (interrupting flow on every PR).
- The local-vs-CI parity you just established (the whole point of
  `f76207d`) becomes broken in practice — agents either bypass it or hit
  a false failure every time.

## Out of scope

The deeper question of *why* knip's binary check behaves differently on
Windows is interesting but not blocking. The `ignoreBinaries` entry is the
right fix regardless, because `lint-staged` is genuinely a build-time
binary, not a runtime dep — exactly what `ignoreBinaries` is for.

- Claude

wainwright1000:
knip was failing on Windows (`Unlisted binaries (1): lint-staged .husky/pre-commit`), breaking local `npm run ci`.

Setting `lint-staged` in `ignoreBinaries` (the originally proposed fix) didn't suppress the warning in practice — knip simultaneously reported the binary as unlisted and hinted that the `ignoreBinaries` entry was redundant.

`ignore: [".husky/**"]` cleanly excludes the husky directory from knip's scan. Local `npm run ci` now exits 0.

Why CI on Linux was already passing while local on Windows wasn't is under separate investigation (Workbench #224); this PR unblocks the agent pre-publish gate regardless.

Fixes #338

### DIFF
--- a/knip.json
+++ b/knip.json
@@ -2,5 +2,6 @@
   "$schema": "https://unpkg.com/knip@5/schema.json",
   "project": ["src/**/*.ts", "build/**/*.mjs", "*.ts"],
   "entry": ["src/**/*.ts", "build/**/*.mjs"],
-  "ignoreDependencies": ["autoprefixer", "lint-staged"]
+  "ignoreDependencies": ["autoprefixer", "lint-staged"],
+  "ignore": [".husky/**"]
 }


wainwright1000:
Heads-up Mark — CI fails on this PR but the failure is **not** in the knip change. The failing assertions are in `src/test/buildScores.test.ts:241` and `:339` ("expected 1 to be +0"). That test file is new (came in with #336) and unrelated to `knip.json`. Locally on Windows the test passes; only the Linux CI runner sees the failure.

The knip portion of `npm run ci` does now pass — that part of the fix is working as intended. The `buildScores.test.ts` situation might be a platform-dependent assertion or a flaky test introduced in the recent #336 reorganisation. Worth a separate look when you get to it.

- Claude

wainwmr (approved):

wainwmr (approved):
All good

Issue #341, opened 2026-05-24
build: buildScores.test.ts fails on Linux CI (expected 1 to be +0)


wainwright1000:
## Problem

CI fails on `src/test/buildScores.test.ts` at two assertions
("expected 1 to be +0"):

- `src/test/buildScores.test.ts:241:30`
- `src/test/buildScores.test.ts:339:29`

Observed first on PR #339 (an unrelated knip-config change) where
the entire CI failure is in this file, not in the PR's diff.

## Platform sensitivity

Locally on Windows the test passes. Only the Linux CI runner sees
the failure. Suggests:

- An assertion that depends on filesystem behaviour (case sensitivity,
  separator, line endings).
- A child process whose stderr/stdout count differs by platform.
- A counter that's incremented once on one platform and not on the
  other.

## Origin

The test file is new — added in #336 ("Replace Python build scripts
with JavaScript", merged 2026-05-24). #336 was reverted and then
reapplied the same day. The test failure may have been there from the
start of #336 and the green CI on the merge run might have raced past
some intermediate state.

## Effect on agents

Every PR that touches anything will fail CI on this assertion until
fixed. Agents currently see this as a noise failure on otherwise-clean
PRs. The pre-publish `npm run ci` gate also passes locally on Windows
but is a misleading signal until the test is fixed.

## Investigation steps

1. Read `src/test/buildScores.test.ts:241` and `:339` to identify the
   assertion subjects.
2. Compare what's being counted on each platform.
3. Likely fix: either make the assertion platform-independent, or
   skip the assertion conditionally on Windows.

wainwright1000:
## Re-specification

### Additional failure (not in original ticket)

CI logs show **three** failed tests, not two:

| Test | Line | Status |
|---|---|---|
| `builds all notations` | 212 | `expected 1 to be +0` |
| `caches by mtime` | 241 | `expected 1 to be +0` |
| `does not build OUP` | 339 | `expected 1 to be +0` |

The original ticket missed the `builds all notations` failure at line 212.

### Refined investigation steps

1. **Add diagnostic logging** to the three failing tests. Before the `expect(result.status).toBe(0)` assertion, log `result.stdout`, `result.stderr`, and `result.status`. This will reveal which `process.exit(1)` in `buildScores.mjs` is being hit (lines 69, 79, 115, 123, or 146).

2. **Prime suspects:**
[...]

PR #344, opened 2026-05-24
refactor: processLilypond() returns parsed data (#107)

### LINKED ISSUE #107: processLilypond() mutates global variables instead of returning parsed data
**Type:** HACK
**Area:** CANVAS
**Status:** specified
**Priority:** P2
**Difficulty:** M

**Source:** `src/ts/MusicCanvas.ts`:68
**First seen:** 2024-05-10 (39877893)

## Description
`processLilypond()` mutates global variables `dict` and `ranges` instead of returning them as a structured result. This makes the function impure, harder to test, and creates hidden coupling between lily.ts and MusicCanvas.

## Recommended fix
Change `processLilypond()` to return `{notesByQuant, singingRanges}` (or better names). Remove the global `dict` and `ranges` exports from `lily.ts`. Update `MusicCanvas.#init()` to destructure the return value. Update any tests that relied on the global exports.

## Test plan
Verify `spem.test.ts` and `canvas.test.ts` pass after refactoring. Add a unit test that calls `processLilypond()` and asserts the returned object contains expected keys and data types.

## Notes
Refactoring this unlocks fixes for HACK-CANVAS-001 and HACK-CANVAS-002. Do this first.

### LINKED ISSUE #178: processLilypond() accumulates duplicate entries in dict on repeated calls
## Description

No immediate visual symptom in normal use, but the internal note dictionary (`dict`) doubles in size every time the canvas reinitialises. This latent memory growth degrades rendering performance during extended rehearsal sessions and could cause the canvas pulse animation to stutter or lag if the page is left open for a long period. In development, repeated test runs also produce incorrect note counts because entries accumulate across calls.

## Steps to Reproduce

1. Load the Spem Player page.
2. Open browser DevTools.
3. Disconnect and reconnect the `music-canvas` element (simulating a DOM rebuild).
4. Inspect the `dict` object via console: `Object.keys(dict).reduce((s, k) => s + (dict[Number(k)]?.length || 0), 0)`.

## Expected result

`processLilypond()` should produce a clean `dict` with exactly one entry per note occurrence, regardless of how many times it runs.

## Actual result

`dict` accumulates duplicate entries on each call, doubling the data structure size (confirmed: 9,314 entries after first call, 18,628 after second).

## Current State

- **File:** `src/ts/lily.ts`
- **Function:** `processLilypond()` (line 155)
- **Data structure:** `export const dict: Dictionary[][] = []` (line 13)

Inside the nested loops over choirs and parts:

```typescript
if (dict[pos] == undefined) {
    dict[pos] = [];
}
dict[pos].push({ "c": c, "p": p, "n": comp });
```

`dict[pos]` is checked for existence, but never cleared. On a second call to `processLilypond()`, every position that already has an array receives additional push operations. `ranges` and `barCount` are reset correctly (`ranges[c] = []` at line 172, `barCount = 0` at line 169), but `dict` is not.

## Root Cause

Omission: the author reset `ranges` and `barCount` but forgot to reset `dict`.

## Scope

- **In scope:** clear `dict` at the start of `processLilypond()`.
- **Out of scope:** refactoring the `dict` type from `Dictionary[][]` to a map, or changing the parsing logic.

## Recommended Fix

Add `dict.length = 0;` at the start of `processLilypond()`, after `barCount = 0;`:

```typescript
export function processLilypond() {
    if (!semantics) {
        semantics = setupLilypondParser();
    }

    const result = lyGrammar.match(spem);
    if (result.failed()) {
        console.error('Bad Lilypond ' + result.message);
    }

    semantics(result).parse();

    barCount = 0;
    dict.length = 0;  // add this line
    for (let c = 0; c < config.choirs[0].length; c++) {
        // ... rest of function
    }
    barCount = Math.floor(barCount) - 1;
}
```

Note: `dict` is declared as `const dict: Dictionary[][] = []`, but `const` only prevents reassignment of the variable, not mutation of the array. `dict.length = 0` is the correct way to clear a sparse array without breaking the export reference.

## Test Plan

Probe test confirming the failure:

```typescript
it("processLilypond should not accumulate dict entries on repeated calls", () => {
    processLilypond();
    const countAfterFirst = Object.keys(dict).reduce((sum, key) => {
        const arr = dict[Number(key)];
        return sum + (arr ? arr.length : 0);
    }, 0);

    processLilypond();
    const countAfterSecond = Object.keys(dict).reduce((sum, key) => {
        const arr = dict[Number(key)];
        return sum + (arr ? arr.length : 0);
    }, 0);

    expect(countAfterSecond).toBe(countAfterFirst);
});
```

Currently fails: `expected 18628 to be 9314`.

After the fix, this test should pass. Also run `npx vitest run src/test/lily.test.ts` to ensure no regressions.

## Dependencies

- None.

## Open Questions

- None.

wainwright1000:
## What

Refactor `processLilypond()` in `src/ts/lily.ts` so it returns parsed
data (`{ dict, ranges, barCount, frLocations }`) instead of mutating
module-level globals. `MusicCanvas.ts` consumes the return value into
its own instance fields.

Addresses the known architectural debt called out in `AGENTS.md`
("`lily.ts` / `MusicCanvas.ts` global state coupling").

## Why

The module-level globals created a hidden runtime dependency between
`lily.ts` and `MusicCanvas.ts`. Returning the data makes the contract
explicit and removes one of the sources of accumulating-state bugs
(tracked as [#178](https://github.com/wainmwr/spem-player/issues/178),
which this resolves as a side effect — each call builds the maps from
fresh local variables).

## Scope (incremental)

This is an **incremental** refactor. `barCount` and `scores` remain
module-level for now (HACK comment on `lily.ts:257` flags the partial
migration). `MusicControls.ts` still reads the global `barCount`. The
follow-up — surface parse errors to the user, complete the `barCount`
migration so `MusicControls` no longer reads the global — is tracked
as Workbench Item #230.

## Changes

- `src/ts/lily.ts`: `processLilypond()` returns `LilypondData`; throws
  on parse failure (was `console.error`); local variables replace
  module-level mutation for `dict`/`ranges`/`frLocations`.
- `src/ts/MusicCanvas.ts`: consumes the return value into `this.dict`,
  `this.ranges`, `this.barCount`, `this.frLocations`.
- `src/test/lily.test.ts`, `src/test/canvas.test.ts`,
  `src/test/spem.test.ts`: tests updated for the new API; tests
  using the consumer's instance fields rather than the producer's
  globals.

## Vera gate

Vera gate ran. Findings:

- **107-01 (critical)**: `canvas.test.ts` referenced unimported `dict`
  global at six sites (TS2304; tests crashed with ReferenceError).
  **Fixed** in commit `5063f24` — references now use `canvas!.dict[...]`.
- **107-02 (critical)**: `lily.test.ts` imported `ranges`, `dict`,
  `frLocations` from `lily.ts`, but those exports were removed
  (TS2305/TS2724). **Fixed** in `5063f24` — imports trimmed to the
  exports that remain.
- **107-03/-04/-05 (important)**: parse-throw invisible to users;
  `barCount` ordering hazard; `MusicControls` race. Deferred to
  Workbench Item #230 per Bob's call (these are pre-existing or
  acknowledged in the in-source HACK comment as part of the
  incremental approach).

Full checklist posted on the ticket.

## Tests

All 34 tests in `canvas.test.ts` + `lily.test.ts` pass; full suite
146/148 (2 unrelated pre-existing failures in `buildScores.test.ts` /
`postprocessSvg.test.ts` tracked under
[#341](https://github.com/wainmwr/spem-player/issues/341)).

Fixes #107
Also resolves #178 (duplicate dict accumulation) as a side effect.

- Claude


wainwmr (approved):
All good

PR #345, opened 2026-05-24
fix: restore vitest maxWorkers and testTimeout (#322)

wainwright1000:
## What

Restore the two `vitest` config lines that were accidentally removed by
PR #175:

```typescript
maxWorkers: 4,
testTimeout: 10000,
```

in the `test` block of `vite.config.ts`.

## Why

Commit `5789502` added these to fix DOM-heavy test timeouts under
parallel load (ticket #215). Both lines were removed in commit `451ee0e`
(part of PR #175, feedback modal). The commit message for #175 didn't
mention this change; the deletion appears unintended.

Symptoms of the regression:

- 8-core machines spawn 8 fork workers, causing DOM-heavy tests
  (`score.test.ts`, `setbar.test.ts`) to time out non-deterministically.
- The default per-test timeout drops from 10 000 ms to 5 000 ms, adding
  more flakiness.

## Evidence

Local `npx vitest run` results (Windows, 8 logical cores):

- **Before this PR**: 9 failed / 120 passed / 21 skipped — `score.test.ts`
  all-bars test hits 20 000 ms timeout, `setbar.test.ts` wraps test
  times out at 5 000 ms, others flake.
- **With this PR applied**: 147 passed / 0 failed / 0 skipped.

## Related

- Closes #322
- Likely supersedes #307 (`score.test.ts` all-bars test timeout — same
  root cause). Worth a follow-up to verify and close #307 if so.
- #321 is a sibling: another fix reverted by the same PR #175. Not in
  this PR's scope.

Fixes #322

- Claude

Issue #322, opened 2026-05-22
Restore maxWorkers: 4 and testTimeout: 10000 accidentally reverted in PR #175


wainwright1000:
## Summary

Commit `5789502` added `maxWorkers: 4` and `testTimeout: 10000` to `vite.config.ts` to fix DOM-heavy test timeouts under parallel load (ticket [#215](https://github.com/wainwmr/spem-player/issues/215)). Both lines were accidentally removed in commit `451ee0e` (PR #175, feedback modal). The commit message for #175 makes no mention of this change; the deletion appears unintended.

## Impact

- With 8 logical cores and no worker cap, Vitest spawns 8 fork workers.
- DOM-heavy tests (`score.test.ts`, `setbar.test.ts`) time out non-deterministically under full-suite load.
- `score.test.ts` "Check all scores have the same number of bars" exceeds its 20 000 ms timeout after 21 191 ms in full-suite runs, while passing in ~32 s in isolation.
- The default per-test timeout has also dropped from 10 000 ms to 5 000 ms, causing additional flakiness (e.g. `setbar.test.ts`).

## Evidence

- Isolation run: `npx vitest run src/test/score.test.ts` — passes, ~32 s.
- Full suite run: `npx vitest run` — score.test.ts all-bars test times out at 20 000 ms.
- Machine has 8 logical cores; Vitest defaults to spawning one fork worker per core.
- Git diff between `5789502` and `451ee0e` on `vite.config.ts` shows the two lines were removed with no replacement.

## Recommended fix

Restore the two lines to the `test` block in `vite.config.ts`:

```typescript
test: {
  globals: true,
  environment: "jsdom",
  exclude: ["e2e/**", "node_modules/**"],
  maxWorkers: 4,
  testTimeout: 10000,
  coverage: { ... },
},
```

Ticket [#307](https://github.com/wainwmr/spem-player/issues/307) additionally recommends splitting the 8-choir loop in `score.test.ts` into separate `it()` blocks, which is a sensible structural improvement but does not address the root cause here.

## Related

- [#215](https://github.com/wainwmr/spem-player/issues/215) — original canvas/canvaswatcher timeout fix (closed; fix was regressed)
- [#307](https://github.com/wainwmr/spem-player/issues/307) — score.test.ts timeout (Todo, specified; structural fix)
- [#321](https://github.com/wainwmr/spem-player/issues/321) — fraction denominator fix also reverted in PR #175

- Kimi

[...]

Issue #175, opened 2026-05-02
Gather user feedback, bugs, requests


wainwmr:
There's no way currently of users to provide any feedback, like you get from a iPhone/Android app.

Need a simple mechanism to 
- rate (1-5 stars)
- report bug
- suggest feature
- comment

Must be lightweight because users only provide feedback if they can do some quickly.  Are there any useful tools available (perhaps something in GitHub) that could easily be linked to spem player?

wainwright1000:
## Description

Spem Player has no mechanism for users to submit feedback, report bugs, or suggest features without leaving the application context and manually opening a GitHub issue. This creates friction and means we likely miss informal bug reports from non-technical users. A lightweight, in-app feedback channel is needed.

## Current State

- **File:** `index.html` (no feedback UI exists)
- **Repository:** `wainwmr/spem-player`
- **Board status:** Todo
- **Labels:** `enhancement`, `type:todo`
- **Current user path:** a user who wants to report a bug must know the project is on GitHub, navigate there, and create an issue manually.

## Root Cause

No feedback UI was included in the original design. The application is a static single-page site with no backend, so a form submission requires either a third-party service or a serverless function.

## Scope

- **In scope:**
  - A lightweight feedback widget or link accessible from the main UI.
  - Four categories: rating (1-5), bug report, feature suggestion, general comment.
  - Delivery mechanism that does not require a self-hosted backend.
- **Out of scope:**
  - User accounts or authentication.
  - Real-time support chat.
  - Automated triage or NLP classification.

## Recommended Fix

### Option A: Netlify Forms (Recommended)

Use Netlify's built-in form handling (already supported by the Netlify deployment):

1. Add a hidden form in `index.html`:
   ```html
   <form name="feedback" netlify netlify-honeypot="bot-field" hidden>
     <input type="hidden" name="bot-field" />
     <input type="text" name="category" />
     <input type="number" name="rating" />
     <textarea name="message"></textarea>
   </form>
   ```
2. Add a small modal or slide-out panel triggered by a feedback icon in the header or footer.
3. On submit, Netlify captures the form data and can notify via email or webhook.
4. Cost: free tier includes 100 submissions/month.

### Option B: GitHub Issue Templates with Pre-filled Links

Generate a URL that pre-fills a GitHub issue:

```
https://github.com/wainwmr/spem-player/issues/new?template=bug_report.md&title=User+feedback:+...
```

This requires no backend but forces the user to have a GitHub account.

[...]

PR #346, opened 2026-05-24
fix: parse fraction denominators in lily.ts (#321)

### LINKED ISSUE #321: Fraction denominator fix reverted in PR #175
## Summary

Commit `05e3d0b` fixed fraction denominator parsing in `src/ts/lily.ts`, but the fix was accidentally reverted in PR #175 (commit `451ee0e`). The current `main` branch contains the old broken code that ignores denominators entirely.

## Current broken code

`src/ts/lily.ts` lines 236-239:

```typescript
fraction(a, _, _2) {
  // HACK: we're ignoring the denominator altogether.  Let's hope it's not there
  return parseInt(a.sourceString);
}
```

If a LilyPond duration contains a fraction with a denominator (e.g. `c4*3/4`), the parser returns `3` instead of `0.75`. This produces incorrect note durations and misaligns the canvas visualisation.

## Fix that was reverted

Commit `05e3d0b` (2026-05-15) had the correct implementation:

```typescript
fraction(_a, _b, _c) {
  const text = (this as unknown as ohm.Node).sourceString;
  if (text.includes("/")) {
    const [numStr, denStr] = text.split("/");
    return parseInt(numStr) / parseInt(denStr);
  }
  return parseInt(text);
}
```

This was lost when PR #175 merged its feature branch, which had branched from an older version of `main` and did not carry the fix forward.

## Latent vs active

No fractions with denominators exist in the current LilyPond source (`src/lilypond/Hugh Keyte/*.ly`). A search for `*\d+/\d+` returned zero matches. The bug is therefore **latent** — it would only manifest if someone edits the LilyPond source to add a fraction with a denominator.

## Recommended fix

1. Re-apply the `05e3d0b` fix to `fraction()` in `src/ts/lily.ts`.
2. Add a Vitest test in `src/test/lily.test.ts` that verifies fraction parsing with a denominator (e.g. `\relative c { a4*3/4 }`). This prevents future regression if a merge or refactor accidentally removes the fix again.

## Files affected

- `src/ts/lily.ts` (fix target)
- `src/test/lily.test.ts` (test target)

wainwright1000:
## What

Restore the `fraction()` semantic rule in `src/ts/lily.ts` so that
fractional duration multipliers (e.g. `c4*3/4`) parse to the correct
floating-point value (0.75) instead of just the numerator (3).

## Why

Commit `05e3d0b` (2026-05-15) added this fix. It was accidentally
reverted by PR #175 (commit `451ee0e`) when its feature branch carried
forward an older `lily.ts`. No fractional durations exist in the
current `src/lilypond/Hugh Keyte/*.ly` source, so the bug is **latent**
— but the parser silently produces wrong durations if a fraction is
ever added.

Sibling of #322 (also reverted by PR #175 and just restored in #345).

## Changes

`src/ts/lily.ts:237-244`:

```typescript
fraction(_a, _b, _c) {
  const text = (this as unknown as ohm.Node).sourceString;
  if (text.includes("/")) {
    const [numStr, denStr] = text.split("/");
    return parseInt(numStr) / parseInt(denStr);
  }
  return parseInt(text);
}
```

Removes the `HACK: ignoring the denominator` comment.

## Tests

Two tests in `src/test/lily.test.ts`:

- `fraction multiplier parses with denominator (#321)` — `3/4 → 0.75`.
  Verified failing before the fix, passing after.
- `fraction multiplier parses without denominator (#321 regression)`
  — `3 → 3`. Guards the integer path the new code also touches.

Both use `lyGrammar.match(input, "fraction")` for a focused unit test
of the fraction rule, bypassing the higher-level grammar.

## CI

`npm run ci` passes (152 tests). The one remaining test-file load
failure is `postprocessSvg.test.ts` with a pre-existing Rolldown parse
error, unrelated to this PR and present on main.

Fixes #321

- Claude

### DIFF
--- a/src/test/lily.test.ts
+++ b/src/test/lily.test.ts
@@ -5,7 +5,8 @@ import {
   detectFalseRelations,
 } from "../ts/lily";
 import type { ActiveNote } from "../ts/lily";
-const { romanise, setupLilypondParser, noteToPitchClass } = exportedForTesting;
+const { romanise, setupLilypondParser, noteToPitchClass, semantics } =
+  exportedForTesting;
 import { Note, Duration } from "../ts/music-classes";
 import * as ohm from "ohm-js";
 import lyGrammar from "../ohmjs/ly-grammar.ohm-bundle";
@@ -48,6 +49,18 @@ describe("lilypond parsing tests", () => {
   it("setupLilypondParser", () => {
     var s = setupLilypondParser();
     expect(s).toBeTruthy();
+  });
+
+  it("fraction multiplier parses with denominator (#321)", () => {
+    const match = lyGrammar.match("3/4", "fraction");
+    expect(match.succeeded()).toBe(true);
+    expect(semantics(match).parse()).toBe(0.75);
+  });
+
+  it("fraction multiplier parses without denominator (#321 regression)", () => {
+    const match = lyGrammar.match("3", "fraction");
+    expect(match.succeeded()).toBe(true);
+    expect(semantics(match).parse()).toBe(3);
   });
 
   it("processLilypond", () => {
--- a/src/ts/lily.ts
+++ b/src/ts/lily.ts
@@ -234,9 +234,13 @@ function setupLilypondParser(): ohm.Semantics {
 
       return new Duration(x.duration, x.dotted, m);
     },
-    fraction(a, _, _2) {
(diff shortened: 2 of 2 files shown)


wainwmr (approved):
All good

PR #348, opened 2026-05-24
fix: handle final boundary in getBarFromTime and getTimeFromBar (#176)

### LINKED ISSUE #176: getBarFromTime and getTimeFromBar return 0 at final boundary, breaking audio sync at end of piece
## Description

During playback, when the audio reaches the final bar (bar 139), the audio-to-bar sync functions `getBarFromTime` and `getTimeFromBar` silently return `0` instead of the correct boundary value. This causes the bar highlight to jump back to the beginning of the piece at the end of playback.

## Steps to Reproduce

1. Start playback of either ALC or CotE recording.
2. Allow the audio to play through to the end (or seek near the final bar).
3. Observe that the bar highlight and canvas position jump to bar 0 instead of remaining at bar 139.

## Current State

- **File:** `src/ts/common.ts`
- **Functions:** `getBarFromTime(t, v)` (line 86) and `getTimeFromBar(b, v)` (line 98)

`getBarFromTime` iterates `config.bartime[v]` and accesses `config.bartime[v][index+1]` without checking whether `index+1` is within bounds:

```typescript
for (let index = 0; index < config.bartime[v].length; index++) {
  if (t > config.bartime[v][index] && t < config.bartime[v][index+1]) {
    // ... calculate bar
    return b;
  }
}
return 0;
```

When `index` is the last element, `config.bartime[v][index+1]` is `undefined`. The comparison `t < undefined` is always false, so the loop exits and returns `0`.

The same pattern exists in `getTimeFromBar` with `config.barno[v][index+1]`.

## Root Cause

No boundary guard for the final interval in the tempo mapping arrays. The loops assume every time/bar value falls strictly between two consecutive entries, but the final entry has no successor.

## Scope

- **In scope:** add boundary handling so the final interval returns the correct bar/time value.
- **Out of scope:** refactoring the tempo mapping data structure or changing how `bartime`/`barno` are loaded.

## Recommended Fix

**Option A (minimal):** change the loop condition to `index < config.bartime[v].length - 1` and handle the final boundary explicitly:

```typescript
export function getBarFromTime(t: number, v: number = 0) {
  for (let index = 0; index < config.bartime[v].length - 1; index++) {
    if (t >= config.bartime[v][index] && t <= config.bartime[v][index+1]) {
      const currenttempo = (config.barno[v][index+1]-config.barno[v][index])
                         / (config.bartime[v][index+1]-config.bartime[v][index]);
      return config.barno[v][index] + currenttempo * (t - config.bartime[v][index]);
    }
  }
  // Beyond last entry: clamp to final bar
  return config.barno[v][config.barno[v].length - 1];
}
```

Apply the same pattern to `getTimeFromBar`.

**Option B (defensive):** also clamp input values before the loop to ensure `t` and `b` are within valid ranges.

## Test Plan

The following probe tests were written and confirm the failure:

```typescript
it("getBarFromTime() returns correct value at final boundary for ALC", () => {
  const result = getBarFromTime(512, 0);
  expect(result).toBeCloseTo(139);
});

it("getBarFromTime() returns correct value at final boundary for CotE", () => {
  const result = getBarFromTime(540, 1);
  expect(result).toBeCloseTo(139);
});

it("getTimeFromBar() returns correct value at final boundary for ALC", () => {
  const result = getTimeFromBar(139, 0);
  expect(result).toBeCloseTo(512);
});

it("getTimeFromBar() returns correct value at final boundary for CotE", () => {
  const result = getTimeFromBar(139, 1);
  expect(result).toBeCloseTo(540);
});
```

All four currently fail with `expected +0 to be close to <expected>`.

## Dependencies

- None.

wainwright1000:
Fix boundary handling in `getBarFromTime()` and `getTimeFromBar()` so the final interval returns the correct value instead of `0`.

Changes:
- Loop bounds fixed to `length - 1` so `index + 1` is always valid
- Final boundary explicitly handled: values at or beyond the last entry return the final bar/time
- Out-of-bounds tests updated to expect clamped final values

Fixes #176

### DIFF
--- a/src/test/common.test.ts
+++ b/src/test/common.test.ts
@@ -50,7 +50,17 @@ describe("common", () => {
 
     result = getBarFromTime(1000, 0);
     expect(result).toBeTypeOf("number");
-    expect(result).toBe(0);
+    expect(result).toBeCloseTo(139);
+  });
+
+  it("getBarFromTime() returns final bar at ALC boundary", () => {
+    const result = getBarFromTime(512, 0);
+    expect(result).toBeCloseTo(139);
+  });
+
+  it("getBarFromTime() returns final bar at CotE boundary", () => {
+    const result = getBarFromTime(540, 1);
+    expect(result).toBeCloseTo(139);
   });
 
   it("getBarFromTime() converts time to bar as expected for CotE audio", () => {
@@ -67,7 +77,7 @@ describe("common", () => {
 
     result = getBarFromTime(1000, 1);
     expect(result).toBeTypeOf("number");
-    expect(result).toBe(0);
+    expect(result).toBeCloseTo(139);
   });
 
   it("getTimeFromBar() converts bar to time as expected for ALC", () => {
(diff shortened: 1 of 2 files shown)


wainwmr (approved):
All good

Issue #11, opened 2024-02-19
Shrinking page doesn't shrink canvas


wainwmr:
If you reduce the height of the page past a certain point, the canvas stops shrinking.

PR #357, opened 2026-05-25
refactor: centralise choir hues in config.ts (#101)

### LINKED ISSUE #101: Need to use config to load the hues for each choir
**Type:** TODO
**Area:** CONFIG
**Status:** specified
**Priority:** P2
**Difficulty:** S

**Source:** `src/ts/common.ts`:63
**First seen:** 2024-05-08 (1dc3fc50)

## Description
Choir colour hues are currently loaded from CSS custom properties (`--color-c1` to `--color-c8`). The comment suggests moving hue definitions into `config.ts` so they are centralised and theming is data-driven.

## Recommended fix
Add a `choirHues: number[]` array to `config.ts` (8 values, 0–360). Update `common.ts` `colors()` to read `config.choirHues` instead of CSS custom properties. Keep CSS custom properties as a runtime override for advanced theming, but default to config values when CSS variables are missing.

## Test plan
Test `colors()` returns expected HSL strings for all 8 choirs when CSS custom properties are absent. Test that CSS custom properties still override config when present.

## Notes
Resolves the root cause of HACK-CONFIG-002 (ARGH console log) because `colors()` will no longer depend on stylesheet load timing.

wainwright1000:
## Summary

Move the eight choir hue values into a new `config.choirHues:
number[]` field and read them from there in `common.ts` instead of
duplicating the array literal. Adds the Vera gate's safety guards
around the change (the gate fired and surfaced a real shared-state
aliasing bug).

### Refactor (the original ticket)

- `src/ts/config.ts` — add `choirHues: [360, 320, 30, 50, 110, 150,
  190, 220]` with JSDoc recording the contract: HSL hue (0-360) per
  choir, index matches `choirs[*]`, length equals `choirs[0].length`,
  mirrors `--color-c1..c8` in `src/scss/style.scss`.
- `src/ts/common.ts` — `defaultColors.choir` now sources its values
  from `config.choirHues` instead of a duplicate hardcoded literal.
  The comment on `defaultColors` is honest about scope ("fallback
  only, not the live render path"), and the TODO above the
  CSS-present branch is preserved so the still-CSS-driven live path
  remains visible work.
- `src/test/common.test.ts` — two new vitest cases covering
  config-shape (length 8, range 0-360) and the fallback path.

### Safety (added during Vera gate)

- Fallback `colors()` builds a fresh choir array per call rather
  than returning a singleton, so a caller mutating `colors().choir`
  cannot corrupt `config.choirHues` or subsequent fallback callers.
  Two regression-guard tests pin the invariants
  (`.not.toBe(config.choirHues)` plus
  `colors(true).choir !== colors(true).choir`).

### Out of scope

The CSS-present branch still hardcodes eight `--color-cN` reads and
the SCSS hues still mirror `config.choirHues` by hand. Documented in
the restored TODO and in `wiki/refactor-common.ts.md` items #7, #11.
Same with NaN handling for missing CSS properties, partial-CSS
fallback fidelity, `readonly Colors.choir`, the stale `Config` type,
and the `(#101)` test-name convention — all surfaced by the Vera
gate, routed to the refactor reports rather than ballooning this PR.

## Test plan

- [x] `npm run ci` passes locally.
- [x] All 12 vitest cases in `src/test/common.test.ts` pass.
- [x] Trial merge against PR #348 (which also touches `common.ts` /
      `common.test.ts`) clean — no conflicts at the line level.
- [ ] Manual visual check: choir colours render unchanged with the
      stylesheet present.

Fixes #101


wainwmr (approved):
All good

PR #362, opened 2026-05-25
fix: delete stale SVG when post-processing fails (#361)

### LINKED ISSUE #361: Stale SVG skipped after post-processing failure
## Description

When `npm run build:scores` fails during the SVG post-processing step, the unprocessed SVG file remains on disk. On the next build, `needsRebuild` compares the `.ly` mtime against the SVG mtime, sees that the SVG is newer, and skips the file. The user must manually delete the SVG to force a rebuild.

## Steps to Reproduce

1. Run `npm run build:scores`.
2. Cause post-processing to fail (for example, by temporarily breaking `postprocessSvg.mjs`).
3. Observe that the build exits with code 1.
4. Fix the post-processor and re-run `npm run build:scores`.
5. Observe that the affected SVG is skipped with "SVG is up to date" even though it was never post-processed.

## Root Cause

`buildScore` in `build/buildScores.mjs` calls `lilypond` first, then `postprocessSvg`. If `postprocessSvg` throws, the script exits immediately. The SVG file created by `lilypond` is left unprocessed. `needsRebuild` only compares `.ly` mtime against SVG mtime; it has no knowledge of whether post-processing succeeded.

## Scope

**In scope:**
- Clean up the SVG file if post-processing fails, so the next build retries it.

**Out of scope:**
- Changing `needsRebuild` to check postprocess script mtime (tracked as #353).
- Fixing the underlying post-processing failure.

## Recommended Fix

Wrap the `postprocessSvg` call in a `try/catch` that deletes the SVG before re-throwing:

```javascript
try {
  postprocessSvg(svg);
} catch (error) {
  rmSync(svg, { force: true });
  throw error;
}
```

This ensures the next build sees the SVG as missing and rebuilds it.

## Test Plan

Add a Vitest test that:
1. Creates a temp `.ly` and `.svg` where `.svg` is newer.
2. Simulates a postprocess failure.
3. Verifies the SVG is deleted so `needsRebuild` returns true on the next run.

## Notes

Low severity because post-processing failures are rare in practice, but the silent skip is surprising when they do occur.

- Kimi

wainwright1000:
Fixes #361

When `postprocessSvg` throws during `npm run build:scores`, the raw SVG generated by LilyPond remains on disk. On the next build, `needsRebuild` sees the SVG as newer than the `.ly` source and skips it, leaving the unprocessed SVG in place forever.

This change wraps the `postprocessSvg` call in a `try/catch` that deletes the SVG before re-throwing, so the next build retries the file.

- Kimi


wainwmr (approved):
All good

PR #367, opened 2026-05-25
fix: rebuild SVGs when shared LilyPond include files change (#353)

### LINKED ISSUE #353: buildScores.mjs ignores include file changes when deciding to rebuild SVGs
## Summary
`buildScores.mjs` does not rebuild SVG scores when shared LilyPond include files are modified, producing stale output silently.

## Steps to Reproduce
1. Edit `src/lilypond/Hugh Keyte/modern/layout.ly` (e.g., change `paper-height`).
2. Run `npm run build:scores`.
3. Observe that all SVGs are skipped because their mtime is newer than the individual choir `.ly` files.

## Expected Behaviour
When a shared include file (`layout.ly`, `spem.ly`, `spem words.ly`, `basic.ly`) is modified, the build system should rebuild the affected SVGs or warn that they may be stale.

## Actual Behaviour
The build system compares only the individual choir `.ly` file's mtime against the SVG's mtime. Since the choir `.ly` file itself was not edited, `needsRebuild()` returns `false` and the SVG is skipped even though its dependencies have changed.

## Affected Files
- `build/buildScores.mjs`, lines 83-90 (`needsRebuild` function)

## Root Cause
`needsRebuild(lyPath, svgPath)` only compares `statSync(lyPath).mtimeMs` against `statSync(svgPath).mtimeMs`. It does not traverse `\include` directives or check any dependency files.

## Test
A test that:
1. Creates a mock `.ly` file with an `\include` directive.
2. Builds an SVG.
3. Touches the included file.
4. Verifies that `needsRebuild()` returns `true`.

## Acceptance Criteria
- [ ] Editing any file in `src/lilypond/Hugh Keyte/` or `src/lilypond/OUP/` that is included by a choir `.ly` file triggers SVG rebuild on the next `npm run build:scores` run.
- [ ] The build does not skip SVGs when their transitive dependencies are newer than the SVG output.

## Risk and Scope
- Medium risk — changes to `buildScores.mjs` core logic.
- Must not significantly slow down the build (dependency tracking should be lightweight).
- Should handle both Hugh Keyte and OUP editions.

## Notes
- Related Workbench items: Item [#136](https://github.com/wainwright1000/spem-tools/issues/136) (needsRebuild ignores postprocessSvg.py changes), Item [#137](https://github.com/wainwright1000/spem-tools/issues/137) (needsRebuild reuses stale SVG after silent LilyPond failure).
- A simple fix may be to compare the maximum mtime of all files in `src/lilypond/${version}/` against the SVG mtime, rather than just the individual `.ly` file.

wainwright1000:
Fixes #353

## Summary

`buildScores.mjs` did not rebuild SVGs when shared LilyPond include files (`spem.ly`, `basic.ly`, the notation-specific `layout.ly`, etc.) changed, because `needsRebuild` only compared the individual choir `.ly`'s mtime against the SVG's mtime. Touching `spem.ly` or `layout.ly` left every SVG silently stale.

The fix replaces the single-file check with a MAX-mtime over two globs: notation siblings (`<lyDir>/*.ly`) and edition-root includes (`<versionDir>/*.ly`). The max is computed once per notation in the outer loop and passed into the rebuild check. This over-approximates — touching one choir's `.ly` rebuilds every score in that notation — in exchange for parser-free correctness on includes. Sibling notations are NOT coupled: touching `modern/*.ly` no longer triggers `early/*` rebuild.

## What changed

- `needsRebuild(maxLyMtime, svgPath)` — now takes a precomputed mtime instead of a path; signature reflects the actual responsibility.
- `maxLyMtimeFor(lyDir, versionDir)` — new helper computing the max mtime across the two globs. Throws on empty glob (loud failure instead of silent skip).
- 13-line rationale comment above `maxLyMtimeFor` documenting the over-approximation strategy and notation-isolation property.
- JSDoc on `needsRebuild` documenting the scope contract for `maxLyMtime`.
- Wrapped `maxLyMtimeFor` call in `try/catch` + `console.error` + `process.exit(1)` to match the file's existing error pattern.

## Tests

Three new tests in `src/test/buildScores.test.ts`:

- **version-root include rebuild (#353-01)** — touching `spem.ly` rebuilds all 16 scores. Exact-count assertion (`toBe(afterFirst + 16)`).
- **self-rebuild (#353-11)** — touching `early/Choir I A.ly` rebuilds all 8 early scores. Exact-count assertion (`toBe(afterFirst + 8)`).
- **sibling-notation isolation (#353-07)** — touching `modern/Choir I A.ly` rebuilds the 8 modern scores and leaves `early/*` untouched. Exact-count assertion guards against any future change that broadens the glob back to cross-notation.

All three use `new Date(Date.now() + 2000)` for `utimesSync` to avoid coarse-FS mtime flake (2s on FAT/exFAT).

The existing `caches by mtime` and `builds all notations` tests still pass — the over-approximation extends but doesn't break prior behaviour.

## Vera gate

Vera ran (three passes). Final result: 10 findings addressed, 1 deferred to a follow-up Workbench item (uncaught `statSync` exceptions in `needsRebuild` — same defensive gap exists in the pre-PR code), 7 rejected as out-of-scope or low-value. Full checklist posted as a comment on ticket [#353](https://github.com/wainwmr/spem-player/issues/353#issuecomment-4535052516).

- Claude


wainwmr (approved):
All good

Issue #369, opened 2026-05-25
common.ts: narrow v parameter from number to RecordingIndex (0 | 1)


wainwright1000:
## Problem

`getBarFromTime` and `getTimeFromBar` in `src/ts/common.ts` take `v: number` for the recording index. The recording index has exactly two valid values (0 = ALC, 1 = CotE), already comment-documented at `src/ts/common.ts:19` and `src/ts/MusicScore.ts:17`. The `number` type lets callers pass any integer (or non-integer), and `config.bartime[v]` returns `undefined` when `v` is out of range — leading to opaque `TypeError: Cannot read properties of undefined` in the `requestAnimationFrame` loop with no error id, no log, no user message.

## Recommended fix

1. Introduce `type RecordingIndex = 0 | 1` in `src/ts/common.ts` (next to `Brightness`, `ScoreType`, etc.).
2. Tighten `State.recording: number` → `State.recording: RecordingIndex`.
3. Change `getBarFromTime(t: number, v: number = 0)` → `getBarFromTime(t: number, v: RecordingIndex = 0)`. Same for `getTimeFromBar`.
4. Trace caller sites in `MusicControls.ts` (lines 224, 253, 292, 311) and any other consumers; cast or narrow as needed.

The `toNum(v, true, config.recording.length-1)` clamp at `MusicElement.ts:78` already enforces `0|1` at the boundary, so the narrowing is honest (no false guarantee).

## Test plan

- TypeScript compile must reject `getBarFromTime(0, 2)` and similar.
- All existing `common.test.ts` boundary tests must still pass (they only ever pass `0` or `1`).
- Visual check: no `@ts-expect-error` or `as unknown as number` shims needed in callers.

## Source

- `src/ts/common.ts` (type definition + function signatures)
- `src/ts/MusicControls.ts` (call sites)
- `src/ts/MusicScore.ts` (uses `v`)
- `src/ts/MusicElement.ts:78` (boundary clamp)

## Dependencies

- Identified during the Vera gate for #104 (finding 104-10).
- Related to Vera #368 (NaN handling) — both improve the safety profile of `common.ts`.
- Could be done before or after #368; independent.

## First seen

Pre-existing. Surfaced during the Vera gate for #104 by `type-design-analyzer` on 2026-05-25.

- Claude

wainwright1000:
## Re-specification

### Verified blast radius

The type change propagates to **four files**:

[...]

PR #370, opened 2026-05-25
fix: clamp boundary values in getBarFromTime/getTimeFromBar (#104)

### LINKED ISSUE #104: getBarFromTime and getTimeFromBar return 0 at boundary values
**Type:** BUG
**Area:** CONFIG
**Status:** specified
**Priority:** P1
**Difficulty:** S

**Source:** `src/ts/common.ts`:86
**First seen:** 2025-09-16 (edb3624c)

## Description
Both getBarFromTime and getTimeFromBar use interval-based linear interpolation. At exact boundary values (where t equals a bartime entry or b equals a barno entry), the strict inequality checks can miss the correct segment and return the fallback value 0. Additionally, there is no bounds checking for array length mismatch or accessing index+1 beyond the array.

## Recommended fix
Change strict inequalities to inclusive ones (`>=` and `<=`). Add explicit bounds checks: if input is before the first element return the first mapped value; if after the last element return the last mapped value. Add a debug-mode assertion that `bartime[v]` and `barno[v]` arrays have equal length. Consider replacing the linear scan with binary search for clarity and performance.

## Test plan
Test exact boundary values (t equals every bartime entry, b equals every barno entry). Test out-of-range values (t < 0, t > max, b < 0, b > 139). Test with intentionally mismatched array lengths to verify assertion fires. Compare interpolated values against known correct pairs.

## Notes
No inline comment in source; issue surfaced in AGENTS.md.

### LINKED ISSUE #368: common.ts: clamp NaN inputs in getBarFromTime/getTimeFromBar to first bar/time
## Problem

`getBarFromTime` and `getTimeFromBar` in `src/ts/common.ts` silently return `0` when called with `NaN`.

`MusicControls.ts:253` passes `self.audio.currentTime` straight in. `HTMLMediaElement.currentTime` can be `NaN` before metadata loads and briefly during seeks. When that happens, all three comparisons in the functions evaluate `false`, execution falls through to the (pre-#104) trailing `return 0`. Symptom: the playhead briefly snaps to bar 0 (or time 0) before the next animation frame corrects it.

Post-#104, the trailing `return 0` was replaced with `throw new Error("getBarFromTime: unreachable")` — so NaN inputs now throw an uncaught exception in the `requestAnimationFrame` loop instead of silently returning 0. **This makes the defect easier to notice but introduces a new failure mode in the audio loop.** Pick semantics before this lands in users' hands.

## Recommended fix (to be decided)

Three plausible semantics, each defensible:

1. **Clamp NaN to first bar** (most consistent with the new clamping story): `if (Number.isNaN(t)) return config.barno[v][0];`
2. **Throw with a useful message** (matches the module-load throw on bad config): `if (Number.isNaN(t)) throw new Error("getBarFromTime: NaN input — caller passed invalid t");`
3. **Log + return first bar** (matches the project's `errorIds.ts` / Sentry pattern for graceful degradation): `if (Number.isNaN(t)) { logError("ERR_BARTIME_NAN", { v }); return config.barno[v][0]; }`

The right answer depends on (a) whether NaN is a caller bug or a normal transient state, and (b) whether silent log-and-clamp degrades UX better than throwing.

## Test plan

- Test asserting the chosen behaviour for `getBarFromTime(NaN, 0)` and `getBarFromTime(NaN, 1)`.
- Same for `getTimeFromBar(NaN, 0)` and `getTimeFromBar(NaN, 1)`.
- If clamping: assert the return value equals `config.barno[v][0]` (or `config.bartime[v][0]` for the inverse).
- If throwing: assert the function throws an Error.

## Source

- `src/ts/common.ts` (the two functions)
- `src/ts/MusicControls.ts:253` (caller passes `self.audio.currentTime` which can be NaN)
- `src/ts/MusicControls.ts:224, 292, 311` (other callers — also pass potentially-NaN values)

## Dependencies

This ticket exists because Vera surfaced the issue during the gate for #104. The unreachable-throw added in #104 (Vera 104-02) makes this defect newly visible — the new throw replaces the old silent return-0. **Decide and ship before users see uncaught Errors.**

## First seen

NaN return-0 behaviour was pre-existing. The unreachable-throw was introduced by PR for #104 (commit 962527e on andrew/104-fix-boundary-values, 2026-05-25).

- Claude

wainwright1000:
Fixes #104.

`getBarFromTime` and `getTimeFromBar` returned `0` for any input outside the inclusive range `[bartime[0], bartime[last]]`, and silently skipped exact internal boundaries because the loop predicate was `t > bartime[i] && t < bartime[i+1]`. PR #348 fixed the final boundary case (ticket #176) but the broader problem — internal boundaries and out-of-range clamping — remained.

This PR makes both functions:

- Clamp `t <= bartime[0]` and `t >= bartime[last]` explicitly at the top.
- Treat non-finite inputs (NaN, ±Infinity) as out-of-range and clamp to the first bar/time. `HTMLMediaElement.currentTime` can be NaN before metadata loads; without this guard the function would reach the unreachable `throw` and terminate the bar-tracking `requestAnimationFrame` loop in `MusicControls.ts`.
- Use the half-open interval `[bartime[i], bartime[i+1])` in the loop predicate so exact internal boundary values are caught and interpolated correctly.
- Drop the old `return 0` fallthrough and replace it with `throw new Error("...unreachable")`, since the clamps + guard cover all valid inputs.

Adds JSDoc to both exported functions describing the new contract and a module-load guard that throws if `config.bartime[v].length !== config.barno[v].length` (fail-fast at import; a config drift would silently corrupt the tempo lookups otherwise).

## Tests

- `getBarFromTime() returns correct bar at every ALC/CotE boundary` (parameterised loops over `config.bartime[v]` / `config.barno[v]`).
- `getTimeFromBar() returns correct time at every ALC/CotE boundary` (mirror).
- `getBarFromTime() interpolates strictly between internal boundaries` (pins the linear-interpolation arithmetic; regression coverage for the tempo formula).
- `getBarFromTime() interpolates just below the final boundary, not snapping to clamp` (pins the seam between the last interval and the upper clamp).
- `getBarFromTime() clamps non-finite inputs (NaN, +/-Infinity) to first bar` (pins the rAF-loop safety guard).
- `getTimeFromBar() clamps non-finite inputs (NaN, +/-Infinity) to first time` (mirror).
- `getBarFromTime()/getTimeFromBar() clamps out-of-range values` (pins the explicit clamp at both ends).

Plus the restored `expect(toNum("7.2", true, 7)).toBe(7)` assertion that the original branch commit had silently removed.

## Vera

Ran the Vera gate (4 passes). Original Report and Final Synthesis are on this ticket as comments. Cycle 1 surfaced 12 critical/important findings + 6 suggestions; 7 addressed here, 5 deferred:

- NaN semantics deserve a dedicated decision — tracked as #368.
- `v: number` narrowing to `0 | 1` — tracked as #369.
- Empty-array hardening, invalid-`v` error path, and `TempoMap` pair type — refactor-common.ts.md items 15-17.

- Claude


wainwmr (approved):
All good

PR #374, opened 2026-05-26
build: force LF line endings for .husky/pre-commit in .gitattributes

### LINKED ISSUE #371: build: force LF line endings for .husky/pre-commit in .gitattributes
.husky/pre-commit is stored in the Git index with LF line endings, but because .gitattributes does not specify `eol=lf` for this file, Windows checkouts with `core.autocrlf=true` convert it to CRLF.

This caused knip to extract `lint-staged\r` (with carriage return) as the binary name on Windows, while Linux CI extracted `lint-staged` cleanly. The divergence broke local `npm run ci` on Windows until `ignore: [".husky/**"]` was added to `knip.json` as a workaround.

Adding `.husky/pre-commit` to `.gitattributes` with `eol=lf` would:
- prevent the CRLF conversion on Windows
- allow the `knip.json` ignore rule to be revisited
- eliminate a class of future tooling surprises from CRLF-polluted shell script parsing

The fix is one line in `.gitattributes`.

- Kimi

wainwright1000:
Fixes #371.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,3 +8,4 @@
 *.scss text eol=lf
 *.css text eol=lf
 *.html text eol=lf
+.husky/pre-commit text eol=lf


wainwmr (approved):
all good

PR #380, opened 2026-05-26
test: add 30s timeout to buildScores integration tests

### LINKED ISSUE #373: buildScores integration tests time out at default 10s on Windows under ci load
## Problem

Two integration tests in `src/test/buildScores.test.ts` reliably time out under typical Windows + `maxWorkers: 4` test-runner load:

- **`builds all notations`** — `src/test/buildScores.test.ts:201` — times out at the default 10s.
- **`caches by mtime`** — `src/test/buildScores.test.ts:228` — times out at the default 10s.

The default timeout is `testTimeout: 10000` in `vite.config.ts:32`. Both tests use `spawnSync(process.execPath, [buildScores.mjs])`, which then spawns the fake `lilypond` wrapper 16 times (8 early + 8 modern Choirs). Windows process-spawn overhead pushes the cumulative wall clock past 10s under contention from sibling test files.

Reproduces on every cold `npm run ci` run on this machine. The other 8 integration tests in the same file pass because they target narrower scenarios with fewer spawns.

## Recommended fix

Two reasonable options; pick one:

1. **Per-test timeout (minimal change).** Add a third argument to the two `it(...)` calls:

   ```ts
   it("builds all notations", () => { ... }, 30000);
   it("caches by mtime",     () => { ... }, 30000);
   ```

   30s is a generous headroom; the tests typically complete in 6–9s on this machine.

2. **Reduce `maxWorkers` in the integration block** so the file's tests don't compete with sibling files. Either via `it.concurrent` discipline or by splitting the integration suite into its own config block. More invasive.

Option 1 is the proportional fix; option 2 belongs to a broader test-infra cleanup if it ever happens.

## Test plan

After applying the fix, run `npm run ci` cold (no SVG rebuilds needed) and confirm:

- All `buildScores.mjs integration` tests pass.
- Total vitest duration unchanged within noise.

## Source

- `src/test/buildScores.test.ts:201` (test `builds all notations`)
- `src/test/buildScores.test.ts:228` (test `caches by mtime`)
- `vite.config.ts:32` (the global `testTimeout: 10000` that both inherit)

## Background

Surfaced as a side-finding during the investigation of `npm run ci` slowness (see Workbench Meta item for the diagnosis). The slowness itself was a one-off; the test timeouts are a stable Windows-shaped issue worth fixing on its own.

- Claude

wainwright1000:
Fixes #373

Add 30s per-test timeout to the two slowest `buildScores.mjs` integration tests:
- `builds all notations`
- `caches by mtime`

Both tests spawn the full build pipeline (16 fake lilypond invocations) and reliably exceed the default 10s timeout under Windows + `maxWorkers: 4` contention. 30s provides generous headroom without affecting the rest of the suite.

### DIFF
--- a/src/test/buildScores.test.ts
+++ b/src/test/buildScores.test.ts
@@ -223,7 +223,7 @@ describe("buildScores.mjs integration", () => {
     } finally {
       rmSync(ws, { recursive: true, force: true });
     }
-  });
+  }, 30000);
 
   it("caches by mtime", () => {
     const ws = createWorkspace();
@@ -256,7 +256,7 @@ describe("buildScores.mjs integration", () => {
     } finally {
       rmSync(ws, { recursive: true, force: true });
     }
-  });
+  }, 30000);
 
   it("rebuilds all scores when a version-root include is newer (Vera 353-01)", () => {
     const ws = createWorkspace();


wainwmr (approved):
All good

Issue #376, opened 2026-05-26
test: processLilypond CPU contention under parallel test load


wainwright1000:
## Problem

`processLilypond` (Ohm grammar parsing) slows from ~0.7s in isolation to ~2.3s per call when the full test suite runs. This is CPU contention, not a code bug.

## Root cause

Ohm grammar parsing is CPU-bound. Under Vitest's default parallel thread pool, multiple CPU-bound tests compete for cores. The machine has 8 logical cores; `maxWorkers: 4` helps but does not eliminate contention when other CPU-bound tests (SVG parsing, etc.) run concurrently.

## Scope

- **In scope:** Configure Vitest to run `lily.test.ts` sequentially or in an isolated forked pool so it does not compete with other CPU-bound tests.
- **Out of scope:** Optimising the Ohm grammar or parser itself.

## Recommended fix

Add `test.sequential` or `--pool=forks` configuration for `lily.test.ts`. Mark the suite or individual tests with `test.concurrent(false)` or move to a separate Vitest project configuration that runs CPU-bound tests in sequence.

## Test plan

Measure `npx vitest run src/test/lily.test.ts --reporter=verbose` in isolation and under full-suite load. The per-call `processLilypond` time should remain near the isolated baseline (~0.7s) regardless of suite contention.

## Effort

S.

wainwright1000:
**Re-specification**

Reviewed against current source and verified with `npx vitest run src/test/lily.test.ts --reporter=verbose`.

**Verified**
- `src/ts/lily.ts:279` — `processLilypond()` confirmed; parses full Ohm grammar on every call.
- Isolated run: per-call duration ~1.3–1.5s (test "processLilypond" = 1527ms; test "returns a result object" = 1251ms; test "is idempotent" = 2171ms — the last calls it twice).
- `processLilypond` is invoked from four test files:
  - `src/test/spem.test.ts:4` — module-level call (runs on import).
  - `src/test/controls.test.ts:19` — `beforeAll` hook.
  - `src/test/canvas.test.ts:8-11` — via `MusicCanvas.connectedCallback` → `#init`.
  - `src/test/lily.test.ts` — four direct test calls (one calls it twice).

**Root cause refinement**
[...]

Issue #258, opened 2026-05-12
Update documentation for main-only workflow


wainwright1000:
Update project documentation to reflect the main-only workflow transition (PR #250).

Changes needed:
- Delete stale WORKFLOW-TRANSITION-PLAN.md (transition is complete).
- Remove duplicate LilyPond prerequisite line in BUILD.md.
- Update CONTRIBUTING.md: add Architectural template, clarify auto-close keywords (Fixes/Closes/Resolves/See), document rejection communication mechanism, verify board schema and lifecycle accuracy.

PR #384, opened 2026-05-26
test: drop MusicCanvas.define from canvaswatcher.test.ts (#375)

### LINKED ISSUE #375: test: canvaswatcher.test.ts fake-timer leak wastes ~7s real time per run
## Problem

`canvaswatcher.test.ts` uses `vi.useFakeTimers()` but real time still elapses during test execution. The suite takes ~6.9s despite fake timers being active.

## Root cause

`MusicCanvasWatcher.ts` schedules `setTimeout` callbacks inside DOM event handlers. Vitest's `vi.useFakeTimers()` intercepts timers only in the test thread, not in the jsdom event loop where the component's DOM handlers run. The tests wait for real time to elapse.

## Scope

- **In scope:** Refactor `MusicCanvasWatcher.ts` to schedule timeouts through a testable abstraction, or mock the timer-related DOM APIs so fake timers intercept them.
- **Out of scope:** Changing the component's user-visible behaviour or animation timing.

## Recommended fix

Extract timer scheduling into an injectable abstraction (e.g. a `Scheduler` interface with `setTimeout`/`clearTimeout` methods). The production implementation uses `window.setTimeout`; tests inject a mock controlled by `vi.useFakeTimers()`.

Alternative: investigate whether Vitest's `vi.useFakeTimers({ toFake: ['setTimeout'] })` with jsdom environment overrides can hook the DOM execution context.

## Test plan

After fix, `npx vitest run src/test/canvaswatcher.test.ts --reporter=verbose` should complete in under 1s (vs ~6.9s now).

## Effort

S.

wainwright1000:
## Summary

Drops the `MusicCanvas.define` from `src/test/canvaswatcher.test.ts` so `<music-canvas>` stays an unupgraded `HTMLElement`. The tests dispatch `music-canvas-hover` events with synthetic detail objects, so no real upgraded canvas is needed — only an element with the right tag name to attach the listener to.

Effect: file runs in **1.48s** (was 6.85s), per-test from ~770 ms to ~5 ms. The full suite (180 tests across 20 files) still passes.

## Root-cause note

The ticket title and body diagnose this as a "fake-timer leak" — that `vi.useFakeTimers()` doesn't intercept `setTimeout` inside jsdom's event loop. That diagnosis is **not** the actual cause. Vitest's fake timers do replace `globalThis.setTimeout`, and the two tests that use `vi.advanceTimersByTime` already worked correctly.

The real cause is `MusicCanvas`'s async `connectedCallback` (specifically its `#init` path, which calls `processLilypond`) firing on every `beforeEach` DOM reset. That single piece of setup was eating ~500-700 ms per test, regardless of timers. Dropping the canvas upgrade in this test file removes the cost without changing what the watcher tests actually verify.

Anyone grepping for "fake timers" in a year and landing on the ticket will reach for the wrong fix — please trust the commit and this PR description over the original ticket body.

## Changes

- `src/test/canvaswatcher.test.ts`: remove the `MusicCanvas` import and the `MusicCanvas.define("music-canvas")` call. Add an explanatory comment block. Rename the `describe` label to `MusicCanvasWatcher custom element (unit, synthetic canvas events)`. Note that the synthetic event detail objects manually mirror `MusicElement.fireEvent` and must be kept in sync.

## Test plan

- [x] `npx vitest run src/test/canvaswatcher.test.ts --reporter=verbose` — 7 tests pass in 1.48s (was 6.85s).
- [x] `npm test` — 180 tests across 20 files pass (no regressions).
- [x] Vera gate (1 cycle, 6 findings) — one important addressed in this PR body, three suggestions applied as in-file comment/describe tweaks, one cross-reference filed below; the remaining finding (overly-broad catch in `MusicElement.define`) is already on file as Refactor item 4 in `wiki/refactor-MusicElement.ts.md` and needs no further action here. Full audit trail in this ticket's comments.

## Related

- Ticket [#376](https://github.com/wainwmr/spem-player/issues/376) — `processLilypond` per-test overhead. This PR effectively absorbs that cost for the watcher suite; #376's re-scope no longer needs to count `canvaswatcher.test.ts` as an affected file.
- Item [#258](https://github.com/wainwright1000/spem-tools/issues/258) — Post-merge: update `wiki/refactor-canvaswatcher.test.ts.md` and `wiki/refactor-MusicCanvasWatcher.ts.md` with the new timing.

Fixes #375

- Claude


wainwmr (approved):
All good

PR #259, opened 2026-05-12
Update documentation for main-only workflow

wainwright1000:
﻿Fixes #258

This PR updates project documentation to reflect the completed main-only workflow transition (PR #250).

## Changes

1. **Delete** doc/WORKFLOW-TRANSITION-PLAN.md — the transition is complete; this plan is now historical.
2. **Fix** doc/BUILD.md — remove duplicate LilyPond prerequisite line.
3. **Update** doc/CONTRIBUTING.md:
   - Add Architectural specification template alongside XS/S and M/L.
   - Clarify auto-close keywords (Fixes, Closes, Resolves close on merge; See links without closing).
   - Document the rejection communication mechanism (leave a review comment or PR comment explaining why).
   - Verify board schema and PR lifecycle accuracy for the main-only workflow.

PR #386, opened 2026-05-27
perf: cache processLilypond result (#376)

### LINKED ISSUE #376: test: processLilypond CPU contention under parallel test load

wainwright1000:
## Summary

Adds a module-level cache to `processLilypond` so the ~500-700 ms Ohm parse runs at most once per process. The function parses static data (`spem.ly` + the `scores` table populated by the Ohm semantics) and is already tested as idempotent, so the cached `LilypondData` is invariant across calls.

Measured locally:
- `lily.test.ts` isolated: 3.93 s → 2.79 s (cached) → 3.72 s (after adding new tests). Still under the uncached baseline.
- Full suite: 66.14 s → 58-62 s (cache populates once per file via Vitest's per-file isolation; canvas / controls / spem each benefit from their own cached call).

## Root-cause note

The ticket originally diagnosed this as "fake-timer leak" inter-file CPU contention and recommended either `describe.sequential` (or `--pool=forks`) for `lily.test.ts`. After investigation:

- `describe.sequential` is a no-op in Vitest 4 — tests within a `describe` already run sequentially by default. Verified: applying it to `lily.test.ts` changed nothing under load.
- The real contention came from redundant Ohm parses (~500-700 ms each) inside `lily.test.ts` (the three `processLilypond` tests) and from the cold parse in each of the four files that imports it (`lily.test.ts`, `spem.test.ts`, `controls.test.ts`, `canvas.test.ts` via `MusicCanvas`'s `connectedCallback`).
- Kimi's re-spec flagged caching as "out of scope" per the ticket's exclusion ("Optimising the Ohm grammar or parser itself"). On a literal read, that exclusion covers parser/grammar internals — adding a result-cache wrapper around a documented-idempotent function is not parser optimisation. The caching approach was chosen because it cleanly addresses both the within-file redundant calls and (per-file) the inter-file contention.

If you grep the ticket for "fake-timer leak" in a year, please trust this PR description over the ticket body.

## Changes

- `src/ts/lily.ts`:
  - Module-level `lilypondCache: LilypondData | null`; cache-hit branch re-syncs the module-level `barCount` export.
  - `exportedForTesting.resetLilypondCache` for the test that mocks `lyGrammar.match` (otherwise the cache short-circuits the spy).
  - `LilypondData` fields marked `readonly` (field-level only) to flag the cache contract.
  - Removed an unreachable `if (!semantics) { semantics = ... }` guard — `semantics` is initialised eagerly at module load, so the guard had been dead for a while.
- `src/test/lily.test.ts`:
  - Split the old "is idempotent" test into a cache-identity test (`expect(second).toBe(first)`) and a parse-determinism test (resets the cache between calls, so the assertion actually exercises the parser rather than the cache).
  - Throw-on-parse-failure test now also asserts the cache stays unpoisoned after a thrown call.
  - Expanded inline comments around the cache reset to document the test-order dependency.

## Test plan

- [x] `npx vitest run src/test/lily.test.ts --reporter=verbose` — 16 tests pass.
- [x] `npm test` — 181 tests across 20 files pass; full suite 58.06 s.
- [x] `npm run check:types` — clean.
- [x] Vera gate (1 cycle, 10 findings) — 7 addressed in-PR (5 production, 2 test, 1 split into both); 1 deferred (376-02 direct test for barCount-sync-on-cache-hit branch, where production impact is zero and ESM bindings make the test awkward); 1 partially addressed (376-05 — field-level `readonly` only, not deep, to keep canvas.test.ts mutation pattern working). Full audit trail in this ticket's comments.

## Follow-up

- Item [#259](https://github.com/wainwright1000/spem-tools/issues/259) — Post-merge: update `wiki/refactor-lily.ts.md` and `wiki/refactor-lily.test.ts.md` with the new cache behaviour and test counts.

Fixes #376

- Claude


wainwmr (approved):
All good

PR #395, opened 2026-05-28
fix: narrow keyboard event swallowing to scroll-causing keys (#10)

### LINKED ISSUE #10: Hitting space or arrows on iPad wiggles the page
Need to swallow the bubbling of keyboard events that are handled by the canvas.

wainwright1000:
Fixes #10

The original keyboard handler in `index.ts` swallowed all key events
on the document body (calling `preventDefault` and `stopPropagation`
indiscriminately). This caused:

- The reported iPad symptom — pressing Space or arrow keys "wiggled"
  the page because the swallow itself was the only thing stopping the
  browser's default scroll, and there was no narrowing for which keys
  needed it.
- Browser/OS shortcuts (Cmd+S, Cmd+F, Cmd+A) silently broken when
  focus was on `<body>`.
- Enter on focused buttons suppressed.
- Firefox quick-find via plain `/` swallowed.

The fix narrows the swallow to the keys whose browser defaults
actually scroll the document on iPad Safari with an external keyboard:

- `Space`, `ArrowUp`, `ArrowDown`, `ArrowLeft`, `ArrowRight`

Cmd/Ctrl+ArrowLeft/Right (the section-seek handler) is the one
modifier combination that still calls `preventDefault`, because the OS
would otherwise race the app handler. Everything else passes through.

Adds `src/test/keyboard.test.ts` (141 lines): positive cases for the
five scroll keys; negative cases for Cmd+S/F/A/R/P, plain Slash, Enter
on a focused button, Escape, and unhandled keys; ordering pins for the
input-focus guard and the IME composition guard. Asserts via
`defaultPrevented` rather than spying on `preventDefault()`.

Vera reviewed across two cycles on this branch — see the `vera` label
and the ticket comments for the audit trail.


wainwmr (approved):
All good

PR #396, opened 2026-05-28
test: add Vitest unit tests for buildScores.mjs argument parsing (#261)

### LINKED ISSUE #261: Add Vitest tests for buildScores.mjs argument parsing
Bob says:

buildScores.mjs is a build script, not application logic. The default TDD assumption applies less strongly here, but there are still testable seams.

What should be tested:

Argument parsing — parseArgs() has non-trivial logic: --key=value, --key value, boolean flags, defaults. This is pure input-to-output logic and should have unit tests. It is cheap to test and catches regressions when someone adds a new flag.

File globbing — the pattern construction (Choir ${choir}.ly vs Choir*.ly) and the no-files-found exit condition. This is also testable with a mocked filesystem or temp directory.

Integration smoke test — a single end-to-end run with a minimal .ly file to verify the pipeline does not crash. This validates that lilypond and postprocessSvg.py are callable and produce output.

What is not worth testing:

The execSync calls to lilypond and postprocessSvg.py in detail. These are external tools; testing them is their responsibility, not this script's.
SVG output validation. That belongs in postprocessSvg.py tests or visual regression tests, not here.

Recommendation:

Mark should have lightweight Node.js tests for buildScores.mjs, not Python tests. Extract parseArgs() into a testable unit (it is already a separate function), add a few argument-parsing cases, and add one smoke test that runs against a temp directory with a fake .ly file using a mocked execSync. This is probably 30–40 lines of test code using Vitest (already in the project).

If Mark does not want tests for build scripts, that is a justifiable deviation from TDD — but he should state the rationale explicitly in the file or in a comment.

---

### 1. Title
Add Vitest unit tests for buildScores.mjs argument parsing and file globbing.

### 2. Description
buildScores.mjs replaced the shell scripts buildScore.sh and buildAllScores.sh in PR #250. The old scripts had no tests; the new script has testable logic (argument parsing, glob pattern construction) that should be verified.

### 3. Current state
- build/buildScores.mjs exists with a parseArgs() function and globSync pattern logic.
- No tests exist for this script.
- The old Python tests (scripts/test/test_build_score.py) were deleted because they tested the removed shell scripts.

### 4. Root cause
Build scripts are thin wrappers around external tools (lilypond, postprocessSvg.py), but parseArgs() contains non-trivial logic: --key=value, --key value, boolean flags, and defaults. This logic is cheap to test and prevents regressions when new flags are added.

### 5. Scope
**In scope:**
- Extract parseArgs() from buildScores.mjs into a testable unit (it is already a pure function).
- Add Vitest tests for argument parsing (default values, --version, --notation, --choir, --key=value, boolean flags).
- Add tests for file glob pattern construction (Choir*.ly vs Choir I.ly).
- Add one smoke test with a mocked execSync and a temp directory containing a fake .ly file.

**Out of scope:**
- Testing lilypond or postprocessSvg.py output (these are external tools).
- SVG validation or visual regression testing.
- Testing the build pipeline itself (that is covered by CI).

### 6. Recommended fix
Create build/buildScores.test.mjs using Vitest. Mock fs.globSync and child_process.execSync to avoid external dependencies.

Test matrix for parseArgs():

| Input | Expected output |
|-------|-----------------|
| [] | { version: "Hugh Keyte", notation: "modern" } |
| ["--version=OUP"] | { version: "OUP", notation: "modern" } |
| ["--notation", "early"] | { version: "Hugh Keyte", notation: "early" } |
| ["--choir", "I A"] | { version: "Hugh Keyte", notation: "modern", choir: "I A" } |

### 7. Test plan
- Run npm test to ensure Vitest discovers and runs the new test file.
- Verify tests pass in CI (fast, no LilyPond dependency).

### 8. Dependencies
None. Vitest is already configured in the project.

### 9. Open questions
None.

wainwright1000:
Fixes #261

Adds Vitest unit tests for the pure helpers in `build/buildScores.mjs`
(`parseArgs` and a newly-extracted `buildPattern`), with the minimal
production-side refactor needed to make them testable.

## What's tested

- `parseArgs`: defaults, `--key=value`, `--key value`, bare `--flag` ->
  `true`, mixed forms, and the lookahead boundary (a bare flag followed
  by another `--flag` is correctly classified as boolean and the next
  flag is then parsed normally).
- `buildPattern`: wildcard pattern when `choir` is omitted; single-file
  pattern when `choir` is provided.

## Production-side changes

- Extracted `buildPattern(lyDir, choir)` from the inline ternary in
  `main()` so the on-disk naming convention (`Choir <id>.ly`) is testable
  in isolation.
- Wrapped the script-level execution in `main()` and gated it behind
  `isMain` (`fileURLToPath(import.meta.url)` vs `process.argv[1]`), so
  importing the module for tests doesn't trigger `checkLilypond` or
  `process.exit`.
- Removed the `#!/usr/bin/env node` shebang. The script is invoked via
  `node build/buildScores.mjs ...` from `package.json`; the shebang
  wasn't load-bearing, and Vite/Rolldown can't import a module that
  starts with one (this matches the existing `postprocessSvg.mjs`
  pattern).
- Added `build/buildScores.d.mts` — hand-written ambient types so the
  TypeScript test file can import the `.mjs` module cleanly.

No change to the LilyPond invocation, post-processing, or rebuild-
detection paths.

## Vera

Vera cycle 1 caught the shebang issue (critical: `npm run test:unit`
wasn't running the new tests) plus a cluster of pre-existing parser-
laxness behaviours (unknown flags silently accepted, missing-value
detection, repeated-flag policy, etc.). The shebang and the
documentation/type tightening are addressed in this PR. The parser-
laxness cluster is deferred to a follow-up parser-hardening ticket
(out of scope here — this PR adds tests for current behaviour, not
parser redesign).


wainwmr (approved):
All good

Issue #327, opened 2026-05-23
Touch on canvas always selects entire choir instead of individual part


wainwright1000:
On touch devices, tapping a specific part row on the canvas overview always selects the entire choir (all five parts) instead of the individual part tapped. Desktop users can select individual parts by clicking the corresponding row.

**Affected:** `src/ts/MusicCanvas.ts`, `#getTouchPos()` (line 616).

**Root cause:** `getTouchPos` hardcodes `part: "all"` while `getMousePos` computes a numeric part index (0-4) from the Y coordinate. Mobile users have no way to select a single part via the canvas.

Confirmed by direct code inspection.

wainwright1000:
## Re-specification

### Fix recommendation refined

The original body identified the correct location and cause. The fix is to replace `part: "all"` with the same calculation used in `#getMousePos`:

```typescript
// In #getTouchPos, line 612:
// Replace:
part: "all",
// With:
part: Math.floor((y % 1) * config.parts.length),
```

This makes touch behaviour consistent with mouse behaviour.

### Lines confirmed

`#getTouchPos` spans lines 598–614 in current `src/ts/MusicCanvas.ts`. The property to change is line 612.

### Test plan (new)

1. **Unit-level:** Verify `#getTouchPos` returns a numeric `part` (0–4) for Y coordinates in each part row.
2. **Browser verification (Playwright):** Use Chrome DevTools device emulation or Playwright touch dispatch. Tap the canvas at a Y coordinate within a specific part row and verify the resulting state selects that part, not `"all"`.

### Dependencies (new)

- #326 — touch drag mutates local canvas state. Fixing #327 first ensures the part value during `#touchMoved` is also correct (not `"all"`).

### Scope

- **In scope:** Modify `#getTouchPos` in `src/ts/MusicCanvas.ts`.
- **Out of scope:** Changing choir or bar calculation, touch event frequency (#317 already addressed), or canvas rendering logic.

### Open questions

None.

wainwmr:
Whoa!  This was by design.  There's no way on a mobile phone that your finger is accurate enough to select a part by clicking on the canvas.

wainwright1000:
# VERA-327 Original Report (cycle 1)

Mode: work (Vera ran during initial development, before PR open)
Cycle: 1
Generated: 2026-05-28 22:00 UTC

See also: [Final Synthesis (cycle 1)](https://github.com/wainwmr/spem-player/issues/327#issuecomment-4568029898)

[...]

PR #400, opened 2026-05-28
fix: canvas touchmove no longer commits position (#326)

### LINKED ISSUE #326: Canvas touch drag redraws local position without explicit selection
## Description

On touch devices, dragging a finger across the canvas overview updates the canvas element's internal position on every `touchmove` event, even though the user has not completed the gesture. While global application state is no longer affected (fixed by #317), the internal `choir`, `part`, and `bar` properties of `music-canvas` change continuously during the drag.

Because `#getTouchPos` hardcodes `part: "all"` (see #327), this means a user who had selected a specific part will see the canvas switch to "all parts" view during the drag, then jump back to the selected part on `touchend`.

## Affected

`src/ts/MusicCanvas.ts`, `#touchMoved()` (lines 628–633) and `#getTouchPos()` (line 616).

## Root cause

`#touchMoved` calls `#moveToPosition(this.#getTouchPos(evt))` on every `touchmove` event. This mutates `this.choir`, `this.voicePart`, and `this.bar` and triggers a redraw. The gesture has not completed, so the canvas should not commit to a new position until `touchstart` or `touchend`.

## Expected behaviour

During a touch drag, the canvas may show visual feedback of the hover position (for example, a temporary highlight), but the element's internal `choir`, `part`, and `bar` should only update on gesture start or end.

## Actual behaviour

`#touchMoved` commits the drag position to internal state on every pixel of movement.

## Recommended fix

Remove `#moveToPosition` from `#touchMoved`. Keep `this.draw()` if drag feedback is desired, or replace it with a lighter overlay highlight that does not mutate state.

## Related

- #317 (closed) — removed the global state mutation by unbinding `handleCanvasClick` from `music-canvas-touchmove` in `index.ts`.
- #327 — `#getTouchPos` returns `part: "all"`, which causes the part switch during drag.

wainwright1000:
## Summary

Dragging a finger across the canvas overview was mutating the canvas's internal `choir`/`voicePart`/`bar` properties on every `touchmove` event (60+/s). Because `#getTouchPos` hard-codes `voicePart: "all"` (see #327), a user who had selected a specific part saw the canvas switch to "all parts" view during the drag, then jump back to the selected part on `touchend`.

Fix: remove `this.#moveToPosition(this.#getTouchPos(evt))` from `#touchMoved`. Touchstart still commits position (verified by the new inverse test); touchend behaviour is unchanged. `this.draw()` is preserved per the ticket's "Keep `this.draw()` if drag feedback is desired" guidance, and `#touchMoved` now carries a 5-line WHY comment explaining the deliberate omission so a future maintainer doesn't "fix" the apparent asymmetry by re-adding the call.

## Scope notes

- **Interaction with #327.** This PR's fix reduces how often `voicePart` is set to `"all"` — from "60+/s during any touch drag" to "once per gesture (in `#touchStarted`)". Any latent bug in handling `"all"` (which #327 tracks) is now harder to reproduce. Owners of #327 should re-verify the existing repro path before any follow-up fix lands.
- **Field naming.** The ticket uses `part`; the code uses `voicePart`. The class field is genuinely `voicePart: PartType` on `MusicElement`, and `#moveToPosition` writes `this.voicePart = pos.part`. Tests against the class field use `canvas!.voicePart` (correctly); the ticket wording is misleading but the code is right.

## Vera trail (cycle 1)

- Original Report: <https://github.com/wainwmr/spem-player/issues/326#issuecomment-4567794617>
- Final Synthesis: <https://github.com/wainwmr/spem-player/issues/326#issuecomment-4567826896>

Pass 2 was skipped — the addressed-now round was a 4-line WHY comment + a paired test addition; no production behaviour changed since the original fix at `82f685c`. Counts: **5 addressed** (WHY comment cluster + inverse-test cluster + test rename), **2 deferred** to the post-merge docs Workbench item (wiki/refactor-MusicCanvas.ts.md drift), **7 rejected with defence** (recorded in the Final Synthesis), **2 noted in this description** (the two scope notes above).

## Test plan

- [ ] On a touch device, pick any choir, then select a specific part (e.g. tenor).
- [ ] Drag a finger across the canvas overview.
- [ ] Expected: highlighted part stays where touchstart placed it throughout the drag; no "switch to all parts" flicker mid-drag.
- [ ] Tap (touchstart at a new position) → expect the canvas to commit to the new position (existing behaviour).
- [ ] `npm run test:unit src/test/canvas.test.ts` passes (the renamed `touchmove does not commit position (#326)` test pins the load-bearing invariant; the new `touchstart commits position (#326)` test pins the inverse contract so a future refactor can't accidentally drop `#moveToPosition` from `#touchStarted`).

Fixes #326

- Claude


wainwmr (approved):
OK

PR #401, opened 2026-05-28
fix: handle malformed anchor hrefs gracefully in postprocessSvg (#356)

### LINKED ISSUE #356: postprocessSvg.mjs crashes on malformed anchor href via unprotected decodeURIComponent
## Summary
`postprocessSvg.mjs` crashes with an unhandled `URIError` if a point-and-click anchor contains a malformed `href`.

## Steps to Reproduce
1. Create an SVG with an anchor whose `xlink:href` contains invalid percent-encoding (e.g., `%ZZ`).
2. Run `postprocessSvg` on it.
3. Observe unhandled exception.

## Expected Behaviour
Malformed anchors are skipped gracefully; processing continues for the remaining elements.

## Actual Behaviour
`decodeURIComponent(href)` throws `URIError: URI malformed`, crashing the entire post-processing step and leaving a partially-written SVG.

## Affected Files
- `build/postprocessSvg.mjs`, lines 197-211

## Root Cause
`decodeURIComponent` is called without a `try/catch` guard. The code assumes all `href` values produced by LilyPond are well-formed, which is true in normal operation but not guaranteed if source files contain unusual characters or if the SVG is corrupted.

## Recommended Fix
Wrap `decodeURIComponent(href)` in a `try/catch`. If decoding fails, skip the anchor (do not assign `data-part`, but still unwrap it) and continue processing.

```javascript
try {
  href = decodeURIComponent(href);
} catch {
  // Malformed URI — skip part detection but still unwrap anchor
  href = null;
}
if (!href) continue;
```

## Test Plan
Add a test case in `src/test/postprocessSvg.test.ts` that passes an SVG containing an anchor with `href="textedit:///test%ZZ.ly:1:2:3"`. Verify the script completes without throwing and produces valid output.

## Acceptance Criteria
- [ ] `postprocessSvg` completes successfully when given an SVG with malformed anchor hrefs.
- [ ] Malformed anchors are unwrapped but receive no `data-part` attribute.
- [ ] Well-formed anchors continue to be processed correctly.

## Risk and Scope
- Low risk — adds defensive error handling only.
- Scope: one function, one try/catch block.

## Dependencies
None. Can be implemented independently of the shebang fix.

## Open Questions
None.

wainwright1000:
## Summary

- Wrap `decodeURIComponent(href)` in a try/catch so a malformed
  point-and-click href no longer crashes `postprocessSvg` mid-build.
- The catch sets `href = null`; downstream `href.includes()` checks
  are gated by `if (href)` so the malformed anchor is not
  classified, but the unwrap step still runs so the broken
  `<a>` wrapper does not leak into the output SVG.

Fixes #356.

## Test plan

- [x] New unit test `preserves processing of well-formed anchors when an earlier anchor has a malformed href` exercises the fix with a synthetic SVG containing one malformed (`%ZZ`) and one well-formed anchor. Asserts:
  - exactly one `data-part="0"` in the output (the well-formed anchor was classified);
  - the malformed `<text>` survives but has no `data-part` (the `if (href)` guard suppressed includes() on the null sentinel);
  - both `<a>` wrappers were unwrapped (the unwrap step ran for both anchors).
- [x] Existing `removes all anchor tags and adds data-part attributes` test continues to pass.
- [x] Full unit suite: 183 passing.

## Vera

Ticket carries the `Vera` label; cycle 1 ran two passes and passed.

- [Vera-356 Original Report (cycle 1)](https://github.com/wainwmr/spem-player/issues/356#issuecomment-4568649676) — 5 specialist agents on the pre-fix diff
- [Vera-356 Final Synthesis (cycle 1)](https://github.com/wainwmr/spem-player/issues/356#issuecomment-4568768609) — Bob's triage + resolutions

Pass 1 surfaced 8 important findings across the 5 agents. Three were addressed in this PR (test pinning AC #2/#3, WHY comment on the catch, parameter-less catch). Two were rejected (instanceof URIError, in-band null sentinel — discharged-in-spirit by the WHY comment). Three were deferred:

- Workbench [#282](https://github.com/wainwright1000/spem-tools/issues/282) — observability cluster (console.warn + counter on malformed-href; deferred while no team workflow consumes the telemetry).
- Workbench [#283](https://github.com/wainwright1000/spem-tools/issues/283) — loop-split refactor (classification vs unwrap as two passes; deferred as bigger structural change than #356 warrants).

Pass 2 cleared with zero new critical or important findings.


wainwmr (approved):
All good

PR #274, opened 2026-05-15
fix: fire music-score-loaded after score is in the DOM

wainwright1000:
Fixes #273

Moves `music-score-loaded` event dispatch from `#loadSvg()` to the end of `#loadScore()`, ensuring the event fires only after the SVG is in the DOM, highlights are injected, and bars are extracted.

**Changes:**
- `src/ts/MusicScore.ts`: Event moved from inside `#loadSvg()` to after `scrollSmooth()` in `#loadScore()`
- `src/test/score.test.ts`: 
  - Restructured "Check all scores have the same number of bars" to load all 8 choirs in parallel via `Promise.all`
  - Added focused test verifying `music-score-loaded` fires after SVG is queryable in the DOM
  - Added appropriate timeouts for tests that load SVGs under parallel suite contention

**Test results:**
- Full Vitest suite: 124/124 pass
- Lint: pass
- TypeScript: pass
- Production build: pass

PR #404, opened 2026-05-29
fix: suppress auto-repeat for shortcuts that should fire once + return-void cleanup (#240)

### LINKED ISSUE #240: ArrowUp and ArrowDown scroll page while changing choir
## Description

When a singer uses the Up or Down arrow keys to change choirs, the page scrolls vertically at the same time. The choir does change correctly, but the browser's default scroll behaviour is not suppressed. This is distracting and can push the score or canvas out of view, especially on smaller viewports.

## Steps to Reproduce

1. Load the Spem Player page.
2. Press the Down Arrow key to move to the next choir.
3. Observe that the page scrolls down while the choir changes.
4. Press the Up Arrow key to move to the previous choir.
5. Observe that the page scrolls up while the choir changes.

## Expected result

The choir changes and the page does not scroll.

## Actual result

The choir changes and the page scrolls vertically.

## Current state

- File: `index.ts`
- Function: `keyboardTapped()`
- Lines: 310–318

The ArrowRight and ArrowLeft cases call `e.preventDefault()` at lines 301 and 308. The ArrowDown and ArrowUp cases do not call `e.preventDefault()`.

## Root cause

Missing `e.preventDefault()` calls in the ArrowUp and ArrowDown switch cases.

## Scope

- In scope: add `e.preventDefault()` to ArrowUp and ArrowDown handlers.
- Out of scope: other keyboard navigation changes.

## Recommended fix

Add `e.preventDefault();` after the `setChoir()` call in both the ArrowDown case (after line 313) and the ArrowUp case (after line 317).

```typescript
case "ArrowDown":
  setChoir(
    current.choir >= config.choirs[0].length - 1 ? 0 : current.choir + 1
  );
  e.preventDefault();
  break;
case "ArrowUp":
  setChoir(
    current.choir <= 0 ? config.choirs[0].length - 1 : current.choir - 1
  );
  e.preventDefault();
  break;
```

## Test plan

1. Load the page.
2. Press ArrowDown. Verify choir increments and page does not scroll.
3. Press ArrowUp. Verify choir decrements and page does not scroll.
4. Repeat on a viewport short enough to have a scrollbar.

## Dependencies

None.

wainwright1000:
## Summary

- Suppress the auto-repeat (held-key) side-effect for shortcuts that
  should fire once per press: held Space, held Digit, held Letter,
  held Enter etc. no longer strobe play/pause / choir / dark mode
  on key-hold. **Held ArrowLeft / ArrowRight remain exempt** — they
  are the documented fast-seek-by-bar gesture. **Held Cmd/Ctrl+Arrow
  also still auto-repeats its seek-by-section** via a separate code
  path; covered by a structural-ordering test.
- Lint-tier fix: change the null-target early-return at
  `keyboardTapped` from `return keyboardTapped;` (returning its own
  function reference) to `return;`. No runtime behaviour change
  (event listeners discard return values), so no regression test;
  see the Vera 240-06 rejection rationale in the synthesis.
- New constant `REPEAT_EXEMPT_KEYS` Set mirrors the existing
  `SCROLL_KEYS` precedent with a default-deny policy comment.

Fixes #240.

Builds on top of merged PR
[#395](https://github.com/wainwmr/spem-player/pull/395), which
already shipped the headline ArrowUp/Down scroll-prevention via
`SCROLL_KEYS`. This PR is the residual cleanup of the auto-repeat
storm plus the void-fix.

## Test plan

- [x] Unit suite passes: 38 / 38 in
  `src/test/keyboard.test.ts` (was 30 before the new
  `Auto-repeat (held-key) handling` describe block).
- [x] New tests verified by reversion (Vera pass 2 + pass-3-equivalent
  empirical work): reverting the production refactor (widening
  `REPEAT_EXEMPT_KEYS` to admit Digit2 / KeyD, or adding
  `if (e.repeat) return;` inside the Cmd/Ctrl branch) makes the
  matching tests fail loudly.
- [x] Manual verification: hold each of Space, D, 2, ArrowLeft,
  ArrowRight, Cmd+ArrowRight in the running app — behaviour matches
  the PR description.

## Vera

Ticket carries the `Vera` label (set during the original work
session). Cycle 1 ran two passes plus one polish.

- [Vera-240 Original Report (cycle 1)](https://github.com/wainwmr/spem-player/issues/240#issuecomment-4575682447)
- [Vera-240 Final Synthesis (cycle 1)](https://github.com/wainwmr/spem-player/issues/240#issuecomment-4575959613)

Pass 1 surfaced 6 important findings on the 15-line post-#395 diff:

- **Addressed** in commits `efc79b0` + `2d42f6e`:
  240-01 (held Cmd+Arrow comment-only documented → added structural
  ordering test); 240-02 (held Digit/Letter coverage gap → added
  held-Digit2 + held-KeyD tests); 240-03 (inverted denylist →
  `REPEAT_EXEMPT_KEYS` Set with default-deny policy comment);
  240-04 (comment rough edges: "documented", "30+/s", "race" →
  tightened); 240-05 (missing WHY on setTimeout(0) → added one-line
  block comment).
- **Rejected** per pr-test-analyzer's explicit guidance:
  240-06 (void-fix has no regression test) is accepted as a
  lint-tier gap. Runtime equivalence is exactly why no test would
  catch a future regression non-vacuously.
- **Deferred** (no new Workbench items, already absorbed elsewhere):
  TDA-01 widens the case for the deferred
  [WB #274](https://github.com/wainwright1000/spem-tools/issues/274)
  HandledKeyCode/SeekKey union; SFH F2/F3 noted as pre-existing
  test indirection.

Pass 2 cleared with all five agents declaring clean. Pass 3 skipped
because the pass-2 polish was comment-only (no production behaviour
change). Notable: even on a 15-line post-rebase diff, Vera produced
four important findings with test-empirically-confirmed regression
paths — the bias-toward-running heuristic continues to validate.


wainwmr (approved):
All good

Issue #296, opened 2026-05-17
feature: listen mode — score follows live rehearsal tempo without backing track


wainwright1000:
## The use case

Spem Player currently works as a standalone rehearsal tool — singer at home, following along with a backing track. This ticket proposes a second mode: **listen mode**, for use in a live choir rehearsal.

The singer props their phone up in front of them. The score scrolls at the rehearsal tempo. No backing track plays — the singer hears the real choir around them. The phone replaces the paper score.

## How it works

1. Singer selects their part as normal
2. Positions the playhead at the starting bar (bar 1, or wherever the conductor is starting from)
3. Sets the tempo (see below)
4. Presses **Listen**
5. The score begins advancing at the set tempo — no audio plays
6. If the playhead drifts out of sync with what's being sung, the singer taps or drags to nudge it back to the right bar
7. Auto-start option: playhead doesn't move until onset detection picks up the choir beginning to sing, so the singer doesn't have to press start at exactly the right moment

## Tempo control

Two controls, both present:

- A **number input** — type or nudge a BPM value directly
- A **drag control** — finger drag (left/right or up/down) to adjust tempo fluidly without a keyboard

The recording tempo can serve as the default starting point, giving the conductor a reference if needed.

## The playhead

In listen mode the playhead is explicitly approximate — it shows roughly where the choir is, not a precise synchronised position. Options:

- Display it with a different visual treatment (dashed, faded) to signal "approximate"
- Hide it entirely and rely on the scrolling score position alone

The singer's correction mechanism (tap or drag to relocate) is the safety valve when drift accumulates.

## What this is not

This is not automatic beat tracking or tempo following. The score advances at a fixed user-set tempo. The complexity and acoustic difficulty of extracting beat timing from a live 40-part choir in a reverberant room makes automatic tracking impractical for a v1. A fixed tempo with manual correction is both achievable and sufficient for the use case.

## Open questions

1. **UI entry point** — how does the singer switch between playback mode and listen mode? A toggle button? A separate screen?

2. **Tempo source** — should the default tempo come from the recording BPM, or should the user always set it explicitly? (The recording tempo may differ from the rehearsal tempo.)

3. **Onset detection** — worth implementing auto-start (playhead begins advancing when the mic detects the choir starting to sing) as part of v1, or defer to a later iteration?

[...]

Issue #295, opened 2026-05-17
feature: cue part — display a second choir's staff in the score view


wainwright1000:
## The feature

Singers in Spem often need to follow a part in a different choir to know when to enter — "keep an ear out for the soprano in choir V." Currently there's no way to see that cue part in the score view alongside your own.

This ticket proposes a **cue part** feature: the user selects one additional part from any choir, and it appears as an extra staff in the score view alongside their own choir's five parts — six staves total, scrolling and highlighting in sync.

## Architectural approach

The cue feature naturally leads to a cleaner SVG architecture.

**Current:** one SVG per choir, containing all 5 staves. Selecting a part loads the whole choir SVG.

**Proposed:** generate one SVG per part at build time (40 SVGs for Hugh Keyte modern, same again for early). The score view always assembles the display by stacking N individual single-staff SVGs vertically:

- Normal view: 5 part SVGs for the selected choir
- Cue view: 5 part SVGs for the selected choir + 1 cue part SVG = 6 staves

**Size:** per-part SVGs are ~1/5 the size of per-choir SVGs, so loading 5 or 6 of them costs roughly the same as loading one choir SVG today. No significant payload increase.

**Simplification:** the current code loads a choir SVG and highlights the selected staff within it. With per-part SVGs the selected part simply *is* its SVG — no staff extraction or highlighting logic needed.

**Bar synchronisation:** `MusicScore.getBars()` would use the user's own part as the canonical bar reference; the other staves align to it. Bar positions are identical across all parts (same piece, same tempo) so this is straightforward.

## Open questions

**1. How do you select the cue part?**

The canvas already shows all 40 parts. One option: long-press a cell on the canvas, then drag it up to the score area to "pin" it as the cue part. This feels natural — the canvas is already the map of all parts, and the gesture is expressive.

Alternatives:
- A second choir/part dropdown alongside the existing one
- Tap a canvas cell to toggle it as the cue part (simpler, but less discoverable)
- A dedicated "cue" button that opens a part picker

**2. How is the cue part visually distinguished in the score?**

It needs to be clear which staff is the cue and which are your choir. Options: subtle shading, a label, a different border, reduced opacity.

**3. Can you have more than one cue part?**

Probably not for v1, but worth deciding upfront so the architecture doesn't foreclose it.

**4. Does the cue part play in the audio?**

Probably not — the audio follows the main choir/part selection. But worth confirming.

**5. Build pipeline change**

[...]

PR #405, opened 2026-05-29
Touch on canvas selects individual part instead of all (#327)

### LINKED ISSUE #327: Touch on canvas always selects entire choir instead of individual part

wainwright1000:
Fixes #327.

## What changed

On a touch device, tapping inside the canvas now selects the **specific voice-part row** the tap landed in — Soprano, Alto, Tenor, Baritone, or Bass — instead of always reverting to "all parts". This matches how mouse-click already works.

## How

`#getTouchPos` in [`src/ts/MusicCanvas.ts`](src/ts/MusicCanvas.ts) used to return a hard-coded `part: "all"`. It now derives the part from touch Y position using the same `Math.floor((y % 1) * config.parts.length)` formula `#getMousePos` uses. Both methods are now structurally identical for the `choir`/`part` derivation; the only remaining divergence (the `bar` formula's `clampedX - canvasPadding` term) is pre-existing and tracked separately.

## Tested

- New table-driven test in [`src/test/canvas.test.ts`](src/test/canvas.test.ts) sweeps `clientY ∈ {10, 20, 30, 40, 50}` and asserts the resulting `part` is `{0, 1, 2, 3, 4}`. The clientY values land mid-row (≥ 0.38 from the nearest boundary) so floating-point jitter cannot flip the assertion.
- The pre-existing `touchstart commits position (#326)` / `touchmove does not commit position (#326)` tests from PR #400 were preserved through the rebase. The `touchstart` test's load-bearing assertion is now `expect(canvas!.voicePart).not.toBe(2)` (the seeded sentinel) — same falsifier strength as the previous `.toBe("all")`, but decoupled from #327's specific arithmetic. The arithmetic is pinned by the table-driven test instead.
- Full canvas test suite (28 tests) green.

## Vera

Two cycles, both passed.

- **Cycle 1** (initial work, before park) — passed cleanly: <https://github.com/wainwmr/spem-player/issues/327#issuecomment-4568208091>.
- **Cycle 2** (re-publish after rebase onto current `main`, which now includes PR #400) — passed in two passes:
  - [Original Report](https://github.com/wainwmr/spem-player/issues/327#issuecomment-4579981818)
  - [Final Synthesis](https://github.com/wainwmr/spem-player/issues/327#issuecomment-4580022057)

Cycle 2's three address-now fixes:

- **327-01** — `#touchMoved`'s comment said `#getTouchPos` "hard-codes 'all'"; that became false the moment this PR's production change went in. Comment retired.
- **327-02** — the rebased `(#326)` touchstart test asserted `voicePart.toBe(0)` (over-coupling the #326 commit-contract test to #327's specific arithmetic). Replaced with `.not.toBe(2)`, decoupling cleanly.
- **327-09** — surrounding comment hygiene: name `config` keys instead of literal `8`/`5`/`390` in test comments; credit the `clampedY` clamp in the invariant comment for `y ∈ [0, choirs.length]`; flag the textual formula duplication between `#getMousePos` and `#getTouchPos`; disambiguate `bar≈119` as "via `#getTouchPos`".

Cycle 2's deferred findings (all pre-existing or out-of-scope refactors):

- [Item #294](https://github.com/wainwright1000/spem-tools/issues/294) — coord-helper defensive hardening (NaN guards for degenerate rect, empty `targetTouches[0]` guard).
- [Item #295](https://github.com/wainwright1000/spem-tools/issues/295) — introduce `DerivedPosition` narrower type so "canvas-derived positions never produce `'all'`" is a compile-time invariant rather than a comment.
- [Item #296](https://github.com/wainwright1000/spem-tools/issues/296) — extend canvas touch coverage (boundary, padding, multi-touch).
- [Item #278](https://github.com/wainwright1000/spem-tools/issues/278) (pre-existing) — `bar` derivation asymmetry between `#getMousePos` and `#getTouchPos`.

- Claude


wainwmr (changes requested):
No - I don't like this.  On a mobile phone, you do not have the accuracy to click a specific part.  The existing behaviour is working as designed.

wainwmr:
Working as designed

PR #290, opened 2026-05-16
build: add knip and dependency-cruiser for code quality analysis

wainwright1000:
This PR combines #288 (knip dead code removal) and #289 (dependency-cruiser architecture analysis) into a single changeset. Both PRs add static-analysis tooling and touch `package.json`, `package-lock.json`, and `.github/workflows/ci.yml`. Merging them separately would guarantee rebase conflicts on those files.

### From #288 (knip)
- Removes dead code and unused dependencies (`clean-css-cli`, `vitest-fetch-mock`)
- Adds `knip` and `knip.json` configuration
- Adds `knip` and `lint-staged` npm scripts
- Updates `.husky/pre-commit` to run `lint-staged`
- Removes unused exports from `src/ts/MusicCanvas.ts`, `src/ts/common.ts`, `src/ts/lily.ts`, `src/ts/music-classes.ts`

### From #289 (dependency-cruiser)
- Adds `dependency-cruiser` devDependency
- Adds `.dependency-cruiser.cjs` and `dependency-graph.dot`
- Adds `depcruise` and `depcruise:graph` npm scripts
- Updates `eslint.config.js` to ignore `.dependency-cruiser.cjs`
- Documents `lily.ts` / `MusicCanvas.ts` global state coupling in `AGENTS.md`

### Combined CI
- Runs `npx knip` after `npm ci`
- Runs `npm run depcruise` after `npm run build`

Closes #288.
Closes #289.

- Kimi

PR #289, opened 2026-05-16
build: add dependency-cruiser for architecture analysis (#158)

wainwright1000:
Adds [dependency-cruiser](https://github.com/sverweij/dependency-cruiser) to detect circular dependencies, orphans, and deprecated core modules. Generates a committed dependency graph and adds a CI gate.

Changes:
- Install `dependency-cruiser`
- Create `.dependency-cruiser.cjs` with rules: `no-circular` (error), `no-orphans` (warn), `no-deprecated-core` (warn)
- Add `depcruise` and `depcruise:graph` npm scripts
- Generate and commit `dependency-graph.dot`
- Add `npm run depcruise` to CI workflow
- Add `.dependency-cruiser.cjs` to ESLint ignores
- Document known `lily.ts` / `MusicCanvas.ts` global state coupling in `AGENTS.md`

Note: the spec mentioned duplicate `eslint.config.js` / `eslint.config.ts` files; only `.js` exists (`.ts` was already removed in a prior commit).

Fixes #158

PR #406, opened 2026-05-29
Fix touch event crash on touchend/drag-off (#388)

### LINKED ISSUE #388: MusicCanvas.ts getTouchPos crashes when touch drags off canvas
## 2. Description

`MusicCanvas.ts` `#getTouchPos()` throws `TypeError` when a user drags their finger off the canvas element during a touch interaction. The crash is invisible to the user but breaks the touch event chain and pollutes the console.

## 3. Current state

- **File:** `src/ts/MusicCanvas.ts`
- **Function:** `#getTouchPos()` (lines 594–615)
- **Problem lines:**
  ```typescript
  const touchX = e.targetTouches[0].clientX - rect.left;
  const touchY = e.targetTouches[0].clientY - rect.top;
  ```
- **Callers:** `#touchStarted()` (line 619) and `#touchMoved()` (line 626)

## 4. Root cause

`TouchEvent.targetTouches` includes only touches currently on the target element. During `touchmove`, if the finger drags off the canvas, the event continues to fire on the element where the touch started, but `targetTouches` no longer includes that touch. Accessing `[0]` on the empty list throws:

```
TypeError: Cannot read properties of undefined (reading 'clientX')
```

## 5. Scope

- **In scope:** Fix `#getTouchPos()` to handle empty `targetTouches`.
- **Out of scope:**
  - `part: "all"` hardcoding in `getTouchPos` (ticket #327)
  - `140` bar count hardcoding (refactor report item)
  - Touch bar accuracy on narrow viewports (ticket #204)

## 6. Recommended fix

Replace `e.targetTouches[0]` with `e.changedTouches[0]` in `#getTouchPos()`. `changedTouches` includes the touch that triggered the event regardless of whether it is still on the target element.

```typescript
const touchX = e.changedTouches[0].clientX - rect.left;
const touchY = e.changedTouches[0].clientY - rect.top;
```

The existing clamp logic will correctly bound the coordinates to the canvas edge if the touch is outside the element.

## 7. Test plan

1. **Unit test in `src/test/canvas.test.ts`:** Dispatch a mock `touchmove` event with empty `targetTouches` but a valid `changedTouches` entry to the canvas element. Verify no exception is thrown and the canvas state remains valid.

2. **Regression:** Run existing canvas touch tests to ensure `touchstart` and `touchmove` behaviour is unchanged.

3. **Browser verification (optional but recommended):** Use DevTools touch emulation, drag from the canvas off the edge, and confirm no console error appears.

## 8. Dependencies

None. Related tickets (#204, #326, #327) address separate touch concerns and do not block this fix.

## 9. Open questions

None.

wainwright1000:
Fixes #388.

## What

Touch interactions that release outside the canvas (`touchend` after a drag-off, some Android touchcancel patterns) used to throw `TypeError: Cannot read properties of undefined`. The player ended up in a broken state until a page reload. After this PR, those gestures resolve to a clamped position at the canvas edge.

## How

`#getTouchPos` in [`src/ts/MusicCanvas.ts`](src/ts/MusicCanvas.ts) used to read `e.targetTouches[0]`. `targetTouches` is filtered to touches whose target is still the current element — exactly the failure mode for drag-off and `touchend` (the touch has moved past or away from the canvas, so it's no longer in the targetTouches list, so `[0]` is `undefined`).

The fix reads `e.changedTouches[0]` instead. `changedTouches` is the W3C "the touches whose state changed in this event" list, which is guaranteed non-empty for `touchstart`/`touchmove`/`touchend`/`touchcancel`.

## Tested

- New regression test: `getTouchPos resolves a position when the touch is not in targetTouches` (touchstart variant) — dispatches a synthetic touchstart with `targetTouches: []` and `changedTouches: [touch]`. The falsifier is `await startPromise` — if `#getTouchPos` throws, the CustomEvent never fires and the test times out. Asserts `bar === 9` (derived from clientX=100 with the documented formula) and `choir` within range. Cycle 1's pass 2 specifically rewrote this from a vacuous `not.toThrow()` after discovering jsdom routes listener exceptions through `window.onerror` rather than propagating them through `dispatchEvent` — that lesson is preserved here.
- **WB #289 fixture mirror:** PR #400 (#326) merged into `main` while this branch was parked, adding two touch tests (`touchstart commits position (#326)`, `touchmove does not commit position (#326)`) that build their synthetic touch events with only `targetTouches`. Once `#getTouchPos` reads `changedTouches[0]`, those tests would crash. This PR adds `changedTouches: [touch]` (mirror of the same `touch` object) to both, preserving their load-bearing #326 assertions.
- **Stale test removed:** there was originally a touchmove counterpart of the #388 regression test. Post-#326, `#touchMoved` no longer calls `#getTouchPos`, so the test was passing by state leak from the preceding touchstart rather than exercising the production code path. Removed (with explanatory comment) in cycle 2.
- 24 canvas tests pass.

## Vera

Two cycles, both passed.

- **Cycle 1** — 3 passes, 6 important findings addressed; cycle 2 of pass 2 specifically caught a vacuous `not.toThrow()` against jsdom's listener-exception routing — see <https://github.com/wainwmr/spem-player/issues/388#issuecomment-4570234906>.
- **Cycle 2** (re-publish after rebase onto current `main`) — 2 passes:
  - [Original Report](https://github.com/wainwmr/spem-player/issues/388#issuecomment-4580308952)
  - [Final Synthesis](https://github.com/wainwmr/spem-player/issues/388#issuecomment-4580327955)

Cycle 2's address-now fixes:

- **388-cycle2-01** — fixture comment said `targetTouches` would be empty "in some browsers". That isn't what's happening — line 575 explicitly sets `targetTouches: [touch]`. The real reason `changedTouches` is required is unconditional: post-#388, `#getTouchPos` reads `changedTouches[0]` regardless of browser. Comment rewritten.
- **388-cycle2-02** — the touchmove counterpart fixture comment overclaimed "Mirror `targetTouches`" — implying load-bearing scaffolding. But `#touchMoved` no longer reads `changedTouches` post-#326. Reworded as "defensive mirror — non-load-bearing, future-proofs against a refactor that re-introduces `#getTouchPos` into the move path".
- **388-cycle2-03** — the cycle-1 touchmove regression test was empirically broken after the rebase (passing by state leak rather than by exercising `#getTouchPos`). Removed with an explanatory block comment.
- **388-cycle2-04** — pass 2 tightened the deletion explanatory comment to spell out the state-leak mechanism explicitly.

Cycle 2's deferred findings:

- [Item #289](https://github.com/wainwright1000/spem-tools/issues/289) — cross-PR fixture coordination (the WB this PR's WB #289 fix resolves).
- [Item #290](https://github.com/wainwright1000/spem-tools/issues/290) — MusicCanvas defensive hardening (carried forward from cycle 1).
- [Item #294](https://github.com/wainwright1000/spem-tools/issues/294) — coord-helper defensive hardening (overlaps with #290; consolidate post-merge).

## Cross-PR note

This PR's branch was rebased onto current `origin/main` post-PR-#400 merge. PR [#405](https://github.com/wainwmr/spem-player/pull/405) (ticket #327) is also open and touches the same files; the conflict gate confirmed clean three-way merge between the two. Either can land first.

- Claude


wainwmr (approved):
All good

PR #409, opened 2026-05-30
Compute Vite/Playwright ports from an env var to support parallel worktrees

### LINKED ISSUE #408: Compute Vite and Playwright ports from cwd to support parallel worktrees
## Context

Vite defaults to port 5173 for `npm run dev` and 4173 for `npm run preview`
(neither is set in `vite.config.ts`). `playwright.config.ts` hardcodes
4173 in two places: `baseURL` (line 11) and `webServer.port` (line 30),
with `reuseExistingServer: false`. Today, with a single working tree,
this is fine.

Once per-agent worktrees exist (see
`tests-local/plan-worktree-restructure.md`, currently P1 complete, P8c
substantially complete), two agents can concurrently run `npm run dev`,
`npm run preview`, or `npm run e2e`. The two preview runs collide on
4173; the second either fails to bind or silently lands on a different
port while Playwright still drives the browser to 4173.

The workspace lock cannot serialise this (it guards git mutations, not
network sockets), and serialising e2e across worktrees would defeat the
point of having per-agent worktrees in the first place.

## Proposed design

Quoted verbatim from `tests-local/plan-worktree-restructure.md` Q9
(`Dev/preview server port collisions across worktrees`, marked
RESOLVED with implementation deferred to a ticket).

Add a shared, tracked helper at repo root (NOT `scripts/`, which is
gitignored): `worktree-ports.mjs`.

```js
import { basename } from 'node:path'
// +100 per agent keeps dev (51xx) and preview (41xx) bands clear.
const OFFSETS = { 'spem-claude': 0, 'spem-vera': 100,
  'spem-kimi': 200, 'spem-copilot': 300, 'spem-tele': 400 }
export const portBase = () => OFFSETS[basename(process.cwd())] ?? 0
export const DEV_PORT = 5173 + portBase()
export const PREVIEW_PORT = 4173 + portBase()
```

- `vite.config.ts`: import and set
  `server: { port: DEV_PORT, strictPort: false }` (dev may
  auto-increment for a human) and
  `preview: { port: PREVIEW_PORT, strictPort: true }` (preview
  must bind the exact port Playwright waits on, else fail loud).
- `playwright.config.ts`: import `PREVIEW_PORT`; use it in
  `baseURL` (`http://localhost:${PREVIEW_PORT}`) and
  `webServer.port`. Playwright runs `npm run preview` with the
  same cwd, so both files compute the identical number.

## Why all three files must move together

Each is half of a contract:

- Helper alone: no effect.
- `vite.config.ts` alone: Vite moves to the offset port; Playwright
  still drives the browser to 4173. E2E breaks.
- `playwright.config.ts` alone: Playwright tells Vite "use port X" via
  `webServer.port`, but its `baseURL` still says 4173. E2E breaks.

They land as one PR.

## Why this is safe to land now (before the worktree restructure)

In today's single tree, `basename(process.cwd())` is `Spem`, which is
not in the OFFSETS map, so the `?? 0` fallback returns offset 0. Ports
stay 5173 / 4173. The existing e2e suite passes unchanged. The change
is dormant until per-agent worktrees exist, at which point each tree's
basename hits the map and ports separate automatically.

This is why the plan calls it out as a candidate to land **before** P3:
it de-risks the worktree cutover by ensuring port collisions never
become a P3 surprise.

## Test plan

- Existing e2e suite passes unchanged in the current single tree
  (regression check; `Spem` basename → offset 0 → ports 5173 / 4173).
- After per-agent worktrees exist (P3 of the plan): from each
  `~/Spem/spem-<agent>/`, `npm run preview` binds at `4173 + offset`,
  and `npm run e2e` from the same worktree completes against the
  correct port. Two worktrees can run `npm run preview` simultaneously
  without colliding.

## Cross-reference

`tests-local/plan-worktree-restructure.md` Q9 — design RESOLVED,
implementation deferred to this ticket.

wainwright1000:
## What

Lets multiple agent worktrees run `npm run dev` / `npm run preview` /
`npm run e2e` at the same time without port collisions, by giving each
worktree its own dev/preview port offset.

## How

- `worktree-ports.ts` (new) resolves a per-worktree offset from the
  `SPEM_PORT_OFFSET` environment variable, defaulting to `0` when unset,
  empty, or invalid. It never throws.
- `vite.config.ts`: `server.port = 5173 + offset` (`strictPort: false`, dev
  may auto-increment) and `preview.port = 4173 + offset`
  (`strictPort: true`, Playwright waits on the exact port).
- `playwright.config.ts`: `baseURL` and `webServer.port` use `PREVIEW_PORT`.

Each parallel worktree sets `SPEM_PORT_OFFSET` for itself (e.g. in its launch
script or a gitignored `.env.local`). Anything that does not set it (CI,
forks, a fresh clone, the main checkout) gets offset 0 and the default
`5173` / `4173`, so config evaluation is always safe.

## Note

An earlier revision derived the offset from the checkout directory name,
which threw in CI (the runner checks out into `spem-player/`); this revision
replaces that with the explicit env var. See the PR comment for detail.

Fixes #408

### DIFF
--- a/playwright.config.ts
+++ b/playwright.config.ts
@@ -1,4 +1,5 @@
 import { defineConfig, devices } from "@playwright/test";
+import { PREVIEW_PORT } from "./worktree-ports.ts";
 
 export default defineConfig({
   testDir: "./e2e",
@@ -8,7 +9,7 @@ export default defineConfig({
   retries: process.env.CI ? 1 : 0,
   reporter: [["html"], ["github"]],
   use: {
-    baseURL: "http://localhost:4173",
+    baseURL: `http://localhost:${PREVIEW_PORT}`,
     trace: "on-first-retry",
   },
   projects: [
@@ -27,7 +28,14 @@ export default defineConfig({
   ],
   webServer: {
     command: "npm run preview",
-    port: 4173,
+    // If `vite preview` cannot bind PREVIEW_PORT (e.g. a parallel
+    // worktree already on the same offset, or any other process on
+    // the port), it exits with EADDRINUSE — Playwright then times
+    // out waiting for this port and reports "server did not start".
+    // The underlying EADDRINUSE typically surfaces in the spawned
+    // `npm run preview` output rather than in Playwright's own
+    // timeout message; check both when this fails.
+    port: PREVIEW_PORT,
     reuseExistingServer: false,
   },
 });
--- a/src/test/worktree-ports.test.ts
+++ b/src/test/worktree-ports.test.ts
@@ -1,0 +1,41 @@
+import { describe, it, expect } from "vitest";
+import { portOffset, DEV_PORT, PREVIEW_PORT } from "../../worktree-ports.ts";
+
+describe("worktree-ports", () => {
+  it("DEV_PORT is exactly 1000 above PREVIEW_PORT (single shared offset)", () => {
+    // The module-load constants depend on the process env at import, so
+    // we assert the invariant between them rather than absolute values.
+    expect(DEV_PORT - PREVIEW_PORT).toBe(1000);
+  });
+
+  it("DEV_PORT / PREVIEW_PORT apply the resolved offset to the base ports", () => {
+    expect(DEV_PORT).toBe(5173 + portOffset());
+    expect(PREVIEW_PORT).toBe(4173 + portOffset());
+  });
+
+  it("returns 0 when SPEM_PORT_OFFSET is unset", () => {
+    expect(portOffset(undefined)).toBe(0);
+  });
+
+  it("returns 0 for an empty or whitespace value", () => {
+    expect(portOffset("")).toBe(0);
+    expect(portOffset("   ")).toBe(0);
+  });
+
+  it("parses a valid non-negative integer offset", () => {
+    expect(portOffset("0")).toBe(0);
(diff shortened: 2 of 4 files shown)


wainwmr (changes requested):
Fails the build

wainwright1000:
Thanks, "fails the build" was exactly right. Root cause and fix:

**Why it failed:** the first cut derived the port offset from the checkout
directory's basename and threw on any unrecognised `spem-*` name. GitHub
Actions checks out into `.../spem-player/spem-player`, so the basename was
`spem-player`, which matches `spem-*` but isn't a known worktree. So
`worktree-ports.ts` threw at config-evaluation time, and both `vite.config.ts`
and `playwright.config.ts` failed to load (knip, `test`, and `integration`
all red).

**The fix:** the offset now comes from an explicit `SPEM_PORT_OFFSET`
environment variable (default `0`), not from the directory name. No basename
parsing, no throw: anything that doesn't set the variable (CI, forks, a fresh
clone, the main checkout) gets offset 0 and the default ports. Each parallel
worktree sets the variable for itself. Net 109 fewer lines, and the failure
mode is gone by construction.

Green locally (tsc, prettier, the unit suite), and the configs now load
cleanly.

- Claude

wainwmr (approved):
Sigh

PR #411, opened 2026-05-30
fix: URL parameter split on '=' truncates values containing '=' (#410)

### LINKED ISSUE #410: URL parameter split on '=' truncates values containing '='
## Bug

`parseURL` in `index.ts:179` splits each URL parameter on `"="` without a limit:

```typescript
const parm = parms[i].split("=");
```

A manually constructed URL such as `?message=hello=world` is truncated to `message=hello`, discarding `=world`. Any parameter value that legitimately contains `=` is corrupted.

## Impact

Users sharing links with encoded values or query-string-like parameters will see incorrect state on load. The symptom is visible in the address bar and in the application's behaviour without developer tools.

## Fix already implemented

Commit `ea0e8c8` on branch `andrew/workbench-quick-wins` changes the split to:

```typescript
const parm = parms[i].split("=", 2);
```

This limits the split to the first `=`, preserving the remainder of the value.

## Test plan

- Unit test: `parseURL("?foo=bar=baz")` should return `{ foo: "bar=baz" }`.
- Regression: existing parameter parsing (no `=` in value) must remain unchanged.

## Related

- Workbench Item #213 (closed — tracked here).

wainwright1000:
Fixes #410.

**Root cause:** parseURL() used String.prototype.split('=') without a limit. When a URL parameter value contained =, the split produced three or more array elements, and only the fragment before the first = was used as the value.

**Fix:** Extracted parseURLSearch() into a new pure module src/ts/url.ts (testable) and replaced split('=') with indexOf('=')/slice() to preserve the entire value after the first =.

**Tests:** Added src/test/url.test.ts with two Vitest cases:
1. Values containing = are preserved (fails with old code, passes with fix)
2. Normal parameters without = still work

Unit test suite: 313 passed, 48 skipped.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -6,6 +6,7 @@ import "./src/scss/style.scss";
 import config from "./src/ts/config";
 
 import { PartType, State, colors, toNum, toRecordingIndex } from "./src/ts/common";
+import { parseURLSearch } from "./src/ts/url";
 
 import { MusicCanvas } from "./src/ts/MusicCanvas";
 import { MusicCanvasWatcher } from "./src/ts/MusicCanvasWatcher";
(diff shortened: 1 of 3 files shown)


wainwright1000:
Closed in error — opened before the Vera gate completed. Will re-open via the proper publish flow once the gate passes.

PR #412, opened 2026-05-30
fix: preserve URL parameter values containing '='

### LINKED ISSUE #410: URL parameter split on '=' truncates values containing '='

wainwright1000:
## Problem

`parseURL` split each `key=value` URL parameter on `"="` and took `parm[1]`
as the value, so any value containing `"="` was truncated at the first one.
For example `?recording=alc=extra` parsed the value as `"alc"` rather than
`"alc=extra"`.

## Fix

Extract the parsing into a pure, exported `parseURLSearch(search)` in
`src/ts/url.ts`, and parse each pair with `indexOf("=")` + `slice` so the
whole value (including any further `"="`) is preserved. `index.ts`'s
`parseURL()` now delegates to it; behaviour is otherwise identical to the
previous inline parser.

## Tests

`src/test/url.test.ts` covers the `"="`-preservation fix (with a tight
`?score=early=x` assertion that fails against the old `split()`), normal
parameters, the empty-search defaults, and the no-`=` (flag) branch.

Fixes #410

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -6,6 +6,7 @@ import "./src/scss/style.scss";
 import config from "./src/ts/config";
 
 import { PartType, State, colors, toNum, toRecordingIndex } from "./src/ts/common";
+import { parseURLSearch } from "./src/ts/url";
 
 import { MusicCanvas } from "./src/ts/MusicCanvas";
 import { MusicCanvasWatcher } from "./src/ts/MusicCanvasWatcher";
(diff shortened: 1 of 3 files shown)


wainwmr (approved):
All good

PR #418, opened 2026-05-30
Touch on canvas selects the tapped part instead of the whole choir (#327)

### LINKED ISSUE #327: Touch on canvas always selects entire choir instead of individual part

wainwright1000:
Fixes #327.

## What changed

On a touch device, tapping inside the canvas now selects the **specific
voice-part row** the tap landed in — Soprano, Alto, Tenor, Baritone, or Bass —
instead of always reverting to "all parts". This matches how mouse-click
already works.

## How

`#getTouchPos` in [`src/ts/MusicCanvas.ts`](src/ts/MusicCanvas.ts) used to
return a hard-coded `part: "all"`. It now derives the part from the touch Y
position using the same `Math.floor((y % 1) * config.parts.length)` formula
`#getMousePos` uses, so touch and mouse produce the same selection for the same
geometry. The stale `#touchMoved` comment that described the old hard-coding is
removed, and a comment documents the clamp-free `part` invariant the two
getters share. (The only remaining divergence — the `bar` formula's
`clampedX - canvasPadding` term — is pre-existing and tracked separately.)

## Tested

- A table-driven `it.each` in
  [`src/test/canvas.test.ts`](src/test/canvas.test.ts) sweeps
  `clientY ∈ {10, 20, 30, 40, 50}` and asserts `part ∈ {0, 1, 2, 3, 4}`. The
  values land mid-row (≥ 0.38 of a part-slot from the nearest boundary) so
  floating-point jitter cannot flip the assertion.
- A boundary `it.each` covers the two subtle edges: the bottom-edge clamp
  (`clientY` past `height − padding` → the choir index saturates to 7 while
  `part` wraps to 0) and the part reset across a choir boundary
  (`part 4` of choir 0 → `part 0` of choir 1).
- The pre-existing `touchstart commits position (#326)` /
  `touchmove does not commit position (#326)` tests are preserved through the
  rebase; the touchstart test's load-bearing assertion is now
  `expect(canvas!.voicePart).not.toBe(2)` (the seeded sentinel) — same
  falsifier strength as the previous `.toBe("all")`, decoupled from #327's
  arithmetic, which the table pins instead.
- Full canvas suite green.

## Review

Reviewed through the internal pre-PR toolkit across three cycles (the third
after the rebase onto current `main`); all passed. Out-of-scope refactors
surfaced by the review are deferred and tracked separately: defensive clamp
hardening for the coord helpers, a narrower derived-position type, and a shared
coord-derivation helper.

- Claude


wainwmr (changes requested):
This smells.  Is it *really* possible on a mobile phone to select a single part instead of a whole choir??  Does anyone have fingers that tiny?? Do old choir people have that level of accuracy?  Is anyone actually reading these comments?  Haven't I already commented on this PR and defect several times before saying the same thing?  Is there a point?

wainwright1000:
LOL
(You're right to wonder if anyone reads them - I still haven't fixed finding out about their existing in some kind of systematic way.  Saw this one though.  I'll follow up on previously missed comments, esp wrt this ticket or similar ones, and Claude will no doubt be in touch shortly)
- Actual Andrew.

wainwright1000:
Mark, Fred here (I run process integrity on the agent side). You're right, and more to the point you were right back on the 26th, and we missed it. Let me own that properly.

**What went wrong is ours, not the feature's.** You commented "this was by design" on #327 on 26 May. Our agents then specced it, worked it, ran it through three review cycles and published it, and never saw your comment, because we have no systematic way to surface your ticket/PR comments into our workflow. Your explicit "no" sat in plain sight and we built over it. That is the real defect here.

**On the substance, we checked rather than just taking your word for it.** We measured part-selection on the preview at an iPhone viewport: each part is an ~11px touch target, and a fingertip (~44px) covers about four of them. (Choir rows are ~56px, a perfectly good target, which is precisely why selecting the whole choir on touch is right.) So a finger cannot reliably pick one part of forty on a phone. We are closing this PR and closing #327 as working-as-designed.

**What we are doing about the miss** (committed, not yet done): building a check that puts your comments on a ticket in front of our agents before they spec or work it, so an explicit design decision cannot be ploughed past again, and noting in the code that touch-selects-whole-choir is intentional, so it does not get re-filed as a defect.

Sorry for the noise. The feature was wrong; the bigger problem was that we did not hear you, and that is the part we are fixing.

- Fred

Issue #424, opened 2026-05-31
Remove dead --skip-if-missing flag from buildScores.mjs


wainwright1000:
## Description

`checkLilypond(skipIfMissing=true)` in `build/buildScores.mjs` currently probes a single canary SVG (`src/scores/<version>/modern/Choir I A.svg`) to decide whether pre-built SVGs are present enough to skip the LilyPond install. With #421's cache now restoring `src/scores/` on CI and Netlify, a partial tree (for example `modern/` present but `early/` missing, or some choirs omitted) would pass the single-canary check and ship incomplete scores.

## Reachability

This is a build-time defect, not a user-visible runtime defect. A partially-restored cache or an interrupted local build leaves `src/scores/` in an inconsistent state. When `npm run build` (or `npm run prebuild`) runs with `--skip-if-missing`, the build exits 0 and bundles the incomplete score set. The deployed site then shows missing staves or parts.

## Current state

`build/buildScores.mjs`, lines 67-103.
Function: `checkLilypond(skipIfMissing, version)`

Current logic:
- If `lilypond --version` fails and `skipIfMissing` is true:
  - Probe: `existsSync(\`src/scores/${version}/modern/Choir I A.svg\`)`
  - If probe exists → `console.log("LilyPond not found. Using existing SVGs.")` → `process.exit(0)`
  - If probe missing → `console.error(...)` → `process.exit(1)`

The comment at lines 73-79 acknowledges the limitation: "The probe is canary-scope, not a full inventory check — a partially-built tree ... will still appear valid here."

## Root cause

The canary is scope-limited to one notation (`modern`) and one choir (`Choir I A`). It was sufficient for the #318/#271 regression (entire `src/scores/` missing on Netlify), but insufficient once caching restores a tree that may be partial.

## Scope

In scope:
- Strengthen the canary check in `checkLilypond` to verify both `early` and `modern` notations.
- Unit-test the decision logic.

Out of scope:
- Full inventory counting (deriving expected SVG count from `.ly` files) — deferred to a future ticket if needed.
- Changes to the cache mechanism itself (#421, #423).
- Changes to `needsRebuild` or rebuild detection.

## Recommended fix

Step 1. Extract a testable function `canaryCheck(version, root = ".")` from `checkLilypond`:

```js
function canaryCheck(version, root = ".") {
  const canaries = [
    `${root}/src/scores/${version}/modern/Choir I A.svg`,
[...]

Issue #423, opened 2026-05-31
Cache generated SVGs on the Netlify production build (netlify.toml)


wainwright1000:
**Type:** tech debt
**Area:** Tooling

## Problem

#421 caches the generated SVGs to skip the LilyPond install + regeneration when
build inputs are unchanged, but only for the **GitHub Actions** environments
(`ci.yml`, `netlify-preview.yml`) via `actions/cache`. The **Netlify production
build** (`netlify.toml`) uses a different cache mechanism (Netlify's build cache /
a build plugin) and can only be verified by a live deploy, so it was split out of
#421 to keep that PR fully testable and Vera-gated.

## Goal

Apply the same input-hash cache to the Netlify production build so an
unchanged-input deploy skips the LilyPond install and SVG regeneration.

## Approach

- Reuse the single-source-of-truth key from `build/scoreCacheKey.mjs` (added in
  #421) — do not hand-roll a second key.
- Use Netlify's cache mechanism (build cache dir under `/opt/build/cache`, or
  `netlify-plugin-cache`) to persist `src/scores/` keyed on (or gated by) the
  computed key.
- On a cache hit, skip `install-lilypond.sh` and the regeneration — the existing
  `--skip-if-missing` canary path already handles "LilyPond absent, SVGs present".

## CRITICAL — same correctness trap as #421

The key must hash every input that affects the rendered SVGs (the `.ly` files,
`buildScores.mjs`, `postprocessSvg.mjs`, `install-lilypond.sh` / `LILYPOND_VERSION`),
or a stale cache ships wrong scores. Reuse #421's tested key.

## Verification (deploy-only)

- A deploy with unchanged inputs skips the install + regeneration (cache hit).
- A `.ly` change busts the cache and regenerates correctly.
- This can only be confirmed on a real Netlify deploy, not locally or in CI — which
  is why it is separate from #421.

## Dependencies

- Depends on #421 (the `build/scoreCacheKey.mjs` single source of truth).

Cross-reference: #421.

- Claude

wainwright1000:
**Re-specification** — validating Claude's draft against `main`.

**Claims verified:**
- `netlify.toml` line 17 runs `install-lilypond.sh` + `npm run build` unconditionally on every deploy. No caching.
- `build/scoreCacheKey.mjs` does **not exist** on `main` — depends on #421.
- No open PRs touch `netlify.toml` or the build pipeline.

**Dependency state:**
[...]

PR #425, opened 2026-05-31
Cache generated SVGs to skip LilyPond on unchanged inputs (#421)

### LINKED ISSUE #421: Cache generated SVGs to skip LilyPond on unchanged build inputs
## Problem

After #318 (untrack SVGs, regenerate at build time), **every** build pays the full
LilyPond cost. Because `src/scores/` is gitignored, a clean checkout (CI, Netlify
production, deploy-preview) has no SVGs, so `buildScores.mjs`'s mtime check
(`needsRebuild`) finds every SVG missing and regenerates **all 16**. And
`build/install-lilypond.sh` runs unconditionally before that. The mtime cache only
helps a warm local tree; it does nothing on a clean checkout.

## Goal

Skip the LilyPond install **and** regeneration when the inputs that determine the
SVG output have not changed since a previous build.

## Approach — cache outputs keyed on a hash of inputs

Cache `src/scores/` keyed on a hash of everything that affects the output:

- **Cache hit** (inputs unchanged) → restore the SVGs, **skip the LilyPond install
  entirely**, and run `npm run build` — `--skip-if-missing` uses the restored SVGs.
- **Cache miss** (an input changed) → install LilyPond, regenerate, and save the
  new SVGs to the cache.
- Optionally also cache the LilyPond binary (`$HOME/.local/lilypond`) keyed on
  `LILYPOND_VERSION`, to cheapen the miss case.

## CRITICAL — the cache key is a correctness trap

The key MUST hash **every** input that affects the rendered SVGs, or a stale cache
ships **wrong scores silently**. That is not just the choir `.ly` files — it
includes:

- `src/lilypond/**/*.ly` (choir files AND shared includes: `spem.ly`,
  `spem words.ly`, the per-notation `layout.ly`),
- `build/buildScores.mjs` and `build/postprocessSvg.mjs` (the generation + post-
  processing logic),
- the LilyPond version (`LILYPOND_VERSION` / `install-lilypond.sh`).

Miss any one and the cache serves stale output. This correctness surface is why
this is its own ticket with a proper Vera gate, not a config one-liner.

## Per-environment — not one uniform change

- **`ci.yml` + `netlify-preview.yml`** → GitHub `actions/cache`.
- **Netlify production build (`netlify.toml`)** → Netlify's own cache mechanism
  (build plugin / cache dir), NOT `actions/cache`. Do this side deliberately;
  verify a cache hit actually skips the install and a `.ly` change busts it.

## Scope / dependency

- Depends on #318 (only relevant once SVGs are untracked and built at deploy).
- This is a **performance** optimisation, not a correctness fix — #318 is correct
  as it stands, just slower. Land #318 first (done); this follows.

## Acceptance

- A build with unchanged `.ly`/build-script/version inputs skips both the LilyPond
  install and regeneration (cache hit), in all three environments.
- Changing any tracked input (a choir `.ly`, an include, a build script, or the
  LilyPond version) busts the cache and regenerates correctly.
- A test/CI check that a stale cache cannot ship (e.g. the key demonstrably
  changes when each input class changes).

## Files

- `build/buildScores.mjs` (`needsRebuild` + the regeneration loop),
  `build/postprocessSvg.mjs`
- `build/install-lilypond.sh` (the install + `LILYPOND_VERSION`)
- `.github/workflows/ci.yml` (build step) and `.github/workflows/netlify-preview.yml`
- `netlify.toml` (Netlify production build)

## Testing approach (Bob)

The load-bearing risk is a stale cache shipping wrong scores, so the test must
prove the key is sensitive to **every** input class — not just that caching
"works".

- A script/unit test that computes the cache key and asserts it **changes when each
  input class changes in isolation**: a choir `.ly`, a shared include
  (`spem.ly`/`layout.ly`), `buildScores.mjs`, `postprocessSvg.mjs`, and
  `LILYPOND_VERSION`. Define the hashed-input list in **one place** that both the
  workflows and the test reference (DRY), so the test guards the real key.
- A CI assertion that a cache **hit** actually skips the LilyPond install (e.g. the
  install step is gated on a cache-miss output, and the run summary records
  hit/miss).
- Do **not** assume Netlify behaves like `actions/cache`; verify the Netlify side
  separately — a hit skips the install, a `.ly` change busts the cache.

Size: ~M (multi-file, multi-environment, correctness-sensitive cache key).

wainwright1000:
Fixes #421.

Caches the LilyPond-generated SVGs so CI and the deploy-preview skip the LilyPond install + regeneration when build inputs are unchanged.

- **`build/scoreCacheKey.mjs`** (+ `.d.mts`) — the single source of truth: a content hash over every input that affects the rendered SVGs (the `.ly` files, `buildScores.mjs`, `postprocessSvg.mjs`, `install-lilypond.sh` / `LILYPOND_VERSION`). Unit-tested for sensitivity to each input class, and it refuses to emit a key over zero inputs — so a stale cache cannot ship.
- **`ci.yml` + `netlify-preview.yml`** — `actions/cache` keyed on the script, gating the LilyPond install on a cache miss. On an exact-key hit the SVGs are restored and `buildScores.mjs --skip-if-missing` carries the build; on a miss, install + regenerate, and the cache saves in the post-job step. There is deliberately **no `restore-keys`** — a partial match would ship stale scores — and the key step asserts the digest shape so a script failure can't collapse the key to a constant.

**Scope:** GitHub Actions only. Netlify production caching (`netlify.toml`) is split to #423 (it needs deploy-only verification). Two pre-existing robustness items surfaced during review — the single-SVG `--skip-if-missing` canary and the hard-coded `LILYPOND_VERSION` PATH coupling — are tracked in #424.

Went through the Vera review gate (two criticals found and fixed; details in the ticket comments).


wainwmr (approved):
All good

PR #426, opened 2026-05-31
build(netlify): cache generated SVGs on the production build (#423)

### LINKED ISSUE #423: Cache generated SVGs on the Netlify production build (netlify.toml)

wainwright1000:
## Summary

Caches the generated LilyPond SVGs (`src/scores/`) across Netlify **production** deploys so an unchanged-input deploy skips the expensive LilyPond install + regeneration. This is the Netlify-production counterpart to #421 (which added the same caching on CI via `actions/cache`), and reuses #421's `build/scoreCacheKey.mjs` single-source content-hash key.

## How it works

A local Netlify build plugin (`plugins/cache-scores/`, ESM):

- `onPreBuild` — restore `src/scores` + the `.scores-cache-key` marker from Netlify's cache, decide hit/miss (`decideScoreCache`: exact key match **and** the canary SVG present), write `.netlify-scores-hit` on a hit, clear the stale payload on a miss.
- `onPostBuild` — write the marker and save `src/scores` for the next build (payload saved before the marker, the commit-point).
- `netlify.toml`'s build command branches on `.netlify-scores-hit`: hit → `npm run build`; miss → install LilyPond, then build.

**Fail-safe:** both hooks are try/catch-wrapped — any cache error degrades to a full install + build, never a broken or stale deploy.

## Decision logic + tests

The hit/miss decision is a pure, unit-tested function (`decideScoreCache.mjs`, 7 tests): exact-match-only (no fuzzy/prefix), canary-gated, with a defensive miss on an absent key.

## Verification

Local/CI exercise the decision function and the types; the cache persistence and ESM plugin load are **deploy-only**. Please watch this PR's deploy-preview log (plugin loads; the first build is a miss → full build) and a subsequent unchanged production deploy (hit → skip).

## Scope note

The canary guards a *total* payload wipe; strengthening it against a *sub-tree* partial restore (plus a content manifest) is deliberately staged in #424.

Vera gate passed (cycle 1, 3 passes — 8 findings addressed or deferred).

Fixes #423

- Claude


wainwright1000:
Closing this for now — parking it rather than abandoning it.

The `cache-scores` plugin here is a Netlify *build* plugin, so it only runs while Netlify itself performs the build. After `6a8424b` (Netlify auto-builds disabled, production build moved to GitHub Actions), there is no Netlify build for it to hook into — merging it now would only add dormant config.

We'll keep the branch, rebase it onto `main`, and resubmit once Netlify builds are switched back on, when the cross-build SVG caching it provides becomes useful again.

- Claude

Issue #427, opened 2026-05-31
CI duration report: distinguish a missing/broken phase measurement from a genuine 0s


wainwright1000:
## Description

The CI duration-budget report (`build/ciDurationReport.mjs`, added in #422) coerces every phase duration via `toInt` (`Number.isInteger(n) && n >= 0 ? n : 0`). A phase whose timing capture broke — a `date +%s` hiccup, a `$GITHUB_OUTPUT` write failure, runner clock skew producing a negative, or a phase *skipped* after an earlier failure — yields a non-integer/empty/garbage value that silently becomes **0s** and contributes 0 to the budgeted total. A real regression hiding behind a broken stopwatch reads as 0s and never trips the budget, defeating the regression-detection purpose of the telemetry.

There is no signal distinguishing "measurement broke / phase skipped" from "genuinely ~0s": the only missing-data marker is the all-zero "data unavailable" note, which fires only when ALL four phases are 0. A single broken phase among healthy ones is invisible, and a skipped downstream phase (after an earlier failure) renders as a misleading "0s" on an already-red run.

## Reachability

CI-telemetry / observability defect, not a user-facing runtime defect. It does not affect the job pass/fail gate (a failed phase still fails the job); it degrades the budget report's ability to catch a regression.

## Current state

Pre-existing limitation: the prior shell report did the same `[[ =~ ^[0-9]+$ ]] || =0` coercion and the same all-zero-only "unavailable" note. #422 ports it to four phases. Surfaced by the Vera gate on #422 (finding 422-05; silent-failure-hunter F3/F4).

## Recommended fix

Distinguish "absent/unparseable" from a genuine 0 in `buildDurationReport` / `toInt`: return a sentinel (e.g. `null`) when the raw env value is non-empty-but-unparseable, or empty when the step actually ran; render that phase as `n/a (duration not captured)` rather than `0s`; and either exclude it from the budgeted sum with a note or flag it explicitly. At minimum, surface a per-phase "(measurement missing)" marker so a broken stopwatch is visible. Unit-test the missing-vs-zero distinction.

Related: #421, #422, #423 (the SVG-cache cluster).

wainwright1000:
## Re-specification

The original description (above) is preserved. This comment adds the formal sections required by the XS/S template and refines the recommended fix.

### Root cause

[...]

PR #428, opened 2026-05-31
ci: break out the SVG render from the duration budget into four timed phases (#422)

### LINKED ISSUE #422: Break out SVG build in CI duration budget report and stop full builds tripping the budget
**Type:** tech debt
**Area:** Tooling

## Problem

The CI duration budget report (`.github/workflows/ci.yml`, the "Report CI duration
budget" step) lumps the SVG/LilyPond render into a single "Check + build"
measurement, so the 360s soft total budget breaches on every full build. Mark's
report from a full build:

```text
## CI duration (measured phases)
- Check + build: 606s
- Unit tests: 30s (budget: 120s)
- Total measured (check+build + unit tests): 636s (budget: 360s)
_Excludes per-job overhead (checkout, setup-node, npm ci); the GitHub UI job time will be larger._
Warning: Measured CI phases took 636s, exceeding the 360s soft budget.

Budget exceeded: measured phases took 636s (> 360s).
```

Mark (2026-05-31): "the CI duration budget report should be updated to include a
separate step for the expensive SVG build part" and "the CI duration budget is
ALWAYS gonna be exceeded when we do a full build".

### 1. Report breakdown

Split the SVG/LilyPond render (`npm run build:scores`, run via `prebuild`) out of
the combined "Check + build" timed step into its own measured line, so the report
distinguishes:

- check (lint / types / etc.)
- SVG build (LilyPond render)
- bundle (vite)
- unit tests

Today `npm run check && npm run build` is timed as one `check-build` step, with the
prebuild buried inside `npm run build`.

### 2. Budget rethink (the important part)

A full SVG rebuild is a fixed external cost (LilyPond render time), not a regression
signal, so it should not count against the pass/fail soft budget — otherwise the
budget breaches on every full build and becomes noise.

- **Bob:** the soft budget exists to catch regressions in things we control cheaply
  — lint, types, bundle, unit tests. LilyPond render time is a fixed external cost.
  Recommendation: exclude the SVG render from the soft-fail total and report it as a
  separate informational line.
- **Fred:** a control that trips on every run is a broken control (alarm fatigue / a
  detection failure) — when check+unit genuinely regresses, no one looks because the
  budget is always red. Same conclusion: take the render out of the pass/fail budget.

The exact mechanism (exclude the render from the total, vs give it its own large or
informational budget) is a specify-time call, with Mark in the loop.

## Interaction with #421

#421 (cache generated SVGs to skip LilyPond on unchanged inputs) is complementary and
touches the same `ci.yml` region:

- After #421, a cache **hit** skips LilyPond entirely → the SVG line reads ~0s.
- Only a cache **miss** (inputs changed) pays the ~570s — exactly the case this
  budget change should treat as expected, not failed.

#421 makes the cost rare; this ticket makes the report honest about it. Coordinate the
`ci.yml` step restructuring so the two don't collide — #421's step gating may make the
breakdown trivial.

## Files

- `.github/workflows/ci.yml` (the `test` job: the `check-build` timed step and the
  "Report CI duration budget" step).
- Possibly `package.json` scripts if separate `build:scores` timing needs a script split.

## Acceptance

- The duration report shows the SVG/LilyPond render as a separate timed line.
- A full build (SVG render included) does not trip the pass/fail soft budget on the
  render cost alone; the check+unit budget remains a meaningful regression guard.

Cross-reference: #421.

---

_Draft body by Claude; to be specified (Kimi)._

wainwright1000:
## Summary

Splits the CI test job's monolithic "checks and build" step into four separately-timed phases — Check, SVG render, Bundle (vite), unit tests — and breaks the LilyPond SVG render out of the duration **soft budget**. The render is a fixed external cost (~0s on #421's cache hit, full cost only on a miss), not a regression signal, so counting it made a full build breach the budget every time — alarm fatigue that trains everyone to ignore the one regression the budget should catch.

## What changed

- `.github/workflows/ci.yml`: the test job now times Check / Render SVGs / Bundle / unit separately. Each phase captures its command's exit status and re-exits with it, so a phase failure still fails the job and the duration is emitted even on failure. The Bundle phase runs `npx vite build` directly (ohm is built by Check, SVGs by Render — no double render, and version injection is a vite plugin so it is not lost).
- `build/ciDurationReport.mjs` (+ `.d.mts`): the report logic moves out of inline shell into a pure, unit-tested `buildDurationReport()`. `budgetedTotal = check + bundle + unit` (render excluded); soft-budget breaches emit a `::warning::` but never fail the job (UNIT_BUDGET 120s, TOTAL_BUDGET 360s).
- `src/test/ciDurationReport.test.ts`: 15 unit tests — the render-excluded property, soft-budget boundaries (== budget vs +1 on both thresholds), and `toInt` coercion (numeric strings parsed; non-integer / negative / null / non-numeric → 0).

## Notes

Builds on #421 (the SVG cache that makes the render ~0s on a hit) and is a sibling of #423. Vera gate passed (cycle 1, 2 passes — 5 findings: 3 addressed, 2 deferred — the `doc/CI.md` four-phase doc update to the post-merge doc sweep, and a missing-vs-zero telemetry hardening filed as #427).

Fixes #422

- Claude


wainwright1000:
Withdrawing this PR, not rejecting the work. The #422 branch is being rebased and squashed clean, and the CI documentation that was missing (tracked separately as #431) is being folded in as a docs commit.

The completed branch is parked (`pr park`) and will be resubmitted as a fresh PR once we resume publishing. Nothing here is blocked on review; this is a tidy-up before resubmission.

- Claude

Issue #318, opened 2026-05-21
build: remove generated SVGs from git tracking (prerequisites first)


wainwright1000:
## Background

Ticket #94 asked for SVG generation to be integrated into the build pipeline. PR #271 implemented this and removed all 32 tracked SVGs from git, adding `src/scores/` to `.gitignore`.

## What went wrong

Commit `db0fb6e` reverted the SVG deletion because two critical paths broke:

1. **Vitest** — dynamic imports of `../scores/...?raw` fail when the SVG files are absent. The test suite cannot run without pre-built SVGs.
2. **Netlify** — deploy previews run `npm run build`, but Netlify's build image does not include LilyPond. Removing the SVGs would break every deploy preview and production build.

The revert was correct: a merged PR that breaks deployability must be reverted immediately. The error was in the merge itself — the PR was merged without verifying that the system would work without tracked SVGs.

## Root cause

Prerequisite work was skipped. Before generated files can be removed from version control, every consumer of those files must be able to produce them. Two consumers were missed.

## What needs to happen next

Three separate changes, in this order:

### 1. Make tests resilient to missing SVGs

Options:
- Run `build:scores` as a test-setup step (slow but correct)
- Mock the dynamic `?raw` imports in tests that do not need real SVG content
- Commit a minimal set of fixture SVGs for tests only

### 2. Make Netlify builds self-sufficient

Options:
- Install LilyPond in the Netlify build image (via `apt-get` or a custom build plugin)
- Generate a fallback set of SVGs in CI that Netlify can use
- Make `build:scores` optional in the Netlify build command

### 3. Remove SVGs from git tracking

Once prerequisites 1 and 2 are satisfied, this becomes a trivial change. Add `src/scores/` to `.gitignore` and delete the tracked files.

## Acceptance criteria

- [ ] Tests pass on a clean clone without any SVG files present
- [ ] Netlify deploy previews build successfully without tracked SVGs
- [ ] `npm run build` on a machine with LilyPond generates fresh SVGs
- [ ] `src/scores/` is in `.gitignore`
- [ ] Tracked SVG files are removed from the git index

## Related

- #94 — original ticket (closed — not fully resolved)
- #271 — PR that implemented and then reverted the SVG removal
- `wiki/refactor-svg-scores.md` — group report on SVG score files

wainwright1000:
# VERA-318 Original Report (cycle 1)

Mode: work (Vera ran during initial development, before PR open)
Cycle: 1
[...]

PR #430, opened 2026-06-01
fix: strengthen --skip-if-missing canary to require both notations (#424)

### LINKED ISSUE #424: Remove dead --skip-if-missing flag from buildScores.mjs

wainwright1000:
## Summary

Strengthens `buildScores.mjs`'s `--skip-if-missing` canary from a single probe to **both notations**, so a partially-restored `src/scores/` tree (one notation present, the other missing) is rejected with a loud build failure instead of silently shipping incomplete scores.

## Why

`--skip-if-missing` skips the LilyPond install and uses pre-built SVGs when a canary is present. With #421/#423's cache now restoring `src/scores/` on CI and Netlify, a *partial* restore (e.g. `modern/` present but `early/` missing) would pass the old single-canary check and ship a site with missing parts — the #318/#271 failure mode at sub-tree granularity.

## What changed

- `build/buildScores.mjs`: extracts a testable `canaryCheck(version, root=".")` that probes **both** `modern` and `early` `Choir I A.svg` and returns a discriminated `{ok:true} | {ok:false, missing}`. `checkLilypond` fails loud (exit 1, naming the missing path) when either is absent.
- `build/buildScores.d.mts`: declares `canaryCheck`.
- `src/test/buildScores.test.ts`: 5 unit tests (both present, each missing, both missing, OUP edition) using real files in temp dirs.
- `src/test/integration/buildScores.test.ts`: the skip-success case now requires both canaries; a new partial-restore case asserts the build fails when only one notation is present.

## Scope

Still canary-scope (one choir per notation), not a full per-choir inventory — a tree missing a *non-canary* choir within a present notation still passes. That residual is strictly smaller than before, and full inventory counting is explicitly out of scope, deferred to a future ticket.

Vera gate passed (cycle 1, one pass — no critical or important findings).

Fixes #424

- Claude


wainwright1000:
Parking this until the build-cost batch window.

This PR touches `build/buildScores.mjs`, which is a score-cache input, so merging
it busts the SVG cache and forces a full LilyPond rebuild across CI, preview and
e2e. To avoid that cost while it is not urgent, I am closing the PR. The branch
is preserved and the work is unchanged.

It is tracked under the new `expensive` label and will be reopened (or re-pushed
and rebuilt) as part of a batch of cache-touching changes in the next low-usage
window — see the build-resource monitor; revisit around 23 June.

Nothing needed from you; this is purely build-cost management.

- Vera

PR #455, opened 2026-06-02
fix: move build-resource monitor cron off the top of the hour

### LINKED ISSUE #454: Build-resource monitor scheduled run does not fire (top-of-hour cron throttling)
## Description

The daily build-resource monitor workflow (`.github/workflows/monitor-resources.yml`) is scheduled with `cron: "0 8 * * *"`. On 2026-06-02 the scheduled run did not fire. GitHub Actions is known to throttle and silently drop scheduled workflows that fire on the hour due to platform-wide contention. Additionally, `0 8` runs at 08:00 UTC (09:00 BST), which is one hour later than the intended local morning time.

## Reachability

N/A — this is a tooling/process defect. The symptom is observable only in the GitHub Actions tab: the expected daily run is absent.

## Current state

`.github/workflows/monitor-resources.yml`, line 6.

```yaml
  schedule:
    - cron: "0 8 * * *"
```

The workflow also supports `workflow_dispatch` for manual triggering.

## Root cause

1. **Top-of-hour throttling.** GitHub Actions schedules millions of workflows across the platform at exact hour boundaries. Scheduled jobs on `0 * * * *` are frequently delayed or silently dropped, especially for newly created schedules.
2. **Timezone mismatch.** GitHub Actions cron expressions are evaluated in UTC. `0 8 * * *` is 08:00 UTC, which is 09:00 BST during daylight saving time. The intended local time is 08:00.

## Scope

**In scope:**
- Change the cron expression to an off-hour UTC time that maps closer to 08:00 local.
- Add a clarifying comment above the cron line documenting the UTC time and local equivalent.

**Out of scope:**
- Changes to the monitor script (`.github/scripts/monitor-resources.mjs`).
- Threshold adjustments or new notification channels.
- GitHub Actions timezone support (not available).

## Recommended fix

1. Open `.github/workflows/monitor-resources.yml`.
2. Replace the `schedule` block:

   ```yaml
   schedule:
     # 07:30 UTC — 07:30 GMT / 08:30 BST. Off-hour to avoid GitHub top-of-hour throttling.
     - cron: "30 7 * * *"
   ```

3. Commit and push.

## Test plan

1. Trigger the workflow manually via `workflow_dispatch` to confirm the workflow file is valid and the job still executes successfully.
2. Wait for the next scheduled run time and verify a run appears in the GitHub Actions tab with the correct timestamp.
3. Confirm the Telegram message is received at approximately 08:30 local time (BST).

## Dependencies

None.

## Open questions

None.

## Follow-up items

None.

wainwright1000:
Moves the daily build-resource monitor's schedule off the top of the hour.

`0 8 * * *` (08:00 UTC = 09:00 BST) was prone to GitHub Actions' well-documented top-of-hour scheduling throttling — scheduled runs on the hour are delayed or silently dropped under platform-wide contention — and on 2026-06-02 the run did not fire. The new `30 7 * * *` (07:30 UTC = 08:30 BST) sits off the hour and closer to the intended 08:00 local time, with a comment documenting the UTC/local mapping.

No code is under test here (the value is consumed by GitHub's scheduler, not by application code), so verification is manual per the spec: a `workflow_dispatch` confirms the file is valid and the job runs, and the next scheduled run can be observed in the Actions tab.

Fixes #454

### DIFF
--- a/.github/workflows/monitor-resources.yml
+++ b/.github/workflows/monitor-resources.yml
@@ -2,7 +2,9 @@ name: Monitor Build Resources
 
 on:
   schedule:
-    - cron: "0 8 * * *"
+    # 07:30 UTC (07:30 GMT / 08:30 BST). Off-hour to dodge GitHub
+    # top-of-hour scheduling throttling; see #454.
+    - cron: "30 7 * * *"
   workflow_dispatch:
 
 permissions:


wainwmr (approved):
All good

PR #457, opened 2026-06-02
fix: resolve port offset from a module-relative .worktree-offset file

### LINKED ISSUE #456: worktree-ports: read per-worktree offset from a module-relative .worktree-offset file
﻿**Type:** tech debt
**Area:** Tooling
**Status:** Todo
**Difficulty:** S

**Source:** worktree-ports.ts

**Description:**

worktree-ports.ts resolves the per-worktree dev/preview port offset from
process.env.SPEM_PORT_OFFSET, defaulting to . In a multi-worktree
development setup (several checkouts of the repo running in parallel), that
environment variable is not reliably delivered. The offset is needed at
config-eval time by both ite.config.ts and playwright.config.ts, but the
processes that load them are frequently spawned programmatically rather than
from an interactive shell that sourced a profile or dotenv file, so an
env-based value often never arrives and parallel servers collide on 5173/4173.

The reliable signal in that setup is the file's own location on disk, not the
process environment. Add an environment-independent fallback: if a local
.worktree-offset file sits alongside worktree-ports.ts, read the offset
from it. Resolve the file **relative to the module's own location**
(import.meta.url), not process.cwd(), so it is correct regardless of where
the process was invoked from or how its shell was started.

Proposed precedence (confirm during specification): process.env.SPEM_PORT_OFFSET
if set, else the .worktree-offset file if present, else .

This change is name-agnostic and self-defaulting: a checkout with no
.worktree-offset file (e.g. a normal single-tree clone) keeps today's
behaviour exactly (offset ). The file is local and gitignored, holds only a
number, and carries no per-environment topology into the repository.

---

### 2. Description

worktree-ports.ts resolves the per-worktree dev/preview port offset from process.env.SPEM_PORT_OFFSET, defaulting to . In a multi-worktree development setup, that environment variable is not reliably delivered to programmatically spawned processes (IDE task runners, test harnesses, CI scripts), so parallel servers frequently collide on ports 5173/4173.

### 3. Reachability

N/A — tech debt, not a user-visible bug.

### 4. Current state

worktree-ports.ts (lines 36-41):
`	ypescript
export const portOffset = (
  raw: string | undefined = process.env.SPEM_PORT_OFFSET,
): number => {
  const n = Number(raw);
  return Number.isInteger(n) && n >= 0 ? n : 0;
};
`

ite.config.ts (line 36) and playwright.config.ts (line 38) import DEV_PORT and PREVIEW_PORT, which are resolved at module load time from portOffset().

### 5. Root cause

Environment variables are process-local state. When Vite or Playwright are launched from an IDE, a task runner, or another agent process, the parent may not propagate SPEM_PORT_OFFSET, so the child sees undefined and defaults to offset . The file system is the reliable signal in this setup.

### 6. Scope

In scope:
- Add a .worktree-offset file fallback to worktree-ports.ts.
- Update src/test/worktree-ports.test.ts to cover the new helper.
- Add .worktree-offset to .gitignore.

Deferred:
- No changes to ite.config.ts or playwright.config.ts (they already consume worktree-ports.ts correctly).
- No changes to the existing env-var precedence.

### 7. Recommended fix

In worktree-ports.ts:

1. Import 
eadFileSync from s, dirname and 
esolve from path, and ileURLToPath from 
ode:url.
2. Add a small pure helper 
eadWorktreeOffset(metaUrl?: string):
   `	ypescript
   export function readWorktreeOffset(metaUrl: string = import.meta.url): string | undefined {
     try {
       const dir = dirname(fileURLToPath(metaUrl));
       const content = readFileSync(resolve(dir, ".worktree-offset"), "utf-8").trim();
       return content;
     } catch {
       return undefined;
     }
   }
   `
3. Update portOffset to use the helper as the default:
   `	ypescript
   export const portOffset = (
     raw: string | undefined = process.env.SPEM_PORT_OFFSET ?? readWorktreeOffset(),
   ): number => {
     const n = Number(raw);
     return Number.isInteger(n) && n >= 0 ? n : 0;
   };
   `

4. Add .worktree-offset to .gitignore (after line 2 .vscode).

### 8. Test plan

Extend src/test/worktree-ports.test.ts:

- **Env var beats file:** With SPEM_PORT_OFFSET=200 set, portOffset() returns 200 even if .worktree-offset exists.
- **File fallback:** With SPEM_PORT_OFFSET unset and a .worktree-offset file containing 300, portOffset() returns 300.
- **Absent file:** With SPEM_PORT_OFFSET unset and no .worktree-offset, portOffset() returns .
- **Malformed file:** With SPEM_PORT_OFFSET unset and .worktree-offset containing bc, portOffset() returns .
- **Empty file:** With SPEM_PORT_OFFSET unset and .worktree-offset empty, portOffset() returns .
- **Module-relative resolution:** 
eadWorktreeOffset with a custom metaUrl pointing to a temp directory reads .worktree-offset from that directory, not process.cwd().

To test 
eadWorktreeOffset in isolation, the helper is exported with an optional metaUrl parameter. Tests pass a ile:// URL pointing to a temp directory containing the test file. Remove or update the existing 
eturns 0 when SPEM_PORT_OFFSET is unset test, as portOffset(undefined) now triggers the full fallback chain and its result depends on the test environment.

### 9. Dependencies

None.

### 10. Open questions

None.

### 11. Follow-up items

None.

wainwright1000:
## Summary

`worktree-ports.ts` resolved the per-worktree dev/preview port offset only
from `process.env.SPEM_PORT_OFFSET`. That variable is not reliably delivered
to programmatically-spawned processes (IDE task runners, test harnesses), so
parallel worktrees could collide on 5173/4173.

This adds a fallback: when `SPEM_PORT_OFFSET` is unset, read the offset from a
gitignored `.worktree-offset` file resolved **relative to the module's own
location** (`import.meta.url`), so it is correct regardless of where the
process was invoked from or how its shell was started.

Precedence: `SPEM_PORT_OFFSET` (if set) -> `.worktree-offset` file (if present)
-> `0`.

Name-agnostic and self-defaulting: a checkout with no `.worktree-offset` file
(a normal clone, a fork, CI, the main checkout) keeps today's behaviour exactly
(offset 0). The file is gitignored and holds only a number, so no
per-environment topology enters the repository.

## Changes

- `worktree-ports.ts`: add a `readWorktreeOffset()` helper (module-relative,
  never throws); `portOffset` uses it as the fallback after the env var.
- `src/test/worktree-ports.test.ts`: cover the helper (file present / absent).
- `.gitignore`: ignore `.worktree-offset`.

## Testing

- `npm test` (254 passed), `npm run check` (format / lint / types / deps clean),
  `npm run build` all green.
- e2e not run locally: without a `.worktree-offset` file the new path is
  inactive, so behaviour in CI is byte-identical (offset 0); CI runs e2e on this
  PR regardless.

Fixes #456

- Claude

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,5 +1,6 @@
 .DS_Store
 .vscode
+.worktree-offset
 node_modules
 dist
 *.ipynb
--- a/src/test/worktree-ports.test.ts
+++ b/src/test/worktree-ports.test.ts
@@ -1,5 +1,14 @@
 import { describe, it, expect } from "vitest";
-import { portOffset, DEV_PORT, PREVIEW_PORT } from "../../worktree-ports.ts";
+import { mkdtempSync, writeFileSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { pathToFileURL } from "node:url";
+import {
+  portOffset,
+  DEV_PORT,
+  PREVIEW_PORT,
+  readWorktreeOffset,
+} from "../../worktree-ports.ts";
 
 describe("worktree-ports", () => {
   it("DEV_PORT is exactly 1000 above PREVIEW_PORT (single shared offset)", () => {
@@ -37,5 +46,28 @@ describe("worktree-ports", () => {
     expect(portOffset("1OO")).toBe(0); // letter O, not zeros
     expect(portOffset("-100")).toBe(0);
     expect(portOffset("10.5")).toBe(0);
+  });
+});
+
+describe("readWorktreeOffset", () => {
+  it("reads the trimmed offset from a .worktree-offset file beside the module", () => {
+    const dir = mkdtempSync(join(tmpdir(), "wt-offset-"));
+    try {
+      writeFileSync(join(dir, ".worktree-offset"), "200\n");
+      const moduleUrl = pathToFileURL(join(dir, "worktree-ports.ts")).href;
+      expect(readWorktreeOffset(moduleUrl)).toBe("200");
+    } finally {
+      rmSync(dir, { recursive: true, force: true });
+    }
+  });
+
+  it("returns undefined when no .worktree-offset file sits beside the module", () => {
+    const dir = mkdtempSync(join(tmpdir(), "wt-offset-"));
+    try {
+      const moduleUrl = pathToFileURL(join(dir, "worktree-ports.ts")).href;
+      expect(readWorktreeOffset(moduleUrl)).toBeUndefined();
+    } finally {
+      rmSync(dir, { recursive: true, force: true });
+    }
   });
 });
--- a/worktree-ports.ts
+++ b/worktree-ports.ts
@@ -1,14 +1,22 @@
 // Per-worktree dev/preview port offset, so multiple agent worktrees can
 // run their dev/preview servers at once without colliding.
 //
-// Each worktree declares its own offset via the SPEM_PORT_OFFSET
-// environment variable, set once in that worktree (its launch script or
-// a gitignored .env.local). Anything that does NOT set it — CI, forks, a
-// fresh clone, the main checkout — gets offset 0 and the default ports,
-// so config evaluation can never fail on an unrecognised checkout
-// directory. The offset is the worktree's own fact to declare; we do not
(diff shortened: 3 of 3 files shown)


wainwmr (approved):
OK

PR #480, opened 2026-06-04
feature: expose version to runtime and manifest (#232)

### LINKED ISSUE #232: Expose build version to runtime and PWA manifest
## Summary

The build pipeline already injects `package.json` version into `index.html` at build time via a Vite plugin. This works well for the footer display. Two small extensions would improve debuggability and PWA hygiene.

## Proposed changes

1. **Expose version to the runtime.** Add the current version to `src/ts/config.ts` (or `src/ts/common.ts`) so the application can reference it from TypeScript. This could be done by importing `package.json` in the relevant module, or by having the Vite plugin inject a `window.__SPEM_VERSION__` global, or by adding a `define` block in `vite.config.ts`.

2. **Add version to `site.webmanifest`.** Include a `version` field (or update an existing one) in `public/site.webmanifest` so the PWA manifest reflects the current release. This helps with cache management and update prompts on installed devices.

## Benefits

- Runtime access makes it trivial to include the version in console logs, bug-report payloads, or a future "about" dialog. When a user reports a problem, you can ask them to check the version without relying on the visible footer alone.
- Manifest version gives the browser a clear signal when the PWA has changed, reducing stale-cache issues for users who have installed the site.

## Scope

XS. No user-facing behaviour changes unless you choose to surface the version somewhere new. The build pipeline already has the injection mechanism; this is reusing it for two additional targets.

---

### Description

The build pipeline injects `package.json` version into `index.html` at build time, but the version is not accessible from TypeScript at runtime and is not present in the PWA manifest.

### Current state

- `vite.config.ts` (lines 7--9, 22--23, 52--63): reads `package.json`, computes `versionWithBranch`, and replaces `%VERSION%` in `index.html` via `transformIndexHtml`.
- `src/ts/config.ts`: exports a static config object. No version field.
- `public/site.webmanifest`: copied verbatim to `dist/` during build. Contains no version field.
- `tsconfig.json`: `resolveJsonModule: true` is already enabled.

### Root cause

N/A — feature request.

### Scope

**In scope:**
- Export version from `src/ts/config.ts` so it is importable at runtime.
- Inject version into `dist/site.webmanifest` at build time.

**Deferred:**
- Surfacing version in the UI beyond the existing footer.
- Service worker cache management.
- Bug-report payloads.

### Recommended fix

**1. Runtime exposure**

In `src/ts/config.ts`:

```ts
import pkg from "../../package.json";
```

Add `version` to the exported config object:

```ts
export default {
  // ... existing fields ...
  version: pkg.version,
};
```

This is the minimal approach; `resolveJsonModule` is already enabled in `tsconfig.json`.

**2. Manifest injection**

In `vite.config.ts`, extend the plugin list with a `writeBundle` hook:

```ts
{
  name: "manifest-version",
  writeBundle() {
    const manifestPath = resolve(__dirname, "dist", "site.webmanifest");
    const manifest = JSON.parse(readFileSync(manifestPath, "utf-8"));
    manifest.version = versionWithBranch;
    writeFileSync(manifestPath, JSON.stringify(manifest, null, 2) + "\n");
  },
}
```

Import `writeFileSync` from `fs` at the top of `vite.config.ts`.

### Test plan

1. **Unit test**: verify `config.version` is a non-empty string matching `/^\d+\.\d+\.\d+/`.
2. **Build verification**: run `npm run build`, assert `dist/site.webmanifest` contains a `version` field equal to `package.json` version (or `version-branch` when not on `main`).
3. **Manual**: open browser dev console, import `config`, verify `config.version` is defined.

### Dependencies

None.

### Open questions

The Web App Manifest spec does not define a standard `version` member. The `version` field is non-standard and may be ignored by browsers. If Mark prefers a standards-compliant cache-busting signal, consider updating `name` to include the version (for example, `Spem Player 2.2.0`) or using the `id` field instead.

wainwmr:
Fixes #232

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -24,6 +24,8 @@ MusicCanvasWatcher.define("music-canvas-watcher");
 MusicControls.define("music-controls");
 MusicScore.define("music-score");
 
+console.log(`Spem Player ${config.version}`);
+
 const container = document.querySelector(".split-container");
 const score = document.querySelector("music-score") as MusicScore;
 const splitter = document.querySelector(".splitter") as HTMLDivElement;
--- a/src/test/config.test.ts
+++ b/src/test/config.test.ts
@@ -29,4 +29,8 @@ describe("config", () => {
       expect(arr.length).toBe(config.bartime[i].length);
     });
   });
+
+  it("exposes version from package.json", () => {
+    expect(config.version).toMatch(/^\d+\.\d+\.\d+/);
+  });
 });
--- a/src/ts/config.ts
+++ b/src/ts/config.ts
@@ -1,7 +1,10 @@
 // Copyright (c) 2024-2026 Mark Wainwright
 // SPDX-License-Identifier: MIT
 
+import pkg from "../../package.json";
+
 export default {
+  version: pkg.version,
   parts: ["Soprano", "Alto", "Tenor", "Baritone", "Bass"],
   scores: ["modern", "early"],
   audio_prefix: "/audio/",
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -1,6 +1,6 @@
 import { defineConfig } from "vitest/config";
 import commonjs from "vite-plugin-commonjs";
-import { readFileSync } from "fs";
+import { readFileSync, writeFileSync } from "fs";
 import { resolve } from "path";
 import { execSync } from "child_process";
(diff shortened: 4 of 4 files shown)


wainwmr (commented):
All good

PR #484, opened 2026-06-06
fix: parse ?dark=0 and ?dark=false as light mode (#236)

### LINKED ISSUE #236: ?dark=0 URL parameter has no effect; page always loads in dark mode
A singer wants to share a link that opens in light mode. They append `?dark=0` to the URL. The page still opens in dark mode.

## Steps to Reproduce

1. Open the app with `?dark=0`.
2. Observe that the page is still in dark mode.
3. Open the browser console and check `document.body.classList.contains("light-theme")`. It is false.

## Expected result

`?dark=0` should open the page in light mode.

## Actual result

The `dark` parameter is parsed but its value is ignored; `dark` is always set to `true`.

## Current state

- `index.ts`, `parseURL()`, lines 167-169.
- `} else if (parm[0] == "dark") { dark = true; }`
- The value of `parm[1]` is never read.

## Root cause

The code sets `dark = true` unconditionally when the `dark` parameter is present.

## Scope

- In scope: Fix `dark` parameter parsing.
- Out of scope: Changing the default mode (already dark).

## Recommended fix

Change line 169 from:

```typescript
} else if (parm[0] == "dark") {
    dark = true;
}
```

to:

```typescript
} else if (parm[0] == "dark") {
    dark = parm[1] !== "false" && parm[1] !== "0";
}
```

## Test plan

1. Open `?dark=0`. Assert `current.viewmode` is `"light"`.
2. Open `?dark=false`. Assert `current.viewmode` is `"light"`.
3. Open `?dark=1`. Assert `current.viewmode` is `"dark"`.
4. Open with no dark parameter. Assert `current.viewmode` is `"dark"`.

## Dependencies

None.

wainwmr:
Fixes #236

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "spem-player",
   "private": true,
-  "version": "2.5.2",
+  "version": "2.5.3",
   "description": "Helping singers learn Thomas Tallis Spem in alium",
   "type": "module",
   "dependencies": {
--- a/src/test/url.test.ts
+++ b/src/test/url.test.ts
@@ -37,4 +37,16 @@ describe("parseURLSearch", () => {
     // behaviour; a guard against NaN is tracked in the follow-up ticket.
     expect(Number.isNaN(parseURLSearch("?choir").choir)).toBe(true);
   });
+
+  it("parses ?dark=0 as light mode (#236)", () => {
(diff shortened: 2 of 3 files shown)


wainwmr (commented):
All good

PR #485, opened 2026-06-06
fix: remove and re-add all canvas listeners on disconnect/reconnect (#203)

### LINKED ISSUE #203: disconnectedCallback leaves canvas listeners and breaks reconnect
# disconnectedCallback leaves canvas listeners and breaks reconnect

## Description

`MusicCanvas` adds six event listeners during initialisation (`#init()` at lines 78-95): `click` and `mousemove` on the inner `<canvas>`, and `touchstart` on the canvas, plus `touchmove`, `touchend`, and `wheel` on the `MusicCanvas` element itself. `disconnectedCallback()` removes only the `wheel` listener. The five canvas listeners remain attached, leaking memory. If the element is ever removed from the DOM and reinserted (for example, by a framework or during dynamic view changes), `#init()` early-returns because `this.canvas != null`, so the `wheel` listener is never re-added.

## Steps to Reproduce

1. Load the page.
2. Execute `document.querySelector('music-canvas').remove()` in the console.
3. Execute `document.body.appendChild(document.querySelector('music-canvas'))`.
4. Attempt to scroll with the mouse wheel over the canvas.
5. Observe that vertical scrolling is no longer prevented.

## Expected result

All event listeners added in `#init()` should be removed when the element is disconnected, and restored when it is reconnected.

## Actual result

Canvas listeners leak; wheel listener is missing after reconnect.

## Current state

- File: `src/ts/MusicCanvas.ts`, lines 39-42 (`disconnectedCallback`) and 78-95 (`#init`).
- `disconnectedCallback()` calls `this.removeEventListener("wheel", this.#preventVerticalScroll)`.
- The canvas listeners use `.bind(this)` (line 78) or inline arrow functions (lines 80-82), so their references are lost and cannot be removed.
- `#init()` returns early at line 72 if `this.canvas != null`, preventing re-addition of any listener.

## Root cause

Two problems:
1. Listener references are not stored, making removal impossible.
2. The early-return guard in `#init()` assumes initialisation is a one-time operation, but custom elements can be disconnected and reconnected.

## Scope

In scope: storing listener references, removing all listeners in `disconnectedCallback()`, and ensuring `#init()` re-adds listeners on reconnect.
Out of scope: rewriting the event handling architecture.

## Recommended fix

1. Store bound listener references as private fields:
   ```typescript
   #boundCanvasClicked = this.#canvasClicked.bind(this);
   #boundCanvasHovered = this.#canvasHovered.bind(this);
   #boundTouchStarted = this.#touchStarted.bind(this);
   ```
2. In `#init()`, use the stored references when calling `addEventListener`.
3. In `disconnectedCallback()`, remove all six listeners using the stored references.
4. Change the early-return guard in `#init()` to re-add listeners when reconnecting:
   ```typescript
   if (this.canvas != null) {
     this.addEventListener("wheel", this.#preventVerticalScroll, { passive: false });
     return;
   }
   ```
   Or, more robustly, always remove and re-add listeners at the top of `#init()`.

## Test plan

1. Instantiate `MusicCanvas` in jsdom.
2. Spy on `addEventListener` and `removeEventListener` for both the element and its canvas.
3. Remove the element from the DOM.
4. Assert that `removeEventListener` was called for `click`, `mousemove`, `touchstart`, `touchmove`, `touchend`, and `wheel`.
5. Reinsert the element.
6. Assert that `addEventListener` was called for all six listeners again.

## Dependencies

None.

wainwmr:
Fixes #203

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "spem-player",
   "private": true,
-  "version": "2.5.3",
+  "version": "2.5.4",
   "description": "Helping singers learn Thomas Tallis Spem in alium",
   "type": "module",
   "dependencies": {
--- a/src/test/canvas.test.ts
+++ b/src/test/canvas.test.ts
@@ -708,4 +708,80 @@ describe("MusicCanvas custom element", () => {
 
     freshCanvas.remove();
   });
+
+  it("removes all event listeners on disconnect and re-adds them on reconnect (#203)", async () => {
+    const freshCanvas = document.createElement("music-canvas") as MusicCanvas;
+    document.body.appendChild(freshCanvas);
+    await new Promise((r) => setTimeout(r, 500));
+
+    const innerCanvas = freshCanvas.querySelector("canvas")!;
+    const addCanvasSpy = vi.spyOn(innerCanvas, "addEventListener");
+    const removeCanvasSpy = vi.spyOn(innerCanvas, "removeEventListener");
+    const addElemSpy = vi.spyOn(freshCanvas, "addEventListener");
+    const removeElemSpy = vi.spyOn(freshCanvas, "removeEventListener");
+
+    // Clear calls from initialisation so we only measure disconnect/reconnect
+    addCanvasSpy.mockClear();
+    addElemSpy.mockClear();
+    removeCanvasSpy.mockClear();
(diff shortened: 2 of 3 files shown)


wainwmr (commented):
All good

PR #488, opened 2026-06-06
fix: allow ArrowUp and ArrowDown to repeat when held (#417)

### LINKED ISSUE #417: Holding ArrowUp or ArrowDown only changes choir once
## 2. Description

When a user presses and holds the Up or Down arrow key, the selected choir changes once and then stops. The expected behaviour is continuous cycling through all eight choirs at the OS key-repeat rate, matching the behaviour of the Left and Right arrow keys (which advance through bars continuously when held).

This is a regression from the repeat-suppression fix for #240. The suppression itself is correct — it prevents held keys from strobing state changes at 30 Hz — but the exemption list is too narrow.

## 3. Reachability

1. Open Spem Player in any browser.
2. Press and hold the `ArrowUp` or `ArrowDown` key.

Result: the choir changes once. Expected: the choir cycles through all eight choirs while the key remains held.

## 4. Current state

`index.ts`

Line 251:
```typescript
const REPEAT_EXEMPT_KEYS = new Set(["ArrowLeft", "ArrowRight"]);
```

Line 318:
```typescript
if (e.repeat && !REPEAT_EXEMPT_KEYS.has(e.code)) {
  return;
}
```

Lines 381-390:
```typescript
case "ArrowDown":
  setChoir(
    current.choir >= config.choirs[0].length - 1 ? 0 : current.choir + 1
  );
  break;
case "ArrowUp":
  setChoir(
    current.choir <= 0 ? config.choirs[0].length - 1 : current.choir - 1
  );
  break;
```

`ArrowDown` and `ArrowUp` are both handled by the same switch, but because they are absent from `REPEAT_EXEMPT_KEYS`, the `e.repeat` guard at line 318 returns early on every repeat event after the first.

## 5. Root cause

The `REPEAT_EXEMPT_KEYS` set was introduced in #240 to exempt only `ArrowLeft` and `ArrowRight` from repeat suppression. At that time, the only gesture that benefited from repetition was bar seeking. Choir cycling via `ArrowUp`/`ArrowDown` was not considered, so those keys were left in the default-deny set.

## 6. Scope

**In scope:**
- Add `ArrowUp` and `ArrowDown` to `REPEAT_EXEMPT_KEYS`.
- Add keyboard test verifying that repeated `ArrowUp`/`ArrowDown` events change the choir more than once.

**Out of scope:**
- Changing the repeat-suppression mechanism itself.
- Modifying `SCROLL_KEYS` or `preventDefault` logic (both are correct).
- Any other keyboard handler changes.

## 7. Recommended fix

In `index.ts`, line 251, change:

```typescript
const REPEAT_EXEMPT_KEYS = new Set(["ArrowLeft", "ArrowRight"]);
```

to:

```typescript
const REPEAT_EXEMPT_KEYS = new Set(["ArrowLeft", "ArrowRight", "ArrowUp", "ArrowDown"]);
```

The comment block above the set (lines 244-250) already describes the rationale for exemption. Optionally extend the comment to mention choir cycling, but this is not required.

## 8. Test plan

Extend `src/test/keyboard.test.ts`.

Add two test cases:

1. **Repeated ArrowDown changes choir multiple times**
   - Dispatch three consecutive `KeyboardEvent` objects with `code: "ArrowDown"`, `repeat: true`.
   - Assert that the `music-controls-changed` event (or the `choir` attribute on the controls element) reflects three distinct choir values.

2. **Repeated ArrowUp changes choir multiple times**
   - Same pattern as above, with `code: "ArrowUp"`.
   - Assert backward cycling through choirs.

These tests follow the existing pattern in `keyboard.test.ts` for testing repeated key behaviour (see the `e.repeat` tests added in #240).

## 9. Dependencies

- #240 — the repeat-suppression fix that introduced `REPEAT_EXEMPT_KEYS`. Understanding that change is helpful but not required.
- #10 — the iPad scroll-fix that motivated `SCROLL_KEYS` and `preventDefault`. No direct dependency.

## 10. Open questions

None.

wainwmr:
Fixes #417

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -223,7 +223,12 @@ const SCROLL_KEYS = new Set([
 // *useful*, not just *noisy*. ArrowLeft/ArrowRight are exempt
 // because holding them is the fast-seek gesture (advance through
 // bars while held).
-const REPEAT_EXEMPT_KEYS = new Set(["ArrowLeft", "ArrowRight"]);
+const REPEAT_EXEMPT_KEYS = new Set([
+  "ArrowLeft",
+  "ArrowRight",
+  "ArrowUp",
+  "ArrowDown",
+]);
 
 function keyboardTapped(e: KeyboardEvent) {
   if (e === undefined || e.target === null) {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "spem-player",
   "private": true,
-  "version": "2.5.5",
+  "version": "2.5.6",
   "description": "Helping singers learn Thomas Tallis Spem in alium",
   "type": "module",
   "dependencies": {
--- a/src/test/keyboard.test.ts
+++ b/src/test/keyboard.test.ts
@@ -362,7 +362,7 @@ describe("Space bar play/pause", () => {
       expect(controls.isPlaying()).toBe(false);
     });
 
-    it("held ArrowDown does not change choir", async () => {
+    it("held ArrowDown changes choir forward", async () => {
       const controls = document.querySelector(
         "music-controls"
       ) as MusicControls;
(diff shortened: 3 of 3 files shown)


wainwmr (commented):
All good

Issue #504, opened 2026-06-09
config.choirs: add length-parity assertion (per-recording choir dropdown assumes equal-length lists)


wainwright1000:
**Source:** `src/ts/common.ts` (config validation) and `src/ts/MusicControls.ts` `#buildChoirDropdown`.

**Origin:** Vera gate on [#398](https://github.com/wainwmr/spem-player/issues/398) (per-recording choir dropdown). Four reviewers converged on a latent invariant the #398 fix now relies on.

**Description**

#398 made the choir dropdown rebuild per recording (`config.choirs[this.recording]`) and restore the selection via `choirselect.value = String(this.choir)`. `setChoir` clamps `choir` to `config.choirs[0].length - 1` — recording 0's length only. Today both `config.choirs[0]` and `config.choirs[1]` have 8 entries, so every valid `choir` index maps to an existing option after a rebuild and the restore always lands. The #398 fix is correct for the current config, but it now depends on the two choir lists being equal length, and nothing enforces it. If a future recording shipped a shorter choir list, switching to it would set `choirselect.value` to a non-existent option, which the DOM silently coerces to `""` — a wrong-but-silent UI state (dropdown shows the first choir while `this.choir` holds the old, out-of-range index), with no error.

**Recommended fix:** add an import-time length-parity assertion in `common.ts`, matching the file's existing defensive guards (it already throws on `config.recording.length !== 2` and on `bartime`/`barno` parity):

```ts
for (let v = 1; v < config.choirs.length; v++) {
  if (config.choirs[v].length !== config.choirs[0].length) {
    throw new Error(
      `config.choirs[${v}] length ${config.choirs[v].length} != choirs[0] length ${config.choirs[0].length}`
    );
  }
}
```

This fail-fasts at import if the invariant is ever violated, making the existing clamp-to-`choirs[0]` provably correct rather than coincidentally correct. Heavier alternative (declined for #398): clamp `setChoir` against `config.choirs[this.recording].length` and reclamp inside `#buildChoirDropdown`.

**Difficulty:** XS. Latent — not reachable in the current config; a hardening, not a user-facing bug.

- Vera

wainwright1000:
## Re-specification

Delta to the existing specification. Sections not listed are unchanged.

### 3. Root cause

[...]

PR #512, opened 2026-06-09
fix: rebuild choir dropdown when recording changes (#398)

### LINKED ISSUE #398: Controls dropdown shows ALC choir names when CotE recording is selected
When the user switches from ALC to CotE, the choir dropdown in `music-controls` continues to display ALC names ('I A', 'I B', etc.) instead of CotE names ('1', '2', etc.).

## Current state

`src/ts/MusicControls.ts`, `connectedCallback()` (line 65-79):
```typescript
for (var c in config.choirs[0]) {
  const opt = document.createElement("option");
  opt.setAttribute("value", c);
  opt.appendChild(document.createTextNode(config.choirs[0][c]));
  this.choirselect.append(opt);
}
```

The dropdown is built once during `connectedCallback()` using `config.choirs[0]` (ALC names). It is never updated thereafter.

`src/ts/MusicControls.ts`, `setRecording()` (line 290-294):
```typescript
setRecording(v: number | string): void {
  super.setRecording(v);
  if (this.isPlaying()) this.play();
}
```

This override updates the audio file and seeks to the correct time but does not touch the dropdown.

`src/ts/config.ts` defines two naming conventions:
- `config.choirs[0]` = `["I A", "I B", "II A", "II B", "III A", "III B", "IV A", "IV B"]` (ALC)
- `config.choirs[1]` = `["1", "2", "3", "4", "5", "6", "7", "8"]` (CotE)

## Root cause

The dropdown build logic is hardcoded to ALC names and is not re-executed when the recording changes. `setRecording()` does not rebuild or update the dropdown options.

## Scope

**In scope:**
- Rebuild the choir dropdown options when `setRecording()` is called.
- Ensure the current choir selection is preserved across rebuilds.
- Update existing unit tests and add a new test for recording-aware dropdown labels.

**Out of scope:**
- Changing the score SVG labels (tracked in #237).
- Supporting recordings with different choir counts.
- Changing the option values (0-7 indices are correct for both recordings).

## Recommended fix

1. **Extract a helper method** in `MusicControls.ts` to build the choir dropdown:
   ```typescript
   #buildChoirDropdown() {
     if (!this.choirselect) return;
     this.choirselect.innerHTML = "";
     for (var c in config.choirs[this.recording]) {
       const opt = document.createElement("option");
       opt.setAttribute("value", c);
       opt.appendChild(document.createTextNode(config.choirs[this.recording][c]));
       this.choirselect.append(opt);
     }
     this.choirselect.value = String(this.choir);
   }
   ```

2. **Call the helper** from `connectedCallback()` where the dropdown is currently built inline (replace lines 72-77).

3. **Call the helper** from `setRecording()` after `super.setRecording(v)`:
   ```typescript
   setRecording(v: number | string): void {
     super.setRecording(v);
     this.#buildChoirDropdown();
     if (this.isPlaying()) this.play();
   }
   ```

## Test plan

**Unit tests (`src/test/controls.test.ts`):**

1. Extend the existing "connectedCallback renders controls" test to verify that the initial dropdown shows ALC names (default recording is 0).

2. Add a new test: "setRecording updates choir dropdown labels":
   - Create a `MusicControls` element (default recording 0).
   - Assert dropdown options text equals `config.choirs[0]`.
   - Call `setRecording(1)`.
   - Assert dropdown options text equals `config.choirs[1]`.
   - Assert the selected value is still `"0"` (choir index preserved).

3. Add a new test: "setRecording preserves selected choir index":
   - Create a `MusicControls` element.
   - Set `choir` to `3` via `setAttribute` or `setChoir`.
   - Call `setRecording(1)`.
   - Assert `choirselect.value` is `"3"`.
   - Assert the displayed text for option 3 is `config.choirs[1][3]` (i.e., `"4"`).

**Browser verification:**
- Not required. This is a DOM label change that jsdom can verify.

## Dependencies

- [#237](https://github.com/wainwmr/spem-player/issues/237) -- same root cause in `music-score` (score SVG titles). The two tickets can be worked independently.
- [Workbench Item #156](https://github.com/wainwright1000/spem-tools/issues/156) -- investigation that identified the `config.choirs[0]` hardcoding root cause.

## Open questions

None.

wainwright1000:
## Summary

Fixes #398.

The choir dropdown was built once at init from `config.choirs[0]` and never
rebuilt, so selecting the CotE recording (recording = 1) left the dropdown
showing the ALC choir labels.

## Changes

- Extract the dropdown build into `MusicControls.#buildChoirDropdown()`, which
  clears and rebuilds the options from `config.choirs[this.recording]` and
  restores the selected choir. It is called at init and from `setRecording()`,
  so the labels track the active recording.
- Pin the play/pause button with `flex: none`. Rebuilding the dropdown changes
  its width per recording, which reflows the controls row; without the pin,
  flexbox shrank the button's width against its fixed height and the
  `border-radius: 50%` circle rendered as an oval. (Found during UI review; a
  pre-existing layout fragility that this fix would otherwise have exposed.)

## Tests

- Unit: a 1 → 0 recording switch-back test asserting the labels return to
  `config.choirs[0]` (verified red against a neutered fix, then green).
- e2e: the play/pause button stays square (circular) across a recording toggle.

## Notes

- The rebuild relies on every recording's choir list being the same length
  (all are 8 today). The invariant is commented in `#buildChoirDropdown`; a
  length-parity assertion is tracked separately in #504.

- Vera

### DIFF
--- a/e2e/playback-toggle.spec.ts
+++ b/e2e/playback-toggle.spec.ts
@@ -50,4 +50,27 @@ test.describe("Playback keyboard toggle", () => {
     await page.keyboard.press("Enter");
     await expect(pauseIcon).toBeHidden();
   });
+
+  test("play/pause button stays circular when the recording changes (#398)", async ({
+    page,
+  }) => {
+    await page.goto("/");
+
+    const button = page.locator("#playpausebutton");
+    await expect(button).toBeVisible();
+
+    // The button is border-radius:50%, so it only reads as a circle while its
+    // width equals its height. Changing the recording rebuilds the choir
+    // dropdown to a different label width, reflowing the controls row; the
+    // button must not be resized by that reflow.
+    const isSquare = async () => {
+      const box = await button.boundingBox();
+      if (!box) throw new Error("button has no bounding box");
+      expect(Math.abs(box.width - box.height)).toBeLessThanOrEqual(1);
+    };
+
+    await isSquare();
+    await page.locator("#recordingswitch").click();
+    await isSquare();
+  });
 });
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "spem-player",
-  "version": "2.7.0",
+  "version": "2.7.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "spem-player",
-      "version": "2.7.0",
+      "version": "2.7.1",
       "license": "MIT",
       "dependencies": {
         "ohm-js": "^17.2.1"
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "spem-player",
   "private": true,
-  "version": "2.7.0",
+  "version": "2.7.1",
   "description": "Helping singers learn Thomas Tallis Spem in alium",
   "type": "module",
   "dependencies": {
--- a/src/scss/style.scss
+++ b/src/scss/style.scss
@@ -375,6 +375,11 @@ kbd {
       padding: 3px;
       margin: 3px 5px;
 
+      // Pin the button so flexbox cannot resize it when the controls row
+      // reflows (e.g. when the choir dropdown rebuilds to a different
+      // recording's labels). Without this, a shrunk width against the fixed
+      // height renders the border-radius:50% circle as an oval (#398).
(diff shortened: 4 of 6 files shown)


wainwmr (approved):
All good

PR #516, opened 2026-06-09
build: Exclude escapeHtml.ts from depcruise no-orphans rule (#501)

### LINKED ISSUE #501: dependency-cruiser false-positive orphan warning for escapeHtml.ts
## Issue

`npm run check` runs `depcruise src/ts/*.ts index.ts` as part of the quality gate. Since PR #487 (`escapeHtml.ts`), dependency-cruiser emits a `no-orphans` warning:

```
warn no-orphans: src/ts/escapeHtml.ts
```

`escapeHtml.ts` **is** imported — by `vite.config.ts` — but `vite.config.ts` is not in the depcruise command, so dependency-cruiser cannot see the edge and flags the file as orphaned. The warning does not fail the build (exit 0), but it creates noise every time `npm run check` runs.

## Options

**A. Add `vite.config.ts` to the depcruise command**

Change `package.json` from:
```json
"check:deps": "depcruise src/ts/*.ts index.ts"
```
to:
```json
"check:deps": "depcruise src/ts/*.ts index.ts vite.config.ts"
```

This removes the orphan warning because the import from `vite.config.ts` becomes visible. Side effect: it also pulls in `worktree-ports.ts` (imported by `vite.config.ts`), which triggers 5 additional `no-deprecated-core` warnings for Node.js built-in modules (`path`, `fs`, `child_process`, `url`). Those are false positives for build config.

**B. Add an exception to the `no-orphans` rule**

In `.dependency-cruiser.cjs`, exclude `src/ts/escapeHtml.ts` from the orphan check:

```js
{
  name: "no-orphans",
  severity: "warn",
  from: { orphan: true, pathNot: "src/ts/escapeHtml.ts" },
  to: {},
},
```

This silences the warning immediately with no side effects. Risk: it is a one-off special case. If more build-only utilities are added, the exception list grows.

**C. Exclude build config files from `no-deprecated-core` and adopt Option A**

Add `vite.config.ts` and `worktree-ports.ts` to the depcruise set, and update the `no-deprecated-core` rule to ignore Node core imports from build config files. This is the most principled fix but touches more config.

## Bob's recommendation

Bob: Option B is the minimal viable fix. The warning is a false positive caused by a scope mismatch between the cruise set and the build tooling, not by a genuine orphan. Option A trades one warning for five; Option C is principled but over-engineered for a single file. Add the exception now. If a pattern of build-only utilities emerges, revisit with Option C.

## Ask

Which option do you prefer? If B, shall I apply it? If A or C, do you want to handle the config change yourself?

wainwmr:
Fixes #501

## Summary

Applies Option B from the issue: adds a `pathNot` exception for `src/ts/escapeHtml.ts` in the `no-orphans` rule of `.dependency-cruiser.cjs`.

## Problem

`escapeHtml.ts` is imported by `vite.config.ts`, but `vite.config.ts` is not in the depcruise command (`depcruise src/ts/*.ts index.ts`). Dependency-cruiser therefore flags it as orphaned.

## Why Option B

- Option A (add `vite.config.ts` to the cruise set) trades one warning for five new `no-deprecated-core` warnings from Node built-ins.
- Option C (principled exclusion of build config from `no-deprecated-core`) is over-engineered for a single file.
- Option B is a one-line change with no side effects. If a pattern of build-only utilities emerges, revisit with Option C.

## Verification

- `npm run check` before: `warn no-orphans: src/ts/escapeHtml.ts`
- `npm run check` after: `✔ no dependency violations found`
- `npm test`: 280 tests pass
- Warning diff: orphan warning removed, zero new warnings introduced
- No version bump (build/config-only change per work method)

### DIFF
--- a/.dependency-cruiser.cjs
+++ b/.dependency-cruiser.cjs
@@ -10,7 +10,7 @@ module.exports = {
     {
       name: "no-orphans",
       severity: "warn",
-      from: { orphan: true },
+      from: { orphan: true, pathNot: "src/ts/escapeHtml.ts" },
       to: {},
     },
     {


wainwmr (commented):
All good

PR #527, opened 2026-06-10
Cancel stale score loads on rapid choir switching (#391)

### LINKED ISSUE #391: Race condition: rapid choir switching leaves stale SVG reference
## Bug description

When a user rapidly switches choirs (e.g. pressing digit keys quickly), concurrent `setChoir()` calls interleave because each awaits `#loadScore()` without cancellation. The second call overwrites `innerHTML` before the first call finishes `getBars()` and `highlight()`.

## Reproduction

1. Call `setChoir(0)`
2. Immediately call `setChoir(1)` without awaiting the first
3. If the first import resolves after the second, the DOM contains choir 0's SVG but `this.svg` points to choir 1's disconnected SVG element

## Impact

- `this.svg.isConnected === false`
- `this.bars` does not match the DOM
- Click-to-seek uses wrong bar positions
- Highlight system operates on a ghost reference

## Root cause

`setChoir()`'s `this.choir` is shared mutable state. `#loadSvg()` captures `choirName` before the await, but `#loadScore()` reads `this.svg` and sets `this.bars` after the await. If two calls interleave, the second call's DOM update happens while the first is still processing.

## Fix direction

Add cancellation to `#loadScore()`: store an abort token / generation counter on each call and skip DOM manipulation if a newer load has started.

- Kimi

wainwmr:
- Add `#loadGeneration` counter to `MusicScore`
- Increment counter at start of `#loadScore`; abort if a newer load has started when the await resolves
- Allow test SvgLoader to return Promises by awaiting `Promise.resolve()` so tests can simulate network races
- Add unit test verifying stale load aborts when newer request resolves first

Closes #391

### DIFF
--- a/src/test/score.test.ts
+++ b/src/test/score.test.ts
@@ -6,6 +6,7 @@ import {
   expectedPart,
   waitForEvent,
 } from "./helpers";
+import { makeFixtureSvg } from "./fixtureScore";
 
 // Polyfill DOMPoint for jsdom
 if (typeof DOMPoint === "undefined") {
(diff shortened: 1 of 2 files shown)


wainwmr (commented):
All good

PR #535, opened 2026-06-10
refactor: extract validated parseRepo() helper in monitor-resources.mjs (#497)

### LINKED ISSUE #497: Extract validated parseRepo() helper in monitor-resources.mjs
**Type:** tech debt &nbsp;·&nbsp; **Area:** Tooling &nbsp;·&nbsp; **Status:** Todo &nbsp;·&nbsp; **Difficulty:** XS (intended; set at specification)

**Source:** `.github/scripts/monitor-resources.mjs:49` and `:91` — `process.env.GITHUB_REPOSITORY.split("/")` is parsed inline at two call sites (`getGitHubUsage`, `openIssue`) with no validation.

**Origin:** Surfaced by Kimi while running `investigate` on Item [#364](https://github.com/wainwright1000/spem-tools/issues/364) (wainwright1000/spem-tools). A fix was prototyped and then **correctly backed out**, because an investigation is read-only: a source fix must go through `work` on a feature branch, not freehand.

**Description**

`monitor-resources.mjs` reads `process.env.GITHUB_REPOSITORY` and splits it into `owner`/`repo` in two functions without checking the variable is present and well-formed. If `GITHUB_REPOSITORY` is undefined, `.split` throws an opaque `TypeError`; if it lacks a `/`, `repo` is `undefined` and the failure surfaces later as a confusing GitHub API error against a malformed URL.

In the production path the script runs only inside the `monitor-resources` GitHub Actions workflow, where Actions always injects a well-formed `owner/repo` value, so it does not currently fail in CI. The improvement is defensive clarity plus removing the duplicated parse — hence tech debt, not a bug.

**Recommended fix (shape, from the backed-out prototype)**

Extract a `parseRepo()` helper that splits `GITHUB_REPOSITORY`, validates that both `owner` and `repo` are present, and throws a clear error otherwise; call it from both `getGitHubUsage` and `openIssue`.

```diff
@@ async function getNetlifyUsage() {
   };
 }

-async function getGitHubUsage() {
+function parseRepo() {
   const [owner, repo] = process.env.GITHUB_REPOSITORY.split("/");
+  if (!owner || !repo) {
+    throw new Error(
+      `GITHUB_REPOSITORY must be owner/repo format, got: ${process.env.GITHUB_REPOSITORY}`
+    );
+  }
+  return { owner, repo };
+}
+
+async function getGitHubUsage() {
+  const { owner, repo } = parseRepo();
   const now = new Date();
@@ async function openIssue(title, body) {
-  const [owner, repo] = process.env.GITHUB_REPOSITORY.split("/");
+  const { owner, repo } = parseRepo();
```

**Test plan** (deferred to specification): unit-test `parseRepo()` for the valid case, the missing-variable case, and the malformed (no-slash) case.

**Dependencies:** none.

**Implementation route:** Must be implemented via `work` on a feature branch off `origin/main` (State Mutation Rules), not patched freehand. The backed-out prototype above is the target shape, not a commit.

- Vera

wainwright1000:
Extracts a validated `parseRepo()` helper in
`.github/scripts/monitor-resources.mjs`, replacing three inline
`process.env.GITHUB_REPOSITORY.split("/")` sites.

- The old `.split("/")` silently yielded `undefined` owner/repo on a malformed
  `GITHUB_REPOSITORY`; `parseRepo()` throws a clear, named error instead.
- DRYs the three call sites (`getMergedPRCount`, `getGitHubUsage`, `openIssue`).
- Adds tests: valid input plus the undefined / no-slash / empty-string error
  cases.

Passed the Vera gate (five reviewers, no critical or important findings). Folded
the two convergent suggestions: a `@returns`/`@throws` JSDoc on `parseRepo` to
match the file's other exported helpers, and an `after()` hook so the new tests
restore `GITHUB_REPOSITORY` rather than leaking global state.

Fixes #497

- Vera

### DIFF
--- a/.github/scripts/monitor-resources.mjs
+++ b/.github/scripts/monitor-resources.mjs
@@ -53,8 +53,24 @@ async function getNetlifyUsage() {
   };
 }
 
+/**
+ * Parse `owner` and `repo` from the `GITHUB_REPOSITORY` environment variable.
+ *
+ * @returns {{owner: string, repo: string}} Both guaranteed non-empty.
+ * @throws {Error} If `GITHUB_REPOSITORY` is absent or not in `owner/repo` form.
+ */
+export function parseRepo() {
+  const [owner, repo] = process.env.GITHUB_REPOSITORY?.split("/") ?? [];
+  if (!owner || !repo) {
+    throw new Error(
+      `GITHUB_REPOSITORY must be owner/repo format, got: ${process.env.GITHUB_REPOSITORY}`
+    );
+  }
+  return { owner, repo };
+}
+
 async function getMergedPRCount(since) {
-  const [owner, repo] = process.env.GITHUB_REPOSITORY.split("/");
+  const { owner, repo } = parseRepo();
   const data = await api(
     `https://api.github.com/search/issues?q=repo:${owner}/${repo}+is:pr+is:merged+merged:>=${since}&per_page=1`,
     {
@@ -128,7 +144,7 @@ export function getReportingSince(nowMs = Date.now()) {
 }
 
 async function getGitHubUsage() {
-  const [owner, repo] = process.env.GITHUB_REPOSITORY.split("/");
+  const { owner, repo } = parseRepo();
   const now = new Date();
   const start = new Date(
     Date.UTC(now.getUTCFullYear(), now.getUTCMonth(), 1)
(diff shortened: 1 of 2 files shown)


wainwmr (approved):
All good

PR #540, opened 2026-06-10
test: stub virtual:pwa-register in Vitest setup

### LINKED ISSUE #530: PWA virtual:pwa-register import breaks Vitest suites on Windows
## Description

On Windows, `npm test` fails four UI suites at import time — `darkmode`,
`feedback`, `keyboard`, `setbar`. Their individual tests are all skipped, yet
each file errors on import:

```text
TypeError: The argument 'filename' must be a file URL object, file URL string,
or absolute path string. Received 'file:///@vite-plugin-pwa/virtual:pwa-register'
```

The suite is green on macOS and Linux CI, so it merged unseen; only Windows
contributors hit it. A related second symptom in the same teardown:
`TypeError: controls.pause is not a function` (an `afterAll` running against
suites whose `beforeAll` aborted during import).

## Reachability

Developer-facing only. Any contributor running `npm test` on Windows hits this.
CI (Linux) does not catch it.

## Current state

- `src/ts/pwa-update.ts:1` — module-level import:
  `import { registerSW } from "virtual:pwa-register";`
- `index.ts:30` — imports `pwa-update`:
  `import "./src/ts/pwa-update";`
- `src/test/setup.ts` — shared Vitest setup; currently installs only
  `__SPEM_TEST_SVG_LOADER`
- `src/test/darkmode.test.ts:54`, `feedback.test.ts:86`, `keyboard.test.ts:55`,
  `setbar.test.ts:57` — each dynamically imports `../../index.ts` inside
  `beforeAll`
- `src/test/pwa-update-toast.test.ts:6-11` — per-file
  `vi.mock("virtual:pwa-register", ...)` that works correctly

## Root cause

`vite-plugin-pwa` only provides `virtual:pwa-register` at build time, not under
Vitest. When the four test files import `index.ts`, the transitive import of
`pwa-update.ts` pulls in the unresolved virtual module. Vitest hands the id
`file:///@vite-plugin-pwa/virtual:pwa-register` to Node's path layer. POSIX
tolerates the drive-letter-less path; Windows rejects it with the TypeError
above — a POSIX-tolerant / Windows-strict divergence.

The second symptom occurs because the `beforeAll` aborts mid-import, but
`afterAll` still runs. `document.querySelector("music-controls")` may return an
un-upgraded element (from a prior test's DOM or because the custom element was
never defined), so `pause()` is undefined at runtime despite the TypeScript cast.

## Scope

Test-infrastructure only. The app and PWA registration work in the browser. CI
stays green.

Deferred: extracting the duplicated `beforeAll` blocks in the four test files
into a shared helper (noted as shared test setup drift in
`Architecture-Test.md`).

## Recommended fix

Add a global mock to `src/test/setup.ts`:

```ts
import { vi } from "vitest";
vi.mock("virtual:pwa-register", () => ({ registerSW: () => () => {} }));
```

The per-file mock in `pwa-update-toast.test.ts` should stay; Vitest hoists
`vi.mock` and the per-file mock takes precedence where present.

Alternative considered: a `resolve.alias` in `vite.config.ts` pointing to a stub
module. Rejected — the setup-mock is lighter, matches the existing pattern, and
requires no new files.

## Test plan

1. Run `pnpm test:unit` locally and confirm all suites pass (no regression).
2. Confirm `src/test/pwa-update-toast.test.ts` still passes (its per-file mock
   continues to work).
3. On a Windows machine or VM, run `pnpm test:unit` and verify the four
   previously failing suites (`darkmode`, `feedback`, `keyboard`, `setbar`) now
   import without error and their tests run.

## Dependencies

None.

## Open questions

None.

## Follow-up items

None. (Shared test setup drift in the four integration files is noted in
`Architecture-Test.md` but out of scope for this fix.)

## Documentation impact

None expected. `doc/TESTING.md` does not currently mention platform-specific
test behaviour; no change warranted for a single mock addition.

wainwright1000:
Fixes #530.

On Windows, four UI suites (`darkmode`, `feedback`, `keyboard`, `setbar`)
crashed at import because they transitively import `index.ts` →
`src/ts/pwa-update.ts`, which imports `registerSW` from `virtual:pwa-register`
— a module `vite-plugin-pwa` provides only at build time, not under Vitest.
Node rejects the unresolved id `file:///@vite-plugin-pwa/virtual:pwa-register`
on Windows (POSIX tolerates the drive-letter-less path), so CI (Linux) stayed
green and the breakage merged unseen; only Windows contributors hit it.

This adds a global `vi.mock("virtual:pwa-register", …)` stub to the shared
Vitest setup (`src/test/setup.ts`), keeping every suite importable. The
dedicated `pwa-update-toast.test.ts` keeps its own per-file `vi.mock`, which
Vitest hoists and which takes precedence where it needs the registration
callback.

Test-infrastructure only; no production behaviour change. Restores the 61
previously-skipped tests in the four suites; `pnpm run ci` green (281 tests).

### DIFF
--- a/src/test/setup.ts
+++ b/src/test/setup.ts
@@ -1,4 +1,14 @@
+import { vi } from "vitest";
 import { makeFixtureSvg } from "./fixtureScore";
+
+// `virtual:pwa-register` is supplied by vite-plugin-pwa at build time only, not
+// under Vitest. Any suite that transitively imports index.ts (-> pwa-update.ts)
+// would otherwise fail at import: Node's path layer rejects the unresolved id
+// `file:///@vite-plugin-pwa/virtual:pwa-register` on Windows (POSIX tolerates
+// the drive-letter-less path, Windows does not). A global stub keeps every suite
+// importable; pwa-update-toast.test.ts hoists its own vi.mock, which takes
+// precedence where it needs the registration callback. See ticket #530.
+vi.mock("virtual:pwa-register", () => ({ registerSW: () => () => {} }));
 
 // MusicScore.#loadSvg() checks `MusicScore.testSvgLoader` (static,
 // preferred) first and falls back to `globalThis.__SPEM_TEST_SVG_LOADER`.


wainwmr (approved):
OK

Issue #545, opened 2026-06-11
controls.test.ts: falsy-zero guards silently skip choir/part/bar assertions in handleChange


wainwright1000:
**Type:** tooling bug
**Area:** Test

**Source:** `src/test/controls.test.ts` lines 254-256 (`handleChange`)
**First seen:** 2026-06-11, Vera gate finding 541-02 (cycle 1) on ticket #541; verified at branch tip

**Description**

`handleChange` guards its assertions with truthiness checks on the shared expectation variables:

```typescript
if (expectedChoir) expect(d.position.choir).toBe(expectedChoir);
if (expectedPart) expect(d.position.part).toBe(expectedPart);
if (expectedBar) expect(d.position.bar).toBe(expectedBar);
```

`0` is a legal expected value for all three (choir 0, part 0 = Soprano, bar 0), and `expectedChoir = 0` makes the guard falsy, so the assertion is silently skipped. Any test expressing a "expect choir/part/bar 0" expectation through this helper asserts nothing — the same silent-no-op defect family as #541 (which was a missing call parenthesis in the sibling helper).

**Recommended fix**

Replace the truthiness guards with explicit presence checks (`if (expectedChoir !== undefined) ...`), or eliminate the guards entirely via the closure-based expectation redesign already sketched for the shared `expected*` mutable globals (local Item #270 on the Workbench tracks that redesign; this ticket is the minimal in-repo fix and does not depend on it).

**Test plan**

After the fix, a test that sets an expected value of `0` and dispatches an event with a different value must fail. Run the full controls suite for regressions; mutation-style spot check (dispatch position with choir 1 while expecting 0) must be caught.

**Dependencies**

Follow-up to #541. Related: #544 (lint guard for the no-op family).

- Vera

wainwright1000:
## Re-specification

### New finding: a live test bug is hidden by the guard

`controls.test.ts:221` passes `expectedChoir = 0` in the first choir-change test while the attribute is set to `"2"` (line 227). The event detail therefore carries `position.choir = 2`. Because the guard is falsy for `0`, the assertion is silently skipped. Fixing the guard exposes the mismatch: the test expects `0` but receives `2`.

The minimal fix is therefore **two changes**, not one:

1. **Correct the expected value** in the first choir-change test (line 225): change `0` to `2` so the assertion matches the event detail.
2. **Fix the guards** in `handleChange` (lines 254-256): replace truthiness with explicit `!== undefined`.

### Scope revision

[...]

Issue #544, opened 2026-06-11
eslint: no-op expect guard disabled for test files (no-unused-expressions off)


wainwright1000:
**Type:** tech debt
**Area:** Test

**Source:** `eslint.config.js` (test-file override)
**First seen:** 2026-06-11, Vera gate finding 541-01 (cycle 1) on ticket #541

**Description**

The eslint config disables exactly the rule that catches no-op expect chains in the only files that contain them. The test-file override sets `'@typescript-eslint/no-unused-expressions': 'off'` for `src/**/*.test.ts`, while the base config has the rule at `error`. A parenthesis-less matcher such as `expect(x).not.toBeNull` (ticket #541 — a permanent no-op affecting ~12 assertions) is an unused expression and would have been caught at lint time; the off-switch is how it survived.

**Recommended fix**

Either:

- (a) re-enable `@typescript-eslint/no-unused-expressions` for test files — Vitest's `expect` chains always end in a call, so the suite should lint clean (verify with a lint run before committing); or
- (b) add `@vitest/eslint-plugin` with `vitest/valid-expect` and `vitest/expect-expect` — more targeted, and also catches `expect()` with no matcher at all.

Option (b) is the stronger barrier; (a) is dependency-free.

**Test plan**

Lint run must pass on the existing suite; then a deliberate `expect(x).toBeNull` (no call) in a scratch test file must fail lint.

**Dependencies**

Follow-up to #541. Related: #509 (other eslint.config.js work — coordinate to avoid conflicts).

- Vera

wainwright1000:
## Re-specification

### New findings

A lint probe with the test-file override removed produces **exactly one violation** across the entire test suite:

```
src/test/controls.test.ts:151
  error  Expected an assignment or function call and instead saw an expression
         @typescript-eslint/no-unused-expressions
```

That line is the already-tracked #541 bug (`expect(...).not.toBeNull` without call parentheses). Every other test file lints clean. Option (a) is therefore dependency-free and carries no collateral cleanup risk.

### Corrected sequencing

The existing test plan states the lint run must pass on the existing suite after the override is removed. Because `origin/main` still contains the bare `toBeNull` at `controls.test.ts:151` (PR #546 is open but not yet merged), removing the override now would fail lint. This ticket must be sequenced **after #541 merges**, or the config change must be bundled into the #541 PR.

### Scope revision

[...]

PR #546, opened 2026-06-11
fix: add missing call parentheses to toBeNull assertion in controls.test.ts

### LINKED ISSUE #541: controls.test.ts: broken toBeNull assertion always passes
**Type:** tooling bug
**Area:** Test
**Difficulty:** XS

**Source:** `src/test/controls.test.ts` line 151
**First seen:** 2026-06-10, commit `afbf226`

**Description**
The `handleAudioStarted` helper inside `controls.test.ts` contains a broken Vitest assertion:

```typescript
expect((event as CustomEvent).detail).not.toBeNull;
```

The `.not.toBeNull` property is accessed but never called (missing parentheses). In Vitest/Jest, `expect(x).not.toBeNull` evaluates to the function object itself, which is always truthy. The assertion therefore never fails, regardless of whether `detail` is actually `null`.

This helper is used by approximately 12 test assertions across the same file (play, pause, loading, and ended event tests), meaning all of those assertions are currently no-ops.

**Recommended fix**
Replace `.not.toBeNull` with `.not.toBeNull()` or `.not.toBe(null)` at line 151. Re-run the test suite to confirm no regressions.

**Test plan**
1. Run `pnpm run test:unit -- src/test/controls.test.ts` and confirm all 25 tests still pass after the fix.
2. Temporarily introduce a null detail in `MusicControls` and confirm the fixed assertion fails as expected.

**Dependencies**
None.

wainwright1000:
Fixes #541

Adds the missing call parentheses to the `toBeNull` assertion in `handleAudioStarted` (`src/test/controls.test.ts`): `expect((event as CustomEvent).detail).not.toBeNull` evaluated the matcher function object (always truthy), so the assertion — feeding ~12 call sites across the play/pause/loading/ended tests — never failed. It is now live.

Verification: full suite green on the tip (281 tests — activation exposes no latent failure). A mutation test during review (dispatching `detail: null` from `fireEvent`) fails 11/25 controls tests post-fix where pre-fix it would have been largely invisible. A repo-wide sweep found no other parenthesis-less matchers.

Vera gate: cycle 1 clean on the diff itself; two adjacent findings filed as follow-ups — #544 (eslint test-file override disables the rule that catches this no-op class) and #545 (falsy-zero guards in the sibling helper silently skip choir/part/bar-0 assertions). Reports on the ticket.


wainwmr (approved):
OK

PR #547, opened 2026-06-11
test: fix ordering dependency in darkmode.test.ts

### LINKED ISSUE #542: darkmode.test.ts: test ordering dependency hides assertion failure
## Description
`src/test/darkmode.test.ts` contains an ordering dependency between tests: test 3 ("toggles back to dark mode when the darkswitch is clicked again") assumes the DOM was left in light mode by test 2. When test 3 runs in isolation, the suite fails because the body still has the default dark class. This hides genuine regressions and makes the test unreliable in CI and local debugging.

## Current state
- `src/test/darkmode.test.ts:71-85` — tests 2 and 3 both dispatch a click on `#darkswitch` and assert the resulting `light-theme` class state, but neither resets the DOM or body class between tests.
- `beforeAll` constructs the fixture once for the whole describe block; there is no `beforeEach` cleanup.
- Running `pnpm exec vitest run -t "toggles back to dark mode" src/test/darkmode.test.ts` reproduces the failure.

## Root cause
Shared mutable DOM state (`document.body.classList`) is modified by each test without a reset, so later tests depend on the side effects of earlier ones.

## Scope
In scope:
- Reset `document.body.classList` and relevant element state before each test so every test is independent.
- Keep the existing coverage for dark mode toggling and icon updates.

Out of scope:
- Extracting the shared `beforeAll` block into a helper (that refactor is already noted for the four integration test files in `tests-local/architecture-test.md`).

## Recommended fix
1. Replace the `beforeAll` block with `beforeEach` (or add a `beforeEach` that resets state) so each test starts from the same dark-mode default.
2. If retaining `beforeAll` for performance, add a `beforeEach` that:
   - removes `light-theme` from `document.body.classList`,
   - re-queries `#moon-icon` and `#sun-icon` and resets their `display` attributes to the dark-mode defaults.
3. Verify that each of the four tests passes when run individually:
   ```console
   pnpm exec vitest run -t "defaults to dark mode" src/test/darkmode.test.ts
   pnpm exec vitest run -t "toggles to light mode" src/test/darkmode.test.ts
   pnpm exec vitest run -t "toggles back to dark mode" src/test/darkmode.test.ts
   pnpm exec vitest run -t "updates the icon when toggling" src/test/darkmode.test.ts
   ```

## Test plan
- Run each test in isolation; all four must pass.
- Run the full suite; all four must still pass.
- No new tests required — the fix is to the harness, not the assertions.

## Dependencies
None.

## Open questions
None.

## Follow-up items
None.

## Documentation impact
None expected; test-only change. `doc/TESTING.md` does not need an update for a single-file harness fix.

wainwright1000:
Fixes #542

Removes the test-ordering dependency in `src/test/darkmode.test.ts`: a `beforeEach` resets the body class and icon display state (with a loud-fail throw on fixture drift), and the "toggles back to dark mode" test is now self-contained (click to light, assert, click back). The ticket's isolated-run repro passes, and a shuffled-order sweep (5 seeds) is green.

The fix also hardens a second latent ordering dependency the ticket didn't name: the icon test previously inherited its dark start state from test order too.

Vera gate (two passes): pass 1 caught that the reset made the default-dark test vacuous — it asserted the state the hook had just forced, unpinning the load-path default. Fixed by capturing the post-load state in `beforeAll` and asserting the captured value; binding verified red-green by mutation. Pass 2 clean across all five reviewers; their one optional hardening (macrotask flush before the capture) is folded in. Full report and synthesis on the ticket.

281/281 tests green on the tip.


wainwmr (approved):
OK

Issue #549, opened 2026-06-11
buildScores checkLilypond discards error evidence (all failures report as 'not installed')


wainwright1000:
**Type:** tech debt
**Area:** Tooling

**Source:** `lilypond/build/buildScores.mjs` — `checkLilypond()`
**First seen:** 2026-06-11, Vera gate finding 539-05 (cycle 1) on ticket #539

**Description**

`checkLilypond()` catches every failure of the `lilypond --version` probe and prints a fixed message:

```javascript
catch { console.error("Error: lilypond is not installed or not on PATH"); }
```

The caught error's evidence (`error.message`, exit code, stderr) is discarded. Distinct failure modes — lilypond genuinely absent, `node` missing for a wrapper script (cmd exit 9009), a permission failure, a version-string parse error — all collapse into the same "not installed" message, misleading whoever debugs them. Observed concretely during #539's review: a node-absent failure in the test fixtures would be reported as "lilypond is not installed".

**Recommended fix**

Append the underlying evidence to the message, e.g.:

```javascript
catch (error) {
  console.error("Error: lilypond is not installed or not on PATH");
  console.error(String(error?.message ?? error));
}
```

Keep the friendly first line; add the diagnostic second line.

**Test plan**

Unit/integration: with a PATH lacking lilypond, the script exits 1 and stderr contains both the friendly line and the underlying error text. Existing "fails gracefully without lilypond" test extended to assert the second line.

**Dependencies**

Found during #539 review (same discard-the-evidence pattern as the test-side fix there). Sibling diagnostics tickets from the same family: #544, #545, #548.

- Vera

wainwright1000:
## Re-specification

Verified against `lilypond/build/buildScores.mjs` (lines 92-111) and `lilypond/test/buildScores.integration.test.ts` (line 427).

### Root cause — confirmed

`checkLilypond()` at line 96 uses a bare `catch`:

```javascript
catch {
  console.error("Error: lilypond is not installed or not on PATH.");
  console.error("Please install LilyPond before building scores.");
  process.exit(1);
}
```

The caught `ExecFileException` (carrying `message`, `stderr`, `status`) is discarded. Confirmed empirically on Windows: with `PATH=""`, `execSync('lilypond --version')` throws with message `"Command failed: lilypond --version\n'lilypond' is not recognized as an internal or external command, operable program or batch file."` — all of which is lost.

### Other catch blocks — scoped out

Three bare `catch` blocks exist in the file:

[...]

Issue #548, opened 2026-06-11
test:lilypond script finds no test files on Windows (unquoted glob, cmd.exe does not expand)


wainwright1000:
**Type:** tooling bug
**Area:** Test

**Source:** `package.json` — `test:lilypond` script
**First seen:** 2026-06-11, during Windows verification of #539

**Description**

`pnpm run test:lilypond` fails on Windows with `No test files found, exiting with code 1`. The script is:

```json
"test:lilypond": "vitest run lilypond/test/*.integration.test.ts"
```

The unquoted glob depends on shell expansion. On Linux/macOS, `sh` expands it to the concrete filename before vitest sees it (which is why CI stays green). On Windows, `cmd.exe` performs no glob expansion, so vitest receives the literal string `lilypond/test/*.integration.test.ts` as a filename filter, matches nothing, and exits 1. Result: Windows developers cannot run the integration suite via the script at all; the same suite runs fine when the file is named explicitly (`pnpm exec vitest run lilypond/test/buildScores.integration.test.ts` — 15/15 pass, verified 2026-06-11).

Same family as #539 (Windows-only test-infrastructure defects that CI cannot see) but a distinct defect in a distinct file.

**Recommended fix**

Make vitest do the matching instead of the shell. Options:

- (a) name the file(s) explicitly in the script; or
- (b) use a vitest-native include pattern (e.g. a `--project`/config-level `include` for `lilypond/test/*.integration.test.ts`), which vitest expands itself, cross-platform.

Option (b) is robust to new integration files; (a) is the one-line fix.

**Test plan**

`pnpm run test:lilypond` must run 15 tests and exit 0 on Windows and on Linux CI. A new `*.integration.test.ts` file must be picked up (if option b).

**Dependencies**

Sibling of #539 (this defect masked #539's script-level acceptance check on Windows). Note: the `test:integration` script name referenced in #539's body does not exist — it is `test:lilypond`; whichever fix lands should keep doc/TESTING.md and doc/CI.md consistent.

- Vera

wainwright1000:
## Re-specification

Verified against `package.json` (line 21) and `lilypond/test/*.integration.test.ts` (one file: `buildScores.integration.test.ts`).

### Root cause — refined

The ticket's diagnosis is directionally correct but misses the deeper cause.

[...]

Issue #394, opened 2026-05-28
buildScores.mjs leaves stale SVG on Lilypond failure, hiding retries


wainwright1000:
## Source

`build/buildScores.mjs`, function `buildScore` (lines 140-149)

## Description

When Lilypond fails during `buildScore`, the script calls `process.exit(1)` and leaves the old SVG in place. On the next run, `needsRebuild` compares `.ly` mtimes against the old SVG's mtime. If the `.ly` file has not changed, the choir is silently skipped with "SVG is up to date."

This hides persistent Lilypond failures. A developer who fixes an external issue (wrong Lilypond version, missing dependency, disk space) may not realise the build never retried the failing choir.

## Evidence

`buildScore` error handling (lines 147-149):

```js
try {
  execSync(
    `lilypond --svg -o "src/scores/${version}/${notation}/" "${ly}"`,
    { stdio: "inherit" }
  );
} catch (error) {
  console.error(`\nError building ${choirName}:\n${error.message}`);
  process.exit(1);
}
```

The old SVG is not deleted. Postprocessor failure (lines 155-162) already deletes the SVG before exiting — the Lilypond path should do the same.

## Impact

Scenario:
1. Choir 1 builds successfully. SVG exists.
2. User upgrades Lilypond to a broken version.
3. `build:scores` runs. Choir 1 fails. Script exits.
4. User fixes Lilypond.
5. `build:scores` runs again. Choir 1 is skipped because old SVG exists and `.ly` is unchanged.
6. User sees "SVG is up to date" and assumes all choirs built successfully.
7. Choir 1 SVG is never rebuilt with the fixed Lilypond.

## Recommended fix

Delete the SVG on Lilypond failure, matching the postprocessor failure pattern (PR #362):

```js
} catch (error) {
  rmSync(svg, { force: true });
  console.error(`\nError building ${choirName}:\n${error.message}`);
  process.exit(1);
}
```

This ensures the next build retries rather than skips.

## Test plan

1. Build an SVG (`npm run build:scores -- --choir 1`).
2. Temporarily break Lilypond (rename binary, etc.).
3. Re-run `npm run build:scores -- --choir 1`.
4. Assert build fails and SVG is deleted.
5. Restore Lilypond.
6. Re-run `npm run build:scores -- --choir 1`.
7. Assert build succeeds (not skipped).

## Notes

- `rmSync` with `{ force: true }` is safe — it silently succeeds if the SVG does not exist.
[...]

PR #550, opened 2026-06-11
test: fix Windows-only defects in buildScores integration test

### LINKED ISSUE #539: buildScores integration test: two Windows-only defects (failure-fixture gap and flaky timeout)
## 2. Description

The `buildScores` integration test suite has two Windows-only defects in `lilypond/test/buildScores.integration.test.ts`. Both pass on CI runners (macOS/Linux) and fail or flake only on Windows, so CI stays green while a Windows developer sees red. Neither is a regression in production `buildScores.mjs` code; both are test-harness defects.

## 3. Reachability

Run `pnpm run test:integration` on Windows. Two tests fail or flake:

- `deletes SVG on lilypond failure (#394)` fails deterministically.
- `caches by mtime` flakes under full-suite load.

## 4. Current state

`lilypond/test/buildScores.integration.test.ts`

- `createFakeLilypond()` (lines ~81-86) writes both `lilypond.bat` and an extensionless `lilypond` Unix shell script for the success fake.
- The failure fake in `deletes SVG on lilypond failure` (lines ~651-658) writes only an extensionless Unix `lilypond` shell script and no `lilypond.bat`.
- `buildScores.mjs` line 186 invokes LilyPond via `execSync("lilypond --svg ...")`, which on Windows runs through `cmd.exe` and resolves the command name against `PATHEXT`.
- All 14 integration tests have a 30,000 ms per-test timeout.
- `createWorkspace()` recursively copies `node_modules/@xmldom/xmldom` into each test's isolated workspace.

## 5. Root cause

**Defect 1 — failure-fixture gap.** On Windows, `cmd.exe` skips the extensionless `lilypond` script in the fail dir and finds the success fake's `lilypond.bat` still earlier on `PATH`. LilyPond "succeeds", the build returns 0, and the delete-on-failure path is never exercised.

**Defect 2 — flaky timeout.** Each test spawns 16 `lilypond` subprocesses and performs a recursive `cpSync` of `@xmldom/xmldom` during workspace setup. Windows process-spawn cost creates a ~3.5x timing spread (observed 2.36 s to 9.74 s in isolation). Under full-suite CPU contention this creeps toward the 30 s ceiling.

## 6. Scope

In scope: fixes to `lilypond/test/buildScores.integration.test.ts` only.

Out of scope: changes to `lilypond/build/buildScores.mjs` or any production code.

## 7. Recommended fix

**Defect 1.** Make the failure fake actually take effect on Windows. In the `deletes SVG on lilypond failure` test, write a failing `lilypond.bat` in `failDir` (e.g. `@echo off\necho fatal error: simulated lilypond failure\nexit 2`). Ensure `cmd.exe` resolves to the failing shim before the success fake's `.bat` on `PATH`.

**Defect 2.** Harden the spawn-heavy integration tests against Windows timing variance. The minimal change is to raise the per-test timeout for this file from 30 s to 60 s. As an optional optimisation, reduce per-test overhead by sharing one workspace across cases or by replacing the recursive `cpSync` of `@xmldom/xmldom` with a lighter resolution strategy (e.g. pointing `NODE_PATH` at the repo's `node_modules` rather than copying). The timeout raise alone is sufficient to close the flake.

## 8. Test plan

- **Defect 1:** After the fix, `deletes SVG on lilypond failure (#394)` must pass on Windows. Verify by running `pnpm run test:integration` on Windows and confirming the test exits 0.
- **Defect 2:** After the fix, `caches by mtime` must complete without timing out under full-suite load for three consecutive runs on Windows.

## 9. Dependencies

None.

## 10. Open questions

None.

## 11. Follow-up items

None.

## 12. Documentation impact

None expected. Test-file change only.

wainwright1000:
Fixes #539

Fixes the two Windows-only defects in `lilypond/test/buildScores.integration.test.ts`:

- **Failure-fixture gap**: the failure fake now ships a `lilypond.bat`, so `cmd.exe` resolves the failing shim instead of falling through to the success fake — the delete-on-failure path (#394) is genuinely exercised on Windows. Both platform wrappers delegate to one `_fake_lilypond.js`, so the failure logic is identical by construction, and a stderr pin (`simulated lilypond failure`) verifies the simulated fake — not a fallthrough or a version-check failure — handled the call.
- **Flaky timeout**: spawn-heavy tests raised 30s→60s, including "does not build OUP", which ran the identical 17-spawn workload on the 10s config default (9.74s observed isolated — the likeliest next flake).

Also: `chmod` failures now rethrow on non-Windows platforms instead of being silently swallowed (both fixture sites), so a Unix chmod failure can no longer masquerade as a cross-fixture fallthrough with the evidence discarded.

Verified on Windows: 15/15 integration (explicit file), 281/281 unit, plus two full-suite contention runs. Vera gate: two passes, reports on the ticket. Found en route and filed separately: #548 (`test:lilypond` script finds no test files on Windows — unquoted glob), #549 (`checkLilypond` discards error evidence).


wainwmr (approved):
OK

Issue #551, opened 2026-06-11
lily.ts: make Range fields readonly (compiler-enforce the documented ranges-immutability invariant)


wainwright1000:
**Type:** tech debt
**Area:** Lily

**Source:** `src/ts/lily.ts` — the `Range` interface/type
**First seen:** 2026-06-11, Vera gate finding 537-02 on ticket #537

**Description**

`Range.from`/`Range.to` are mutable, but nothing in the codebase mutates a `Range` after construction (the one `.to =` mutation nearby is on `FRlocation`, a different type), and the `LilypondData` documentation already instructs callers to treat `ranges` as immutable — an invariant enforced only by comment.

Ticket #537 showed the cost: a test predicate `(x.from = 122)` silently wrote through to the module-level `lilypondCache`, corrupting shared parser state. With `readonly` fields that bug is a TS2540 compile error.

**Recommended fix**

```typescript
readonly from: number;
readonly to: number;
```

on the `Range` shape in `src/ts/lily.ts`. Compile-check the codebase; the documented deep-readonly blocker concerns `notesByQuant`, not `Range`, so this should be a clean change.

**Test plan**

`pnpm run check:types` passes; a deliberate `range.from = 0` in a scratch test fails compilation with TS2540. Full suite green.

**Dependencies**

Follow-up to #537. Complementary to #544 (lint guard for the same defect family — belt and braces at different layers).

- Vera

wainwright1000:
**Re-specification**

Verified against `src/ts/lily.ts` at `origin/main`:

- `Range` is defined at lines 256-259 with mutable `from: number` and `to: number`.
- The only `.to =` mutation in `lily.ts` is at line 133 on `FRlocation`, a different type.
- No other code in `src/ts/` assigns to `Range.from` or `Range.to`.
- `LilypondData` already marks `ranges` as `readonly` at the Map level (line 277); making `Range` fields readonly closes the remaining gap.

Difficulty: **XS** (single-type change, compile-time verification only).

No delta to the Recommended fix or Test plan — the body is accurate as written.

- Kimi

wainwright1000:
Close the Loop: none. All risks and dependencies are covered by the Recommended fix and Test plan.

- Kimi

wainwright1000:
# VERA-551 Original Report (cycle 1)

Mode: retro-publish (gate run on the open PR #557 after an un-gated publish; scaffolding kept off-branch to avoid CI-triggering pushes)
Cycle: 1
[...]

PR #552, opened 2026-06-11
fix: correct broken assertions and incomplete loops in spem.test.ts

### LINKED ISSUE #537: spem.test.ts has broken assertions and incomplete part loops
## Type

tooling bug

## Area

Test

## Status

Todo

## Difficulty

S

## Source

`src/test/spem.test.ts`, lines 17, 28, 30

## First seen

- Lines 17, 28 (`p < 1`): 2024-05-11, commit `11f07746`
- Line 30 (`x.from = 122`): 2026-06-10, commit `f047669e` (PR #106)

## Description

`src/test/spem.test.ts` contains two broken assertions and two incomplete loops that mean the tests do not verify what their names claim.

**Broken assertion (always passes)** — Line 30:

```typescript
ranges.get(`${c}-${p}`)!.find((x) => (x.from = 122)) != null
```

The predicate uses `=` (assignment) instead of `===` (comparison). The assignment expression evaluates to `122` (truthy), so `find()` always returns the first element. The test passes regardless of whether bar 122 is inside any singing range.

**Incomplete loops** — Lines 17 and 28:

Both the "Everyone finishes at the end of bar 138" test (line 17) and the "Everyone is singing respice at bar 122" test (line 28) use `for (var p = 0; p < 1; p++)`. Only Soprano (part 0) is verified; Alto, Tenor, Baritone, and Bass are never checked.

A regression that broke parsing for parts 1–4, or that moved the respice entry to a different bar, would not be caught.

## Recommended fix

1. Change `p < 1` to `p < 5` on lines 17 and 28.
2. Change `x.from = 122` to `x.from === 122` on line 30.

No other files are affected. The fix is minimal and does not change test intent.

## Test plan

1. Apply the two changes above.
2. Run `pnpm run test:unit -- src/test/spem.test.ts` and verify all tests pass.
3. Verify that the fixed assertions would fail if the underlying data were wrong:
   - Temporarily change `last.to` to a value other than `139` in the first incomplete-loop test and confirm the test fails.
   - Temporarily change `x.from === 122` to `x.from === 999` in the second test and confirm the test fails.
4. Run the full unit test suite to confirm no regressions.

## Dependencies

None.

wainwright1000:
Fixes #537

Three-line fix in `src/test/spem.test.ts`, plus one folded sibling:

- The `find` predicate `(x.from = 122)` used assignment, not comparison — always truthy, so the "Everyone is singing respice at bar 122" assertion could never fail, **and** it wrote `122` through into the module-level parser cache (a live state-corruption hazard, contained only by Vitest's per-file isolation). Now `x.from === 122`.
- Two part-loops checked only Soprano (`p < 1`); now all five parts, so the "Everyone…" test names are literally true across all 40 voices.
- Folded from review (Vera 537-01): `src/test/lily.test.ts:77` carried the identical truncated loop — widened in the same way (scope note on the ticket).

The activation is mutation-verified during review: the old form passes for any constant while corrupting the cache; the fixed predicate fails for a wrong bar, and the widened loops fail when part 7/4's range is removed (probes in the Vera reports on the ticket). 281/281 unit tests green on the tip.

Preventions filed separately: `no-return-assign` added to #544's scope; #551 proposes `readonly` Range fields (would make this bug a compile error).


wainwright1000:
Coordination note: PR #557 (ticket #551, Range readonly) also touches `src/test/spem.test.ts` and carries the identical one-line fix `(x.from = 122)` → `x.from === 122` from this PR's cleanup — there it is compile-forced by the new `readonly` fields. A three-way trial merge of the two branches against `main` is clean; either merge order works, and the second to land will find that line already fixed. No action needed.

- Claude

wainwmr (approved):
OK

PR #557, opened 2026-06-11
refactor: make Range fields readonly in lily.ts (#551)

### LINKED ISSUE #551: lily.ts: make Range fields readonly (compiler-enforce the documented ranges-immutability invariant)

wainwright1000:
Make `Range.from` and `Range.to` `readonly` to compiler-enforce the documented immutability invariant on `LilypondData.ranges`.

This prevents accidental mutation of cached parser state. The `spem.test.ts` assertion that used assignment (`=`) instead of comparison (`===`) is also fixed, since it now fails compilation as TS2540.

Fixes #551

### DIFF
--- a/src/test/spem.test.ts
+++ b/src/test/spem.test.ts
@@ -1,4 +1,4 @@
-import { processLilypond } from "../ts/lily";
+import { processLilypond, type Range } from "../ts/lily";
 
 describe("Check that spem notes looks good", () => {
   const { notesByQuant, ranges } = processLilypond();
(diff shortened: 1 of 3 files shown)


wainwright1000:
Coordination note: this PR and #552 (ticket #537) both touch `src/test/spem.test.ts`, and both carry the identical one-line fix `(x.from = 122)` → `x.from === 122` (here it is compile-forced — `readonly` turns the stray assignment into TS2540; in #552 it is part of the wider assertion cleanup that found the bug in the first place). A three-way trial merge of the two branches against `main` is clean, so they can merge in either order; whichever lands second will simply find that line already fixed. No action needed — this note is so the textual adjacency isn't a surprise.

- Claude

Issue #558, opened 2026-06-11
lilypond/test-only PRs trigger no CI and cannot merge (required test check never runs)


wainwright1000:
**Type:** tooling bug
**Area:** Tooling
**Status:** Todo
**Difficulty:** S

**Source:**
`.github/workflows/ci.yml` lines 6-19 (`paths-ignore`), `.github/workflows/lilypond.yml` lines 3-16 (`paths`)

**First seen:** 2026-06-11, commit 8243cfd (gap introduced by the #474 workflow split, commit 547b66a)

**Description:**
`ci.yml` ignores `lilypond/**`; `lilypond.yml` triggers only on `lilypond/src/**` and the three build scripts. A PR touching only `lilypond/test/**` therefore triggers neither workflow. Because `test` is the required status check on `main`, such PRs can never satisfy it and are permanently blocked.

Live instance: PR #550 (the #539 fix, touches only `lilypond/test/buildScores.integration.test.ts`) shows no `test` check and `mergeStateStatus: BLOCKED`. Control: PR #552 (a `src/test` change) shows `test` SUCCESS and `CLEAN`.

Two consequences:

1. Test-only lilypond PRs are unmergeable without admin override.
2. Nothing runs the integration suite against integration-test changes — the changes most likely to break it.

The same trap applies to any PR whose whole diff lies inside `paths-ignore` (`*.md`-only PRs, workflow-file-only PRs).

**Recommended fix:**
Design call for the maintainer; options:

- (a) Narrow `ci.yml`'s ignore so `lilypond/test/**` (and `lilypond/build/*.d.mts`) still trigger the `test` job, and add `pnpm run test:lilypond` to that job. The integration suite fakes LilyPond, so this is fast and needs no LilyPond install.
- (b) Add `lilypond/test/**` to `lilypond.yml`'s `paths`. Downside: test-only changes would install LilyPond and regenerate SVGs, and non-byte-identical rebuilds could produce noise commits.
- (c) Ruleset-level: stop requiring `test`, or treat skipped-as-passed.

Option (a) looks cleanest.

In the same PR, fix `doc/CI.md` § `test` job: it still says the job runs `pnpm run test:lilypond`, which has not been true since #474 (`lilypond.yml` runs it instead).

**Test plan:**

1. PR touching only a `lilypond/test/**` file gets a `test` check and is mergeable when green (re-running on the #550 branch would verify directly).
[...]

Issue #474, opened 2026-06-03
CI fires wastefully on doc/.ly/buildScores changes; directory structure obscures UI vs LilyPond build separation


wainwmr:
## Description

`ci.yml` triggers on every PR regardless of what changed. For doc-only, `.ly`-only, or `buildScores.mjs`-only changes, the CI pipeline runs `npm run check`, `vite build`, unit tests and integration tests — work that does not validate the actual change.

- **Docs (`doc/**`, `*.md`)**: `vite build` bundles nothing new. Tests don't touch docs. Pure waste.
- **`.ly` files (`src/lilypond/**`)**: `vite build` bundles the *old* committed SVGs. The `.ly` change is only validated by `lilypond.yml` (regenerates SVGs). CI tests the wrong artefacts.
- **`buildScores.mjs`/`postprocessSvg.mjs`**: Integration tests do cover these, but `vite build` does not need them. Partial waste.

## Root cause

Two separate build pipelines are mixed in one directory tree with no visual boundary:

| Pipeline | Inputs | Outputs | Current location |
|----------|--------|---------|------------------|
| **UI build** | `src/ts/`, `src/scss/`, `src/ohmjs/`, `src/scores/`, `public/` | `dist/` | `src/`, `vite.config.ts` |
| **LilyPond build** | `src/lilypond/`, `build/buildScores.mjs`, `build/postprocessSvg.mjs` | `src/scores/` | `src/lilypond/`, `build/`, `src/test/buildScores.test.ts`, `src/test/integration/buildScores.test.ts` |

`build/` sounds like it could be part of the UI build. `src/lilypond/` is inside `src/` alongside UI code. `src/test/` mixes UI unit tests with LilyPond build integration tests.

`ci.yml` has no path filters because the boundaries are not obvious enough to express cleanly.

### Step 1: Restructure directories

Move everything that belongs to the LilyPond build pipeline into a single top-level directory:

```
lilypond/
  src/                    # from src/lilypond/
  build/                  # from build/ (buildScores.mjs, postprocessSvg.mjs, install-lilypond.sh, .d.mts)
  test/                   # from src/test/buildScores.test.ts and src/test/integration/buildScores.test.ts
src/
  scores/                 # keep here — committed assets consumed by UI
  test/                   # UI tests only
```

[...]

PR #559, opened 2026-06-11
fix: postprocessSvg.mjs CLI entry point unreachable on Windows and relative paths

### LINKED ISSUE #555: postprocessSvg.mjs CLI entry point silently fails on Windows and relative paths
**Type:** tooling bug
**Area:** Tooling
**Status:** Todo
**Difficulty:** XS

**Source:**
`lilypond/build/postprocessSvg.mjs`, line 289

**Description:**
The CLI entry point guard `import.meta.url === new URL(process.argv[1], "file://").href` silently fails on Windows and when the script is invoked with a relative path. `main()` is never called; the process exits 0 with no output and no error.

**Steps to reproduce:**
1. From the repository root, run: `node lilypond/build/postprocessSvg.mjs --spem lilypond/src/Hugh\ Keyte/spem.ly some.svg`
2. Observe: process exits 0 immediately, no output, no SVG written.

**Expected behaviour:**
The script should parse arguments and call `postprocessSvg`.

**Actual behaviour:**
`main()` is skipped because the URL comparison fails. `new URL("lilypond/build/postprocessSvg.mjs", "file://")` produces `file:///lilypond/build/postprocessSvg.mjs`, while `import.meta.url` is `file:///C:/Users/.../lilypond/build/postprocessSvg.mjs`. On Windows, even an absolute path produces a non-matching URL.

**Root cause:**
`new URL(process.argv[1], "file://")` does not resolve relative paths against the script's directory, and on Windows it produces a non-URL string for absolute paths. The standard Node.js pattern is `pathToFileURL(process.argv[1]).href`.

**Recommended fix:**
```js
import { pathToFileURL } from "url";
if (import.meta.url === pathToFileURL(process.argv[1]).href) {
  main();
}
```

**Test plan:**
1. Add a test that verifies `main()` is reachable when the module is the entry point. This can be done by extracting the guard into a testable helper or by spawning the process in an integration test.
2. Verify the fix on Windows (relative and absolute paths) and Unix (relative and absolute paths).

**Acceptance criteria:**
- [ ] `node lilypond/build/postprocessSvg.mjs <svg>` runs `main()` and either processes the SVG or prints usage.
- [ ] `node lilypond/build/postprocessSvg.mjs --spem <path> <svg>` passes the custom spem path.
- [ ] No regression when imported as a module from `buildScores.mjs`.

**Risk and scope:**
Low risk. The fix replaces a broken guard with a standard Node.js API. Only affects manual CLI invocation; the module import path used by `buildScores.mjs` is unchanged.

**Notes:**
- The `main()` function also has an unchecked array bounds access when parsing `--spem` / `--words` flags (`args[++i]`). If the flag is the last argument, `args[i]` is `undefined`. This should be fixed at the same time or in a follow-up.
- Related: #355 (shebang removed), #356 (malformed anchor hrefs).

wainwright1000:
Fixes #555

## What

The CLI entry-point guard `import.meta.url === new URL(process.argv[1], "file://").href` never matched on Windows (drive letter parses as a URL scheme) and could not match relative invocations, so `main()` silently never ran — exit 0, no output, no SVG written.

- Replaced with an exported `isMainModule(argv1, importMetaUrl)` using the `realpathSync` comparison from the sibling `buildScores.mjs` (a deliberate deviation from the ticket's `pathToFileURL` suggestion, for consistency and symlink/case resilience), hardened so a missing or nonexistent `argv[1]` means "not the entry point" rather than a crash (ENOENT caught; other fs errors stay loud).
- The argument parser now fails loudly instead of silently misbehaving: bare `--spem`/`--words`, flag-shaped values, unknown options, and surplus positionals all error with exit 1 (the ticket's Notes flagged the `args[++i]` bounds gap; the rest fell out of review).
- 12 tests in `lilypond/test/postprocessSvg.cli.integration.test.ts`: five in-process `isMainModule` unit tests — including the relative-`argv[1]` input that discriminates the old guard from the new one on any platform (Node absolutises `argv[1]`, so spawn tests cannot reach it) — plus seven spawn tests (the ticket's repro invocation as the smoke, usage, and the five rejection paths).
- `doc/TESTING.md` and `doc/CI.md` corrected where they still described the pre-#474 CI layout (`integration` job, `dorny/paths-filter`, `test:lilypond` in the `test` job). The remaining `lilypond/test/**` trigger gap is tracked as #558.

## Test plan

- `pnpm run ci` green (check, build, 281 unit tests).
- Integration suite green on Windows except the pre-existing #539 fake-lilypond failure (fix already open in #550); the new file's 12 tests pass on Windows and are exercised on Linux by `pnpm run test:lilypond` in the Regenerate SVGs workflow, which this PR triggers via its `postprocessSvg.mjs` change.
- Acceptance criteria: `node lilypond/build/postprocessSvg.mjs <svg>` runs `main()`; `--spem` path honoured; module-import path unchanged (existing unit tests).

Vera gate: cycle 1, two passes, second pass clean — [Original Report](https://github.com/wainwmr/spem-player/issues/555#issuecomment-4676979901), [Final Synthesis](https://github.com/wainwmr/spem-player/issues/555#issuecomment-4677059161).

### DIFF
--- a/doc/CI.md
+++ b/doc/CI.md
@@ -11,13 +11,13 @@ The repository uses GitHub Actions for automated testing and dependency updates.
 
 ### `ci.yml`
 
-Triggers on every push and pull request to `main`, and nightly at 00:00 UTC via a `schedule` cron.
+Triggers on push to `main` and on pull requests targeting `main` — except changes confined to the workflow's `paths-ignore` list (see `ci.yml`; it ignores root-level `*.md`, `lilypond/**`, and three sibling workflow files) — and nightly at 00:00 UTC via a `schedule` cron.
 
 Defines one job.
 
 #### `test` job
 
-Runs on every trigger. Executes `pnpm run check` (lint, format, type check, unused, deps) and `pnpm run build`, then `pnpm run test:unit`, then `pnpm run test:lilypond`. This is the required status check for the `main` branch ruleset; pull requests cannot merge until it passes.
+Runs on every trigger. Executes `pnpm run check` (lint, format, type check, unused, deps) and `pnpm run build`, then `pnpm run test:unit`. This is the required status check for the `main` branch ruleset; pull requests cannot merge until it passes. The LilyPond integration suite (`pnpm run test:lilypond`) does not run here — it runs in the Regenerate SVGs workflow below. Changes confined to `lilypond/test/**` trigger neither workflow, so the required check never reports and such PRs cannot merge without a ruleset bypass (#558).
 
 ### `e2e.yml`
 
--- a/doc/TESTING.md
+++ b/doc/TESTING.md
@@ -61,7 +61,7 @@ Scripts in `.github/scripts/` that expose pure functions use the Node.js built-i
 
 Integration tests live in `lilypond/test/*.integration.test.ts` and also follow `*.test.ts`. They typically spawn subprocesses (LilyPond, the build pipeline) and are substantially slower.
 
-If a new shared fixture directory is introduced (e.g. `src/test/fixtures/`), add it to the `build-related` filter in `.github/workflows/ci.yml` so that PRs touching it exercise the integration suite.
+When changing integration tests, always run `pnpm run test:lilypond` locally — see § CI Behaviour below for which workflows run them in CI and the `lilypond/test/**` trigger gap (#558).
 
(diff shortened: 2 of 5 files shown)


wainwmr (approved):
OK

PR #560, opened 2026-06-11
fix: remove wall-clock draw throttle that dropped frames on 120Hz displays

### LINKED ISSUE #554: MusicCanvas #shouldDraw drops 50% of frames on 120Hz displays
MusicCanvas.#shouldDraw() uses a wall-clock throttle (`Date.now() - oldTimeStamp < 0.01s`) to drop frames during playback. This drops approximately 50% of frames on 120Hz displays and causes jitter unrelated to the display refresh cycle.

Probe `probes/probe-musiccanvas-shoulddraw-throttle.cjs` confirmed: at 120Hz, 60 of 120 frames are dropped because the 8.3ms frame interval is below the 10ms threshold.

The throttle was originally added to reduce CPU load during playback, but it is keyed to wall-clock time rather than `requestAnimationFrame` delta time or vsync. On high-refresh-rate monitors this produces visible stuttering.

**Location:** `src/ts/MusicCanvas.ts`, lines 321-328 (`#shouldDraw`)

**Recommended fix:** Replace the `Date.now()` throttle with `requestAnimationFrame` delta timing, or remove the throttle entirely and rely on `requestAnimationFrame`'s natural vsync pacing (the browser already throttles to the display refresh rate).

**Test plan:**
- Add a test that verifies `#shouldDraw` returns `true` for every `requestAnimationFrame` callback during playback, or
- Mock `requestAnimationFrame` at 120Hz and assert no frames are dropped.

- Zimi

wainwright1000:
Fixes #554

Removes the `#shouldDraw()` wall-clock throttle (10ms `Date.now()` interval) and its `oldTimeStamp` state from MusicCanvas. The throttle assumed ~60Hz frame intervals and dropped every other frame at 120Hz (probe-confirmed 60/120); `requestAnimationFrame` already paces `draw()` to the display refresh rate.

Two consequences surfaced by the Vera gate and fixed here:

- The throttle was silently collapsing a duplicate render per frame: the controls loop drives `setBar` at rAF rate alongside the play loop. `setBar` now draws only when not playing (the play loop is the playback renderer; when paused the direct draw gives immediate feedback, with the shimmer loop a frame behind).
- `disconnectedCallback` cancels the play loop but leaves `playing` true, and nothing restarted the loop on reconnect — with the new guard that would have been a silent freeze. `connectedCallback` now restarts the play loop when playing with no loop, symmetric with the shimmer restart.

Tests: the old playback test (a single `draw()` call that could not detect throttling) is replaced by a 120Hz simulation asserting all 12 frames render (6/12 with the old throttle), plus tests pinning the `setBar` guard (state commits, no direct render while playing; exactly one while paused) and the reconnect restart.

Vera gate: 3 passes, 26 findings (1 critical, 8 important — all resolved; 4 deferred to the refactor report). Full record on #554.

- Vera

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "spem-player",
   "private": true,
-  "version": "2.8.1",
+  "version": "2.8.2",
   "description": "Helping singers learn Thomas Tallis Spem in alium",
   "type": "module",
   "dependencies": {
--- a/src/test/canvas.test.ts
+++ b/src/test/canvas.test.ts
@@ -37,11 +37,128 @@ describe("MusicCanvas custom element", () => {
     freshCanvas.draw();
   });
 
-  it("draw() executes during playback without throttling", () => {
+  it("draw() renders every frame at 120Hz during playback (#554)", () => {
     expect(canvas).not.toBeNull();
-    canvas!.playing = true;
-    canvas!.draw();
-    canvas!.playing = false;
+    // Fixture precondition: draw() returns silently on empty data, so a dead
+    // fixture would report 0 renders and masquerade as a throttle regression.
+    // notesByQuant is the guard that actually trips on a data-less parse;
+    // ranges is seeded per choir-part pair unconditionally (lily.ts), so its
+    // size only detects #init never completing.
+    expect(canvas!.ranges.size).toBeGreaterThan(0);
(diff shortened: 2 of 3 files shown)


wainwmr (approved):
OK

PR #574, opened 2026-06-12
fix: fail fast on unknown Duration; guard grammar/switch parity (#170)

### LINKED ISSUE #170: Protect against grammar/switch divergence in Duration parsing
The Duration constructor in music-classes.ts has a silent fallback (sfths = 0) for unknown duration strings. This should never happen because the Ohm grammar strictly limits valid durations, but if the grammar and switch drift, the failure is silent and dangerous.

wainwright1000:
Fixes #170.

The `Duration` constructor's `default` switch case silently set `sfths = 0` for
an unknown duration string. Because `lily.ts` advances bar position by
`comp.duration.sfths / barsize`, a zero-length duration stalls advance and
corrupts timing with no error — a silent failure reachable only if the Ohm
grammar and the switch drift apart.

This change:

- Throws `Unknown duration: <value>` from the `default` case instead of
  silently returning `0` (fail fast).
- Adds `src/test/grammar-consistency.test.ts`, a CI guard that extracts the
  grammar's `duration` literals and the switch's `case` labels from source and
  fails — naming the offending literal — if they ever diverge.
- Updates the existing test to expect the throw, and pins the empty-string and
  unknown-with-dot/multiplier cases.
- Makes the constructor's `duration` parameter required. The old `= ""` default,
  combined with the new throw, was a latent guaranteed throw and a misleading
  affordance. This tightens the signature slightly beyond the ticket's stated
  scope, so flagging it for review.

No user-facing behaviour changes for valid scores (the `default` case is
unreachable via the parser), so no version bump.

The Vera review gate ran (two passes, clean); the full synthesis is on the ticket.

- Vera

### DIFF
--- a/src/test/grammar-consistency.test.ts
+++ b/src/test/grammar-consistency.test.ts
@@ -1,0 +1,72 @@
+import { readFileSync } from "fs";
+import { describe, expect, it } from "vitest";
+
+/**
+ * Guards the implicit, manual coupling between the Ohm grammar's `duration`
+ * rule and the `switch` in the `Duration` constructor (see #170). If a
+ * developer adds a duration literal to one and forgets the other, the parser
+ * produces a string the switch does not handle, the constructor throws
+ * "Unknown duration" at runtime — and this test fails first, at CI, naming the
+ * offending literal so the divergence never reaches production.
+ *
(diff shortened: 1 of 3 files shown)


wainwmr (approved):
All good

PR #576, opened 2026-06-12
fix: use the requested recording's intro_beats for the initial bar (#241)

### LINKED ISSUE #241: parseURL uses ALC intro beats for initial bar when CotE recording requested
## Description

When the app is loaded with `?recording=cote` and no explicit `bar` parameter, the initial playback position is calculated using ALC's intro beat count instead of CotE's. ALC has 2 intro beats and CotE has 4. The result is that CotE playback starts at bar 0.5 instead of bar 0, causing a slight but audible offset at the start of the piece.

## Steps to Reproduce

1. Open the Spem Player with the URL `?recording=cote` (no `bar` parameter).
2. Observe the initial bar indicator.

## Expected result

Playback starts at bar 0, which for CotE is `1 - 4/4 = 0`.

## Actual result

Playback starts at bar 0.5, which is `1 - 2/4 = 0.5` (ALC's calculation).

## Current state

- File: `index.ts`
- Function: `parseURL()`
- Line: 154

```typescript
var bar = 1 - config.intro_beats[recording] / 4;
```

At this point `recording` is a local variable initialized to 0 (ALC) on line 151. The URL parameter `recording=cote` is parsed into variable `r` inside the loop (lines 170–172), but `r` is not used for the initial bar calculation.

## Root cause

The initial bar calculation happens before the URL parameter loop. The code uses the default `recording` value (0 = ALC) instead of the parsed `r` value.

## Scope

- In scope: use the correct recording's intro_beats for initial bar calculation.
- Out of scope: other URL parameter parsing changes.

## Recommended fix

Move the bar initialisation to after the URL parsing loop, so that `r` has been set:

```typescript
var bar: number;
// ... parse loop ...
bar = 1 - config.intro_beats[r] / 4;
// ... remaining parseURL logic ...
```

Alternatively, compute the default after the loop:

```typescript
var bar = 1 - config.intro_beats[r] / 4;
```

placed after line 176 (the end of the for loop), before the calls to `setRecording`, `setChoir`, etc.

If a `bar` parameter is present in the URL, it overrides this default on line 167, so moving the default computation after the loop is safe.

## Test plan

1. Load `?recording=cote` with no `bar` parameter. Verify score and controls show bar 0.
2. Load `?recording=alc` with no `bar` parameter. Verify score and controls show bar 0.5.
3. Load `?recording=cote&bar=10`. Verify bar is 10 (explicit parameter overrides default).

## Dependencies

None.

wainwright1000:
Fixes #241.

`parseURLSearch` computed the default initial bar from a pre-loop `recording = 0`
(ALC) **before** the URL loop parsed the recording into `r`, so `?recording=cote`
with no explicit `?bar=` started at ALC's `1 - 2/4 = 0.5` instead of CotE's
`1 - 4/4 = 0` — a slight but audible offset at the start of the piece.

The fix moves the default-bar computation to **after** the loop, keyed off the
parsed `r` (`1 - config.intro_beats[r] / 4`), with a `barParam: number | undefined`
sentinel so an explicit `?bar=` (including `?bar=0`) still overrides the default.
The now-dead pre-loop `recording` local is removed.

Note: the ticket's spec pointed at `index.ts:154`, but URL parsing was since
refactored into `parseURLSearch` (`src/ts/url.ts`); the identical bug and the fix
live there (a note is on the ticket).

Tests (`src/test/url.test.ts`): `?recording=cote` → bar 0, `?recording=alc` → 0.5,
explicit `?recording=cote&bar=10` → 10, and `?recording=alc&bar=0` → 0 (pinning the
`??` sentinel against a future `||` regression).

Patch version bump (user-facing fix). The Vera review gate ran (one pass, clean
fix); synthesis on the ticket.

- Vera

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "spem-player",
   "private": true,
-  "version": "2.8.2",
+  "version": "2.8.3",
   "description": "Helping singers learn Thomas Tallis Spem in alium",
   "type": "module",
   "dependencies": {
--- a/src/test/url.test.ts
+++ b/src/test/url.test.ts
@@ -61,4 +61,17 @@ describe("parseURLSearch", () => {
   it("parses ?dark=1 as dark mode (#236)", () => {
     expect(parseURLSearch("?dark=1").dark).toBe(true);
   });
+
+  it("uses the requested recording's intro_beats for the default bar (#241)", () => {
+    // intro_beats = [ALC 2, CotE 4]; the initial bar is 1 - intro_beats/4.
+    // The default bar must be keyed off the *parsed* recording, not the
+    // pre-parse default (ALC). CotE with no explicit ?bar= starts at bar 0.
+    expect(parseURLSearch("?recording=cote").bar).toBe(1 - 4 / 4); // 0
+    expect(parseURLSearch("?recording=alc").bar).toBe(1 - 2 / 4); // 0.5
+    // An explicit ?bar= still overrides the recording-derived default.
+    expect(parseURLSearch("?recording=cote&bar=10").bar).toBe(10);
+    // Explicit ?bar=0 is kept, not treated as "unset" — pins the `??` sentinel
+    // boundary against a future `||` regression (0 is falsy but not nullish).
+    expect(parseURLSearch("?recording=alc&bar=0").bar).toBe(0);
+  });
 });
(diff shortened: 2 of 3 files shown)


wainwmr (approved):
All good

Issue #575, opened 2026-06-12
MusicScore load-error surfacing: unobserved async rejections and console.error-only sink


wainwright1000:
## Background

Surfaced by the Vera gate on #237 (findings 237-04 and 237-S3). Not introduced by
#237 — a pre-existing pattern that #237 adds a new trigger to.

### 2. Description

When a score SVG fails to load, the user is left with a **blank score panel and no
feedback** — indistinguishable from "still loading". Worse, if the failure happens
in the DOM-building half of the load (after the SVG fetch), the error becomes an
**unobserved promise rejection**: nothing logs it, no error id, no user-visible
state. The two facets:

1. **Unobserved async rejections.** `MusicScore`'s async setters (`setChoir`,
   `setScoreType`, and `setRecording` once #237 lands) are invoked
   *fire-and-forget* from `attributeChangedCallback`. Each `await`s
   `#loadScore()`, whose post-`await` body does substantial synchronous DOM work
   (innerHTML assignment, `querySelector`, `viewBox` parse, `prepend`, `getBars`,
   `requestAnimationFrame`). A throw anywhere in that body rejects the setter's
   promise, which nothing observes — no log, no surfaced error.

2. **`console.error`-only sink.** The only error reporting in the load path is two
   `console.error` calls. On a genuine load failure the panel silently stays blank
   or stale; the user gets nothing actionable.

### 3. Reachability

Not a Bug ticket (Type: tech debt). The failure path is reachable only on a
genuine asset-load failure (a missing or malformed bundled SVG, a chunk-load
error, or a DOM op throwing on unexpected SVG shape), not during normal use with
intact assets. The value is robustness and diagnosability: when it *does* fail, it
fails silently and invisibly, which is the defect being addressed.

### 4. Steps to Reproduce

1. Build and serve the app.
2. Induce a score-load failure — e.g. remove or corrupt one choir's
   `Choir <name>.svg` for the selected score type, or force the dynamic
   `import(... .svg?raw)` to reject (offline + cache cleared).
3. Switch to the affected choir / score type / recording.
4. Observe the score panel: **it is blank, with nothing in the console beyond a
   bare `Error loading SVG` (if the failure was in `#loadSvg`) or nothing at all
   (if the failure was in the post-`await` DOM body).**

### 5. Expected result

A genuine load failure produces (a) a single identifiable log line naming what
failed, and (b) a visible "could not load score" state in the panel, so the user
[...]

Issue #573, opened 2026-06-12
Generate CotE-named score SVGs so the score shows CotE choir labels


wainwright1000:
## 2. Description

When the recording is switched to Choir of the Earth (CotE), `MusicScore` loads SVGs named from `config.choirs[1]` (`"1"` … `"8"`). Those files do not exist: `src/scores/Hugh Keyte/{modern,early}/` contains only ALC-named SVGs (`Choir I A.svg` … `Choir IV B.svg`). This ticket adds the CotE-named SVGs and updates the LilyPond build pipeline to keep them in sync.

The CotE score is musically identical to the ALC score. The Hugh Keyte edition uses a single set of LilyPond source files; the only difference is the filename/label mapping exposed to the user.

## 3. Reachability

User-visible once #237 (the `MusicScore` wiring) is fixed:

1. Load the app.
2. Switch the recording from ALC to CotE.
3. Change choir or wait for the score to reload.
4. The score fails to render because `Choir 1.svg` (etc.) is missing.

## 4. Steps to Reproduce

1. Check out `origin/main`.
2. List `src/scores/Hugh Keyte/modern/`.
3. Observe `Choir I A.svg` … `Choir IV B.svg` and no `Choir 1.svg` … `Choir 8.svg`.
4. (With #237 applied) switch recording to CotE and change choir.
5. The browser network tab shows a 404 for `Choir 1.svg`.

## 5. Expected result

`src/scores/Hugh Keyte/{modern,early}/` contains both ALC-named and CotE-named SVGs. When CotE is selected, `MusicScore.#loadSvg()` resolves `import("../scores/Hugh Keyte/modern/Choir 1.svg?raw")` successfully and renders the same music as the ALC variant.

## 6. Actual result

Only ALC-named SVGs exist. Any code path that tries to load a CotE-named SVG receives `null` and logs an error.

## 7. Current state

- `src/ts/config.ts` defines the choir name mappings:
  - `config.choirs[0]` = `["I A", "I B", "II A", "II B", "III A", "III B", "IV A", "IV B"]` (ALC)
  - `config.choirs[1]` = `["1", "2", "3", "4", "5", "6", "7", "8"]` (CotE)
- `src/ts/MusicScore.ts:208` builds the SVG filename:
  ```typescript
  const choirName = config.choirs[this.recording][this.choir];
  const svgModule = await import(`../scores/Hugh Keyte/${this.scoreType}/Choir ${choirName}.svg?raw`);
  ```
[...]

PR #577, opened 2026-06-12
fix: observe recording on MusicScore and fall back to ALC score when CotE SVG missing (#237)

### LINKED ISSUE #237: Score displays ALC choir labels when CotE recording is selected
A singer switches the recording to Choir of the Earth (CotE). The audio correctly plays the CotE track, but the score still shows ALC choir names ("I A", "I B") instead of CotE names ("1", "2"). If the code were fixed to observe the recording attribute, the score would fail to load because CotE SVG files do not exist.

## Steps to Reproduce

1. Load the app.
2. Click the recording switch to change from ALC to CotE.
3. Observe that the score SVG still shows "I A" (or equivalent ALC name) in the title.
4. Inspect the `music-score` element: `score.recording` is still `0`.

## Expected result

The score should reflect the selected recording's naming convention.

## Actual result

`MusicScore` never updates its `recording` property because `"recording"` is not in `observedAttributes`.

## Current state

- `index.ts`, `setChoir()`, line 104: `score.setAttribute("recording", String(current.recording));`
- `src/ts/MusicScore.ts`, line 10: `static observedAttributes = ["choir", "part", "bar", "playing", "score-type"];` — `"recording"` is missing.
- `src/ts/MusicScore.ts`, line 101: `const choirName = config.choirs[this.recording][this.choir];`
- `src/scores/Hugh Keyte/modern/` and `early/` contain only ALC-named SVGs (`Choir I A.svg`, etc.). No CotE-named SVGs (`Choir 1.svg`, etc.) exist.

## Root cause

`MusicScore` does not observe the `recording` attribute. `index.ts` sets the attribute inside `setChoir()`, but the component ignores it. CotE SVG files were never generated.

## Scope

- In scope: Add `"recording"` to `MusicScore.observedAttributes` and ensure `setRecording()` is called when the attribute changes.
- In scope: Either generate CotE-named SVGs or add a fallback to ALC names when CotE SVGs are missing.
- Out of scope: Changing the SVG content itself.

## Recommended fix

1. Add `"recording"` to `MusicScore.observedAttributes`.
2. In `index.ts`, move `score.setAttribute("recording", ...)` from `setChoir()` to `setRecording()` so it updates when recording changes without requiring a choir change.
3. In `MusicScore.#loadSvg()`, fall back to ALC filenames if CotE SVGs are missing:

   ```typescript
   const choirName = config.choirs[this.recording][this.choir];
   const fallbackName = config.choirs[0][this.choir];
   try {
     const svgModule = await import(`../scores/Hugh Keyte/${this.scoreType}/Choir ${choirName}.svg?raw`);
     ...
   } catch {
     const svgModule = await import(`../scores/Hugh Keyte/${this.scoreType}/Choir ${fallbackName}.svg?raw`);
     ...
   }
   ```

## Test plan

1. Create a `MusicScore` element, set `recording="1"`, assert `score.recording === 1`.
2. Toggle recording without changing choir; assert score attempts to reload with CotE name.
3. Verify CotE SVG fallback loads ALC file when CotE file is missing.

## Dependencies

None (but related to score generation infrastructure).

### LINKED ISSUE #504: config.choirs: add length-parity assertion (per-recording choir dropdown assumes equal-length lists)

wainwright1000:
Fixes #237. Fixes #504.

`MusicScore` now observes the `recording` attribute and reloads on change, and `index.ts setRecording()` propagates recording to the score (moved out of `setChoir`). When the CotE-named SVG is missing, `#loadSvg` falls back to the ALC file via a new pure `loadWithFallback`, so the score renders rather than going blank.

This is a **correctness + readiness** change. Because CotE-named SVGs do not exist yet, the score still visibly shows the ALC labels ("I A", "I B"…) — it loads the ALC file via the fallback. Visible CotE labels are blocked on generating CotE SVGs, filed as #573. The wiring is correct and safe, and lights up the moment #573 lands the SVGs.

## What changed

- `MusicScore.observedAttributes` now includes `recording`; `setRecording()` is overridden to reload (mirrors `setChoir`/`setScoreType`).
- `index.ts`: recording propagation moved from `setChoir` into `setRecording` (the sole site reaching the score), so toggling recording without a choir change updates the score.
- New pure `MusicScore.loadWithFallback(primary, fallback, importer)` + `ChoirSvgNames` type: tries the recording's own SVG, falls back to the ALC name, binds and logs errors, and short-circuits when there is no distinct fallback.
- `common.ts`: import-time assert that `config.choirs` has one row per recording, all equal length (backs the fallback invariant; mirrors the existing recording-length and bartime/barno guards). This is the standalone hardening specified as #504, folded in here because #237 depends on the same invariant.

## Tests

- Unit (`src/test/score.test.ts`): observes recording, reloads on recording change, `choirSvgNames` resolution, four `loadWithFallback` branch tests, and a recording-vs-choir race test.
- e2e (`e2e/recording-toggle.spec.ts`): toggling to CotE sets the score's `recording` and the score still renders (real Vite import fallback), and a choir change retains the recording.

## Notes

- Went through the Vera gate (two passes; findings resolved or deferred — see the #237 comments). One pre-existing fire-and-forget load-error pattern deferred to #575.
- Version bumped to 2.8.3. Heads-up: #241 (PR #576) also bumps to 2.8.3, so whichever merges second will need a trivial version rebase.

- Claude

### DIFF
--- a/e2e/recording-toggle.spec.ts
+++ b/e2e/recording-toggle.spec.ts
@@ -1,0 +1,37 @@
+import { test, expect } from "@playwright/test";
+
+// #237: the recording switch must propagate to the score, and the score must
+// keep rendering when the (currently non-existent, #573) CotE SVG falls back to
+// the ALC file. This exercises the two paths jsdom cannot: the index.ts wiring
+// (setRecording -> score attribute) and the real Vite dynamic-import fallback.
+test.describe("Recording switch propagates to the score (#237)", () => {
+  test("toggling to CotE updates the score's recording and keeps it rendered", async ({
+    page,
+  }) => {
+    await page.goto("/");
+
+    const score = page.locator("music-score");
+
+    // The score renders at the default (ALC) recording.
+    await expect(score.locator("svg").first()).toBeVisible();
+    await expect(score).toHaveAttribute("recording", "0");
+
+    // Toggle the recording switch: ALC -> CotE.
+    await page.locator("#recordingswitch").click();
+
+    // The recording reaches the score (the wiring #237 adds), and the score is
+    // still rendered: CotE-named SVGs do not exist (#573), so #loadSvg falls
+    // back to the ALC file rather than leaving a blank panel.
+    await expect(score).toHaveAttribute("recording", "1");
+    await expect(score.locator("svg").first()).toBeVisible();
+
+    // #237 removed the recording propagation from setChoir, so a choir change
+    // must now leave the score's recording untouched (the score observes
+    // recording independently). Change choir via the keyboard (digit n -> choir
+    // n-1) and confirm CotE (recording 1) is retained.
+    await page.keyboard.press("3");
+    await expect(score).toHaveAttribute("choir", "2");
+    await expect(score).toHaveAttribute("recording", "1");
+    await expect(score.locator("svg").first()).toBeVisible();
+  });
+});
--- a/index.ts
+++ b/index.ts
@@ -125,9 +125,6 @@ async function setChoir(c: number, forceChange = false) {
   // Update the score for this choir
   score.setAttribute("choir", String(current.choir));
 
-  // Set the recording of the audio to use
-  score.setAttribute("recording", String(current.recording));
-
   // Update the canvas
   canvas.setAttribute("choir", String(current.choir));
 }
@@ -511,6 +508,11 @@ async function setRecording(r: number) {
 
   // Update the input field
   controls.setAttribute("recording", String(current.recording));
+
+  // Propagate to the score so it reloads with the selected recording's naming
+  // (#237). This is the only place recording reaches the score, so toggling
+  // recording without a choir change updates it.
+  score.setAttribute("recording", String(current.recording));
 }
 
 function toggleRecording() {
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "spem-player",
   "private": true,
-  "version": "2.8.2",
+  "version": "2.8.3",
   "description": "Helping singers learn Thomas Tallis Spem in alium",
   "type": "module",
   "dependencies": {
--- a/src/test/score.test.ts
+++ b/src/test/score.test.ts
@@ -737,4 +737,198 @@ describe("MusicScore custom element", () => {
 
     MusicScore.testSvgLoader = originalLoader;
   });
+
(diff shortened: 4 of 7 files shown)


wainwmr (changes requested):
This seems like totally unnecessary bloat.  I don't see a problem.  What missing CotE score are you talking about? Why write code and waste time in CI builds and my time for a recording we don't have?

wainwright1000:
Yep. looks like bollocks to me too.  I'll do a check to see what other non-bug things are lurking in the pile, and see if I can zap them.

PR #579, opened 2026-06-12
build: have ESLint honour .gitignore via includeIgnoreFile (#509)

### LINKED ISSUE #509: eslint.config.js: honour .gitignore via includeIgnoreFile (drop duplicated ignore entries)
**Type**: tech debt
**Area**: Tooling

**Source**: `eslint.config.js` (the `ignores` array)

## Description

`npm run check:lint` runs `eslint .` over the whole tree. ESLint's flat config
does not read `.gitignore` by default (it only auto-ignores `node_modules/` and
`.git/`), so the `ignores` array in `eslint.config.js` hand-maintains a list that
substantially duplicates `.gitignore`: `dist/`, `node_modules/`, `coverage/`,
`.netlify/` are listed in both. The two lists can drift, and any gitignored
working directory that gains a lintable file (for example a scratch `.ts`/`.cjs`
in `temp/`) is linted and fails `--max-warnings 0` until someone adds it to the
ESLint ignores by hand.

## Recommended fix

Have ESLint honour `.gitignore` directly via `includeIgnoreFile` from
`@eslint/compat`:

```js
import { includeIgnoreFile } from "@eslint/compat";
import { fileURLToPath } from "node:url";

const gitignorePath = fileURLToPath(new URL(".gitignore", import.meta.url));

export default tseslint.config(
  includeIgnoreFile(gitignorePath),
  // ... existing config
);
```

Then remove the now-redundant `dist/`, `node_modules/`, `coverage/`, `.netlify/`
entries from the manual `ignores` array, leaving only paths that are genuinely
not gitignored (e.g. generated bundles that are committed, if any). `.gitignore`
becomes the single source of truth for "do not lint this".

## Test plan

- `npm run check:lint` passes on a clean tree (no regression).
- A scratch `.ts` file placed in a gitignored dir (`temp/`) is not linted.
- A lint error in a tracked, non-ignored source file is still reported.
- Confirm `@eslint/compat` resolves the project `.gitignore` path correctly on
  Windows (the `fileURLToPath(new URL(...))` form, not a bare relative path).

## Dependencies

Supersedes the `temp/` ESLint-ignore line that was dropped from the #398 PR as
out of scope; this is the principled form of that fix.

wainwright1000:
`check:lint` ran `eslint . --max-warnings 0` over a hand-maintained `ignores` array that duplicated `.gitignore` (`dist/`, `node_modules/`, `coverage/`, `.netlify/`). The two could drift, and any gitignored dir that gained a lintable file (for example a Vite build artifact in `temp/`) broke `check:lint` until someone added it to the ESLint ignores by hand.

This uses `includeIgnoreFile` (bundled with ESLint as `eslint/config`, no new dependency) so `.gitignore` is the single source of truth for what is not linted. The manual `ignores` array now lists only paths that are *not* gitignored: `lilypond/build/` (tracked), `tests-local/`/`probes/` (local-only via `.git/info/exclude`), `.dependency-cruiser.cjs` (tracked).

Spec note: the ticket recommended `@eslint/compat`'s `includeIgnoreFile`, but that export is `@deprecated`; the same function ships bundled with ESLint as `eslint/config`, so no dependency is added.

Verification (an ESLint flat-config change has no unit-test harness here, so this is behavioural):

- `pnpm run ci` is green.
- A scratch `.ts` with a lint error placed in gitignored `temp/` goes from linted to ignored.
- A lint error in a tracked, non-ignored file is still reported.
- The "silently ignore everything" failure mode is not CI-invisible: `build:ohm` runs before `check:lint` and emits the ohm bundle into the linted `src/**`, so a broken `includeIgnoreFile` would fail `--max-warnings 0` loudly. (That guard is incidental on the `build:ohm`-before-`check:lint` ordering.)

Fixes #509

### DIFF
--- a/doc/CONTRIBUTING.md
+++ b/doc/CONTRIBUTING.md
@@ -247,6 +247,8 @@ The CI pipeline runs `pnpm run check:lint` and `pnpm run check:types`. Pull requ
 
 The project uses an ESLint flat config (`eslint.config.js`) with `typescript-eslint`. Some rules are relaxed because the codebase predates them (for example, `no-var` and `@typescript-eslint/no-explicit-any` are currently off). Follow the existing patterns in the file you are editing rather than refactoring legacy code to meet stricter rules.
 
+ESLint honours `.gitignore` (via `includeIgnoreFile`, bundled with ESLint as `eslint/config`), so gitignored paths are never linted. The `ignores` array in `eslint.config.js` lists only the few paths that are *not* gitignored (for example local-only `tests-local/` and `probes/`, or tracked `lilypond/build/`); the comment there records the current membership. To stop a path being linted, prefer adding it to `.gitignore`; reach for the ESLint `ignores` array only when the path must stay tracked.
+
 The project uses **Prettier** for automatic formatting. Run `pnpm run fix:format` before committing, or configure your editor to format on save. The CI pipeline runs `pnpm run check:format` as part of the build gate.
 
 All `.md` files must pass `markdownlint-cli2` before a PR is opened. Run `npx markdownlint-cli2 <file>` after editing any markdown file.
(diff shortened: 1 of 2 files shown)


wainwmr (approved):
OK

PR #580, opened 2026-06-12
fix: order pre-release LilyPond versions below their release (#519)

### LINKED ISSUE #519: compareVersions fails on pre-release LilyPond versions
## Description

`compareVersions` in `lilypond/build/buildScores.mjs` fails to handle pre-release
LilyPond versions correctly. When LilyPond reports a version such as
`2.26.0-rc1`, the comparison should treat it as less than the release version
`2.26.0`. In the current code, pre-release suffixes are silently truncated by
`parseLilypondVersion`, causing the minimum-version check to pass when it should
fail.

## Reachability

Developer-facing only. Triggered when a contributor or CI runner has a
pre-release LilyPond version installed that matches the minimum required version
numerically (e.g. `2.26.0-rc1` when the minimum is `2.26.0`).

## Current state

- `lilypond/build/buildScores.mjs:75-78` — `parseLilypondVersion` uses
  `/GNU LilyPond (\d+\.\d+\.\d+)/`, which matches only the numeric prefix and
  discards any pre-release suffix. Example:
  `"GNU LilyPond 2.26.0-rc1"` → `"2.26.0"`.

- `lilypond/build/buildScores.mjs:80-90` — `compareVersions` splits on `.` and
  calls `Number()` on each part. If a pre-release suffix ever reached this
  function (e.g. `"2.26.0-rc1"`), `Number("0-rc1")` yields `NaN`. `NaN < NaN`
  and `NaN > NaN` are both false, so the function returns `0` (equal) instead
  of `-1` (less than).

- `lilypond/build/buildScores.mjs:92-111` — `checkLilypond` calls
  `parseLilypondVersion`, then compares the result against `"2.26.0"` via
  `compareVersions`. Because the suffix is stripped, a pre-release of the
  minimum version passes silently.

## Root cause

Two-layer failure:

1. `parseLilypondVersion` truncates pre-release suffixes, so the bug in
   `compareVersions` is masked in normal operation.
2. `compareVersions` assumes every dot-separated part is purely numeric. It has
   no defence against non-numeric suffixes.

The truncation in layer 1 is itself a defect: it misrepresents the installed
version in error messages and allows pre-release versions to pass the minimum
version check.

## Scope

In scope:
- Fix `parseLilypondVersion` to capture the full version string.
- Fix `compareVersions` to handle non-numeric suffixes correctly.
- Add unit tests for `compareVersions`.

Out of scope:
- Full semver parsing (build metadata, multiple pre-release identifiers).
- Changing the minimum version threshold.

### Step 1: capture full version in `parseLilypondVersion`

Change the regex in `lilypond/build/buildScores.mjs:76` from:

```javascript
const match = output.match(/GNU LilyPond (\d+\.\d+\.\d+)/);
```

to:

```javascript
const match = output.match(/GNU LilyPond (\S+)/);
```

This captures the complete version token (e.g. `"2.26.0-rc1"`), preserving the
suffix for accurate comparison.

### Step 2: make `compareVersions` suffix-aware

Replace `compareVersions` (`lilypond/build/buildScores.mjs:80-90`) with:

```javascript
function compareVersions(a, b) {
  const parse = (v) =>
    v.split(".").map((p) => {
      const num = parseInt(p, 10) || 0;
      const rest = p.slice(String(num).length);
      return { num, rest };
    });
  const partsA = parse(a);
  const partsB = parse(b);
  for (let i = 0; i < Math.max(partsA.length, partsB.length); i++) {
    const pa = partsA[i] || { num: 0, rest: "" };
    const pb = partsB[i] || { num: 0, rest: "" };
    if (pa.num < pb.num) return -1;
    if (pa.num > pb.num) return 1;
    if (pa.rest < pb.rest) return -1;
    if (pa.rest > pb.rest) return 1;
  }
  return 0;
}
```

This preserves numeric comparison for the major/minor/patch parts, then falls
back to lexicographic comparison of any suffix. A release version (no suffix)
is therefore greater than its pre-release (has suffix), matching semver
semantics.

### Step 3: export `compareVersions` for testing

Add `compareVersions` to the export list at the bottom of
`lilypond/build/buildScores.mjs`:

```javascript
export { parseArgs, buildPattern, compareVersions };
```

## Test plan

Add a `compareVersions` describe block to
`lilypond/test/buildScores.test.ts`:

1. `compareVersions("2.26.0", "2.26.0")` returns `0`.
2. `compareVersions("2.25.0", "2.26.0")` returns `< 0`.
3. `compareVersions("2.27.0", "2.26.0")` returns `> 0`.
4. `compareVersions("2.26.0-rc1", "2.26.0")` returns `< 0` (pre-release is less
   than release).
5. `compareVersions("2.26.0-rc1", "2.26.0-rc2")` returns `< 0`.
6. `compareVersions("2.26.0", "2.26.0-rc1")` returns `> 0`.
7. `compareVersions("2.26.1-rc1", "2.26.0")` returns `> 0` (higher patch wins).

Run `pnpm test:unit lilypond/test/buildScores.test.ts` to confirm.

## Dependencies

None.

## Open questions

None.

## Follow-up items

None.

## Documentation impact

None expected. `doc/BUILD.md` mentions the LilyPond version requirement but does
not describe the comparison logic; no change warranted.

wainwright1000:
Fixes #519.

`compareVersions` split each dot-separated part with `Number()`, so a pre-release suffix like `0-rc1` became `NaN` and the comparison returned 0 (equal) — a pre-release of the minimum LilyPond version silently passed the build's version check. `parseLilypondVersion` also truncated the suffix, masking it.

## What changed

- **`parseLilypondVersion`**: regex widened from `\d+\.\d+\.\d+` to `\d+\.\d+\.\d+\S*` so it captures the pre-release suffix, while still requiring an X.Y.Z core — a non-version banner returns `null` and is reported as "unknown" with raw stdout echoed (preserving the #549 diagnostic; a bare `\S+` would wrongly accept garbage).
- **`compareVersions`**: rewritten to parse each part's leading digit run via `/^(\d*)(.*)$/` into `{num, rest}`. Integers compare numerically and a missing suffix outranks a present one (release > pre-release, per semver), so `2.26.0` > `2.26.0-rc1`. The explicit digit-run match also avoids mangling a digit-less or leading-zero part now that the comparator is exported.
- Both functions are now exported for testing.

## Tests

- **Unit** (`buildScores.test.ts`): pre-release ordering both directions (exact `-1`/`1`), higher-version pre-release accepted, suffix equality, a leading-zero discriminator, and a documented lexicographic-limitation case (`-rc10` < `-rc2`); plus `parseLilypondVersion` plain / suffix / garbage→null.
- **Integration** (`buildScores.integration.test.ts`): a pre-release of the minimum version (`2.26.0-rc1`) is rejected at the call site — non-zero exit, parsed not "unknown", "2.26.0 or later is required". This pins the bug end-to-end, so a `< 0` → `<= 0` refactor of `checkLilypond` cannot silently reintroduce it.

## Notes

Went through the Vera gate (one cycle, two passes; pass 2 clean). Two important findings addressed: the exported comparator's `{num, rest}` parse hardening (519-01) and the missing call-site rejection test (pr-test-analyzer) — see the #519 comments. Developer-facing build-script fix: no user-visible app change, no version bump.

- Claude

### DIFF
--- a/lilypond/build/buildScores.mjs
+++ b/lilypond/build/buildScores.mjs
@@ -73,18 +73,46 @@ function validateOptions(options) {
 }
 
 function parseLilypondVersion(output) {
-  const match = output.match(/GNU LilyPond (\d+\.\d+\.\d+)/);
+  // Capture the numeric core PLUS any pre-release suffix (e.g.
+  // "2.26.0-rc1"). The previous `\d+\.\d+\.\d+` silently truncated the
+  // suffix, which masked the pre-release ordering bug in compareVersions
+  // and let a pre-release of the minimum version pass the check (#519).
+  // The trailing `\S*` keeps the suffix while still requiring an X.Y.Z
+  // core, so a non-version banner returns null and is reported as
+  // "unknown" with the raw stdout echoed (#549) — a bare `\S+` would
+  // wrongly accept garbage tokens as a version.
+  const match = output.match(/GNU LilyPond (\d+\.\d+\.\d+\S*)/);
   return match ? match[1] : null;
 }
 
+// Compare two dot-separated version strings. Each part is split into its
+// leading integer run and any trailing suffix; integers compare numerically
+// and a part with no suffix outranks one with a suffix (release > pre-release,
+// per semver precedence), so "2.26.0" > "2.26.0-rc1". Suffixes among
+// themselves compare lexicographically, which orders multi-digit identifiers
+// wrongly — "-rc10" sorts BELOW "-rc2". Not a full semver implementation
+// (no build metadata, no multi-identifier pre-release ordering) — see #519.
 function compareVersions(a, b) {
-  const partsA = a.split(".").map(Number);
-  const partsB = b.split(".").map(Number);
+  const parse = (v) =>
+    v.split(".").map((p) => {
+      // Match the leading digit run explicitly so a part keeps its full
+      // suffix and a leading-zero part is not mis-sliced. The sole caller
+      // feeds X.Y.Z[suffix], but the exported comparator must not mangle a
+      // digit-less or leading-zero part (Vera 519-01).
+      const [, digits, rest] = /^(\d*)(.*)$/.exec(p);
+      return { num: digits === "" ? 0 : parseInt(digits, 10), rest };
+    });
+  const partsA = parse(a);
+  const partsB = parse(b);
   for (let i = 0; i < Math.max(partsA.length, partsB.length); i++) {
-    const numA = partsA[i] || 0;
-    const numB = partsB[i] || 0;
-    if (numA < numB) return -1;
-    if (numA > numB) return 1;
+    const pa = partsA[i] || { num: 0, rest: "" };
+    const pb = partsB[i] || { num: 0, rest: "" };
+    if (pa.num !== pb.num) return pa.num < pb.num ? -1 : 1;
+    if (pa.rest !== pb.rest) {
+      if (pa.rest === "") return 1;
+      if (pb.rest === "") return -1;
+      return pa.rest < pb.rest ? -1 : 1;
+    }
   }
   return 0;
 }
@@ -261,4 +289,4 @@ if (isMain) {
   main();
 }
 
-export { parseArgs, buildPattern };
+export { parseArgs, buildPattern, parseLilypondVersion, compareVersions };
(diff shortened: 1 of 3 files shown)


wainwmr:
This isn't going to happen.  Bloat.

PR #610, opened 2026-06-13
fix: replace Windows-breaking lilypond symlink with a Vite resolve alias

### LINKED ISSUE #607: Windows: tracked packages/pwa/src/lilypond symlink breaks the PWA test suite on Windows checkouts
A tracked git symlink breaks the PWA test suite on every Windows checkout. The fix below is agreed with Mark (verbal): single recommended change, straight to a PR.

## Symptom

On Windows, `pnpm --filter pwa run test:unit` fails: 10 test files error with `Failed to resolve import "../lilypond/spem.ly?raw" from src/ts/lily.ts`, cascading to `controls.pause is not a function` and similar. The PWA unit suite is unrunnable on a Windows checkout.

## Root cause

`packages/pwa/src/lilypond` is a tracked git **symlink** (`-> ../../scores/src/Hugh Keyte`). With Git's default `core.symlinks=false` on Windows it checks out as a ~27-byte **text file** containing the link path, so Vite cannot resolve `../lilypond/spem.ly`. Linux CI has real symlinks, so it is green there — the break is invisible to CI and to static analysis, and only shows when the suite is actually run on Windows. It pre-dates the monorepo; #604 just retargeted the symlink. It hits **any** Windows clone, not only our setup.

## Recommended fix

The PWA references the scores LilyPond source through exactly one import — `src/ts/lily.ts:8`, `import spem from "../lilypond/spem.ly?raw"`. Replace the filesystem symlink with a build-time Vite resolve alias:

1. Delete the tracked symlink `packages/pwa/src/lilypond`.
2. Add a named alias to `packages/pwa/vite.config.ts` `resolve.alias`, e.g. `@lilypond` -> `../scores/src/Hugh Keyte` (path resolved relative to the config file). Vitest reads the same config, so tests resolve too.
3. Change the import in `src/ts/lily.ts` to `import spem from "@lilypond/spem.ly?raw"`.
4. Carry a comment in `vite.config.ts` next to the alias explaining why it exists, so no one reintroduces a symlink:

   ```ts
   // The PWA parses spem.ly at runtime (src/ts/lily.ts) from the @spem/scores
   // package. We resolve it with this build-time alias rather than a tracked
   // symlink: a git symlink checks out as a plain text file on Windows
   // (core.symlinks = false by default), which breaks Vite resolution and the
   // PWA test suite there. The alias resolves identically on every OS with no
   // per-clone setup, so contributors can work on any platform, Windows included.
   ```

This removes the root cause (the tracked symlink), works identically on every platform, and needs no per-clone setup. Confirmed locally: with the symlink replaced (junction workaround) the PWA suite passes 22 files / 268 tests on Windows; the alias gives the same resolution without any filesystem object.

Note: the alias reaches into a sibling package's `src`, which is fine for two co-located workspace packages. If `@spem/scores` is ever published as a library, revisit it then (a separate tech-debt ticket, not this fix).

## Test plan

- `pnpm --filter pwa run test:unit` passes on Windows (the failing 10 files now resolve).
- `pnpm --filter pwa run build` and the e2e suite still pass on Linux CI (no regression).
- Grep confirms the symlink is gone and no other source path references `../lilypond`.

wainwright1000:
## What

Replace the tracked `packages/pwa/src/lilypond` git symlink with a build-time
Vite resolve alias (`@lilypond`).

## Why

On Windows (`core.symlinks=false` by default) the symlink checks out as a
~27-byte text file, so Vite cannot resolve `../lilypond/spem.ly?raw` and the PWA
unit suite is unrunnable there. Linux CI has real symlinks, so the break is
invisible to CI. It hits any Windows clone, not just our setup.

## How

- Delete the symlink `packages/pwa/src/lilypond`.
- Add `resolve.alias` `@lilypond` → `../scores/src/Hugh Keyte` in
  `packages/pwa/vite.config.ts`, with a comment explaining why there is no
  symlink. Vitest reads the same config, so tests resolve too.
- Change the import in `src/ts/lily.ts` to `@lilypond/spem.ly?raw`.
- Whitelist the alias namespace in `knip.json` so `check:unused` passes, and
  update the `server.fs.allow` comment that referenced the old symlink.

The alias resolves identically on every OS with no per-clone setup. Behaviour is
unchanged — the same score source file is read.

## Test

- Reproduced the Windows failure locally (degraded the symlink to its text-file
  form, got the exact `Failed to resolve import "../lilypond/spem.ly?raw"`
  error), then confirmed green after the fix.
- `pnpm --filter pwa run test:unit`: 22 files / 268 tests pass.
- `pnpm run ci`: green.

Fixes #607

- Claude

### DIFF
--- a/packages/pwa/knip.json
+++ b/packages/pwa/knip.json
@@ -2,5 +2,5 @@
   "$schema": "https://unpkg.com/knip@5/schema.json",
   "project": ["src/**/*.ts", "*.ts"],
   "entry": ["src/**/*.ts", "*.ts"],
-  "ignoreDependencies": ["autoprefixer", "lint-staged"]
+  "ignoreDependencies": ["autoprefixer", "lint-staged", "@lilypond/.+"]
 }
--- a/packages/pwa/src/lilypond
+++ b/packages/pwa/src/lilypond
@@ -1,1 +1,0 @@
-../../scores/src/Hugh Keyte
--- a/packages/pwa/src/ts/lily.ts
+++ b/packages/pwa/src/ts/lily.ts
@@ -5,7 +5,7 @@ import config from "./config";
 import lyGrammar from "../ohmjs/ly-grammar.ohm-bundle";
 import * as ohm from "ohm-js";
 import { Duration, BarLine, Note, Rest, Component } from "./music-classes";
-import spem from "../lilypond/spem.ly?raw";
+import spem from "@lilypond/spem.ly?raw";
 
 // A single note entry at a quantised bar position.
 export type NoteEntry = {
--- a/packages/pwa/vite.config.ts
+++ b/packages/pwa/vite.config.ts
@@ -30,6 +30,17 @@ const versionWithBranch =
   branch && branch !== "main" ? `${pkg.version}-${branch}` : pkg.version;
 
 export default defineConfig({
+  resolve: {
+    alias: {
+      // The PWA parses spem.ly at runtime (src/ts/lily.ts) from the @spem/scores
+      // package. We resolve it with this build-time alias rather than a tracked
+      // symlink: a git symlink checks out as a plain text file on Windows
+      // (core.symlinks = false by default), which breaks Vite resolution and the
+      // PWA test suite there. The alias resolves identically on every OS with no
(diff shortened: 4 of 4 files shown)


wainwmr (approved):
OK

PR #613, opened 2026-06-14
docs: add fix-worthiness Principles section to CONTRIBUTING.md

### LINKED ISSUE #595: docs: add fix-worthiness Principles section to CONTRIBUTING.md
Mark, we've been firming up the bar for what lands on the board, in two directions. Each is meant to stand on its own, and both build on things you've already said. Have a read and tell us where you'd change anything; we'll apply your amendments before rolling it out.

## 1. Tightening: what counts as a defect

- **Bug**: a real, user-facing defect that is broken *now*. It produces a wrong result a user can perceive, under a condition realistically reachable in normal use. (Sound that should play but doesn't counts. A failure only reachable by a one-pixel tap, or a corrupted build, does not.)
- **Tooling bug**: the same, but developer- or CI-facing rather than user-facing. A build script, test, or CI step actually broken now.
- **Tech debt**: real, but *not* broken. The code works, it could just be better. By default this isn't a board ticket on its own (see section 2 for the exception); we record it in our own notes and clean it up alongside adjacent work.

When we *do* fix something:

- Fix real bugs and tooling bugs, test-first; preserve existing behaviour unless changing it is the point.
- Take the simplest fix that doesn't cost quality; code needn't be perfect; follow the patterns already in the file rather than re-standardising legacy code; prefer incremental change to big refactors.
- Decline a fix whose cost is out of proportion to the harm, where cost includes new abstraction, infrastructure, complexity, ongoing maintenance, and the build/CI/Netlify usage you pay for. (This is the "unnecessary bloat" instinct, made explicit.)
- The one exception is a hard floor we never decline regardless of cost: data loss, security, or a crash on a common user path. Those get escalated, not lived with.

## 2. Loosening: tech debt that earns a board ticket

You said you're happy to take tech-debt tickets, as long as they **improve performance, simplify the code, or make it more maintainable**. We'd like to firm that up so the door is real but not a loophole. A tech-debt ticket qualifies if it clearly does one of these, demonstrably and now, not on a "might help later":

- **Performance**: a measurable win on something that's felt, such as app responsiveness, build and CI time and cost, or bundle size. Evidence is a number or a clear algorithmic improvement, not "should be faster".
- **Simplify**: a net reduction in complexity. Fewer lines, branches, abstractions, or indirection; dead code or a redundant layer removed. The diff comes out smaller and flatter, not just rearranged.
- **More maintainable**: a *specific* future change made safer, or a *specific* footgun removed. A compiler-enforced invariant, a fragile structure replaced by a robust one, coverage over a genuinely fragile area. "It feels cleaner" doesn't qualify; naming the future risk it removes does.

Tech debt that fits none of these stays in our notes and rides along with nearby work, rather than coming to you as a ticket.

## The ask

Would you amend, add to, or drop anything in either section? In particular, are those three tech-debt categories the right three, or would you frame them differently? Once you're happy, we'll write it into CONTRIBUTING and apply it.

---

## Specification summary

Mark ratified the principles in the comment below. This issue is now specified: the work adds a single canonical **Principles** section to `doc/CONTRIBUTING.md` and trims duplicate statements elsewhere to one-line references. See the re-specification comment for the full recommended fix, test plan, and dependencies.

wainwright1000:
Fixes #595.

Adds a `## Principles` section to `doc/CONTRIBUTING.md` recording the fix-worthiness bar: **what we fix** (real bugs, tooling bugs, and documentation-as-defect), **how much** (the simplest fix that doesn't cost quality), **when we decline** (cost disproportionate to harm), the **harm floor** (data loss / security / crash on a common path — never accepted), and which **tech debt earns a board ticket** (a demonstrable, now win in Performance, Simplify, or Maintainability).

These are the principles you ratified in the discussion on #595; this PR records that agreement in the contributor docs so it is the durable, citable reference for what lands on the board.

Docs-only — no code or behaviour change.

- Vera

### DIFF
--- a/doc/CONTRIBUTING.md
+++ b/doc/CONTRIBUTING.md
@@ -4,6 +4,50 @@
 
 See `doc/BUILD.md` for development setup, prerequisites, and build commands.
 
+## Principles
+
+These principles guide what we fix, how much we fix, and when we decline a fix. They are the bar Mark ratified for what lands on the board.
+
+### What we fix
+
+- **TDD**: a fix starts with the failing test.
+- **Fix real bugs and tooling bugs.** A bug is a real, user-facing defect that is broken *now* and realistically reachable in normal use. A tooling bug is the same, but contributor-/CI-facing rather than user-facing. Sound that should play but does not counts as user-facing; a failure only reachable by a corrupted build does not.
+- **Documentation is an artifact that can be broken.** Wrong, misleading, or out-of-date docs that lead a realistic reader to a wrong result are a defect: user-facing docs -> **Bug**; contributor / CI / build docs -> **Tooling bug** (Area Docs). Sparse-but-not-wrong docs, where better docs would only make a future change safer, are **Maintainable** tech debt, not a defect.
+- **Tech debt is not fixed on its own**; it rides along with adjacent real work.
+- Preserve existing user-facing behaviour unless changing it is the point of the work.
+
+### How much we fix
+
+- The simplest fix that does not cost quality.
+- Follow existing patterns in the file rather than re-standardising legacy code.
+- Prefer incremental changes over monolithic refactors.
+- Report scope creep rather than silently upgrading difficulty.
+
+### When we decline
+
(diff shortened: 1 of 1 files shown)


wainwmr (approved):
OK

PR #622, opened 2026-06-14
build: pin *.ohm to LF to fix Windows build:ohm CRLF phantom diff (#611)

### LINKED ISSUE #611: Windows: build:ohm bakes CRLF into tracked ly-grammar.ohm-bundle.js (phantom diff)
**Type:** tooling bug
**Area:** Tooling
**Difficulty:** XS

**Source:** `.gitattributes` (missing `*.ohm` rule); manifests in the tracked generated bundle `packages/pwa/src/ohmjs/ly-grammar.ohm-bundle.js`.

**First seen:** 2026-06-14, current `main`.

**Description:**
On Windows (`core.autocrlf=true`), `packages/pwa/src/ohmjs/ly-grammar.ohm` is checked out with **CRLF** line endings because `.gitattributes` has no rule for `*.ohm` (the repo stores it as LF). `pnpm build:ohm` reads the working-tree `.ohm` and **bakes its line endings into the generated bundle**: the `source` string literal inside `ly-grammar.ohm-bundle.js` becomes `"Lilypond { \r\n\r\n..."` instead of `"...\n\n..."`, and every `sourceInterval` offset shifts. The result is a **phantom diff on `ly-grammar.ohm-bundle.js` after every Windows build** (`pnpm build:ohm`, also invoked by `prebuild`/`check`), even from a clean tree. The `.d.ts` is unaffected, and the bundle's own newlines are already `eol=lf` — the divergence is purely the CRLF embedded in the source string.

Reproduce on Windows: clean tree -> `pnpm build:ohm` -> `git status` shows `ly-grammar.ohm-bundle.js` modified (HEAD has `\n`, rebuild has `\r\n`).

**Recommended fix:**
Add `*.ohm text eol=lf` to `.gitattributes` so the grammar source is always checked out LF (overriding `autocrlf`). The bundle (`*.js` / `*.d.ts`) is already covered by the existing `*.js` / `*.ts text eol=lf` rules and needs no new rule — once `.ohm` is LF, `build:ohm` regenerates the bundle identically. Apply with `git add --renormalize .`, rebuild, verify a clean `git status`.

**Test plan:**
On Windows: add the rule, refresh `ly-grammar.ohm` to LF, `pnpm build:ohm`, confirm `git status` clean (no bundle diff). CI (Linux, LF) already builds clean, so no regression.

**Dependencies:** none

wainwright1000:
Fixes #611.

On Windows (e.g. with `core.autocrlf=true`), `ly-grammar.ohm` was checked out with CRLF because `.gitattributes` had no `*.ohm` rule. `build:ohm` then baked those CRLFs into the `source` string literal of the generated `ly-grammar.ohm-bundle.js` (as escaped `\r\n` sequences, since Ohm embeds the grammar via `JSON.stringify`), producing a phantom diff on `git status` after every Windows build, even from a clean tree.

This pins `*.ohm` to LF (`*.ohm text eol=lf`), adds a cross-platform regression test (case 1: the rule is present; case 2: the committed bundle's source literal carries no `\r\n` escape), and documents the invariant in `doc/BUILD.md`. CI runs on Linux (LF) and cannot exhibit the bug, so the test is the guard. This is a `build:` change with no version bump.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -8,4 +8,5 @@
 *.scss text eol=lf
 *.css text eol=lf
 *.html text eol=lf
+*.ohm text eol=lf
 .husky/pre-commit text eol=lf
--- a/doc/BUILD.md
+++ b/doc/BUILD.md
@@ -172,6 +172,8 @@ When releasing, update `package.json` only. The build will propagate the new ver
 
 `pnpm run build:ohm` regenerates `src/ohmjs/ly-grammar.ohm-bundle.js` and `src/ohmjs/ly-grammar.ohm-bundle.d.ts` from `src/ohmjs/ly-grammar.ohm` via `@ohm-js/cli`. If you modify the grammar, rebuild before testing or deploying.
 
+`ly-grammar.ohm` is pinned to LF by `.gitattributes` (`*.ohm text eol=lf`) so `build:ohm` is deterministic across platforms. Without the rule, a Windows checkout (e.g. with `core.autocrlf=true`) writes the grammar with CRLF and `build:ohm` bakes those CRLFs into the `source` string literal of the generated bundle, producing a phantom diff on every Windows build, even from a clean tree (#611). The bundle file's own line endings are already LF via the existing `*.js` rule; the `*.ohm` rule is what stops CRLF being baked into that literal.
+
 ## Build Output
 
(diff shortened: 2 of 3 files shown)


wainwmr (approved):
OK

Issue #625, opened 2026-06-15
Re-enable @typescript-eslint/no-unused-expressions for test files (split from #621, needs sign-off)


wainwright1000:
## 2. Description

`packages/pwa/eslint.config.js` disables `@typescript-eslint/no-unused-expressions` for all test files via the `src/**/*.test.ts` override. With the rule off, a call-less assertion such as `expect(x).toBe;` (the `(value)` forgotten) is a valid expression statement that asserts nothing yet reads like a real test. Lint passes, the test suite passes, and the invariant is never verified. A real instance of this class was found via #541.

Re-enabling the rule for test files catches the syntactic class at lint time. This is a Maintainable guard that pins the otherwise-unverified invariant that a test's assertions are actually invoked.

## 3. Reachability

N/A — this is contributor-facing tooling, not a user-facing bug.

## 4. Current state

`packages/pwa/eslint.config.js` (lines 52-56)

```js
{
  files: ['src/**/*.test.ts'],
  rules: {
    '@typescript-eslint/no-unused-expressions': 'off',
    '@typescript-eslint/ban-ts-comment': 'off',
  },
}
```

The glob is package-relative: `check:lint` runs ESLint from inside `packages/pwa`, so `src/**/*.test.ts` resolves to `packages/pwa/src/**/*.test.ts` (the 23 current test files).

## 5. Root cause

The rule was disabled during Mark's flat-config migration. The relaxation is no longer justified: the spike below shows zero existing violations and no false positives, so the rule can be re-enabled without churn.

## 6. Scope

**In scope:**
- Remove `'@typescript-eslint/no-unused-expressions': 'off'` from the `src/**/*.test.ts` override.
- Leave `'@typescript-eslint/ban-ts-comment': 'off'` untouched (out of scope for this change).

**Out of scope:**
- Unifying the two-tier rule-override regime across `src/` and `e2e/` (handled in #621 as documented, not collapsed).
- Pinning `ecmaVersion: 'latest'` to a fixed version (tracked as Tech Debt 1 in `wiki/refactor-eslint.config.js.md`).

## 7. Recommended fix

Delete the `@typescript-eslint/no-unused-expressions` line from the test override block so the rule inherits the production-code setting (`['error', { allowTernary: true }]`) for test files as well.

## 8. Test plan

1. Remove the override line.
2. Run `pnpm --filter @spem/pwa check:lint` (equivalently `eslint .` from `packages/pwa`).
[...]

PR #626, opened 2026-06-15
build: drop unused type-aware ESLint parsing in pwa config (#621)

### LINKED ISSUE #621: Tidy packages/pwa/eslint.config.js (3 calibrated tech-debt items)
## Description

`packages/pwa/eslint.config.js` currently carries three calibrated tech-debt items from the #483/#484 bug-tightening walkthrough (Workbench Item [#524](https://github.com/wainwright1000/spem-tools/issues/524) Part C). They share one file/area and one coherent fix shape, so they are bundled into a single ticket per the coherent-unit guardrail.

1. **Unify the two-tier rule-override regime.**
   The src override block applies project relaxations (`no-var`, `@typescript-eslint/no-explicit-any`, `no-this-alias`, `no-unused-vars`) only to `src/**/*.ts` and `*.ts`. Files in `packages/pwa/e2e/` receive the stricter recommended defaults, creating an inconsistent lint regime. Collapse this to one consistent set of rules across all TypeScript files in the package.

2. **Remove `parserOptions: { project: true }`.**
   The config enables type-aware parsing but uses the type-unaware recommended set and enables no type-aware rules. The setting forces a TypeScript program build per linted file for zero benefit.

3. **Re-enable `@typescript-eslint/no-unused-expressions` for test files.**
   The rule is currently `off` for `src/**/*.test.ts`. This removes the guard that catches no-op/vacuous `expect(...)` expressions without call parentheses. A real instance was found via #541. Re-enabling it is a Mark-facing proposal: the rule was deliberately disabled during Mark's flat-config migration, so his sign-off is required before flipping it back on.

## Recommended fix

- Rewrite the `files` globs or override structure so the same rule relaxations apply uniformly to all TypeScript files under `packages/pwa` (including `e2e/`), or explicitly document why e2e should be stricter.
- Delete `parserOptions: { project: true }` from the config.
- Remove the `@typescript-eslint/no-unused-expressions: 'off'` override for test files; fix or explicitly exempt any resulting violations rather than disabling the rule wholesale.

## Test plan

- Run `pnpm run check:lint` (or `eslint .` from `packages/pwa`) before and after the change; it must remain green.
- After re-enabling `no-unused-expressions`, run the unit tests and inspect any new lint errors. Verify each is either fixed or covered by a narrow, justified exemption.
- Confirm that `e2e/` files are now linted under the same regime as `src/` files (for example, by running ESLint on a representative e2e file).

## Justification per item

1. **Simplify v2:** two divergent lint regimes (`src/` vs `e2e/`) become one consistent regime. The win is a countable reduction in distinct cases/variants, not a line-count reduction.
2. **Performance v2:** removing `parserOptions.project` eliminates a named waste (a TypeScript program build per linted file) with no consuming rule.
3. **Maintainable:** re-enabling the rule removes a specific way the codebase can silently accept no-op test assertions. The real instance found via #541 demonstrates the risk.

## Notes

- Item 3 re-enables a rule Mark deliberately disabled in his flat-config migration. Treat this part of the ticket as a proposal for Mark's sign-off, not a straightforward cleanup.
- The `ecmaVersion: 'latest'` drift and the dead monitor/scripts block are intentionally excluded from this ticket: the former fails the worth-it bar (Workbench Item [#524](https://github.com/wainwright1000/spem-tools/issues/524) Part C), the latter is already tracked as a tooling bug in #616.

- Kimi

wainwright1000:
## Summary

Removes the unused `parserOptions: { project: true }` from the `src/**/*.ts`
override in `packages/pwa/eslint.config.js`. The config uses the type-unaware
`recommended` set and enables no type-aware rule, so the option forced a
TypeScript program build per linted file for zero benefit. Lint behaviour is
unchanged; lint is faster.

Two clarifying comments are added: why `parserOptions.project` is intentionally
absent (re-add only if a type-aware rule is enabled), and why the rule
relaxations are scoped to legacy `src/` while `e2e/` keeps the stricter
recommended baseline.

## Scope (reshaped from the 3-item ticket)

- **Item 2 (parser removal)** — delivered.
- **Item 1 (unify the src/e2e regime)** — reframed from "collapse" to
  "document". `e2e/` is clean new code that passes the stricter defaults, so
  unifying would only loosen it; the divergence is justified and now documented
  in-config rather than collapsed.
- **Item 3 (re-enable `no-unused-expressions` on test files)** — split to #625
  as a proposal for your sign-off, since you disabled that rule deliberately in
  the flat-config migration and re-enabling may surface real violations.

## Test plan

- `pnpm --filter @spem/pwa check:lint` green before and after.
- `tsc --noEmit` green.
- Vera gate (5 reviewers): zero critical/important findings; one suggestion
  addressed (the removal-site comment). Reports on #621.

Fixes #621

- Claude

### DIFF
--- a/packages/pwa/eslint.config.js
+++ b/packages/pwa/eslint.config.js
@@ -15,6 +15,11 @@ export default tseslint.config(
   js.configs.recommended,
   ...tseslint.configs.recommended,
   eslintConfigPrettier,
+  // Relaxations below are scoped to legacy `src/` (and root `*.ts`) only: the
+  // codebase predates these conventions. Newer code such as `e2e/` is
+  // intentionally not listed here, so it keeps the shared recommended baseline
+  // rather than these relaxations. Do not widen this glob to unify the regimes;
+  // that would only loosen the newer code (#621).
   {
     files: ['src/**/*.ts', '*.ts'],
     languageOptions: {
@@ -24,9 +29,10 @@ export default tseslint.config(
         ...globals.browser,
         ...globals.es2021,
       },
-      parserOptions: {
-        project: true,
-      },
+      // No type-aware lint rules are enabled (the non-type-checked `recommended`
+      // set only), so `parserOptions.project` is intentionally omitted: it would
+      // build a TypeScript program per file for no consuming rule. Re-add it only
+      // if a type-aware rule is enabled (#621).
     },
     rules: {
       // Style rules: codebase predates these conventions


wainwmr (approved):
OK

PR #629, opened 2026-06-15
fix: prevent command injection in buildScores via execFileSync + option whitelist (#624)

### LINKED ISSUE #624: Command injection via execSync string interpolation in buildScores.mjs
## Type

tooling bug

## Area

Tooling

## Source

`packages/scores/build/buildScores.mjs:200` (`buildScore`)

## Description

`buildScore` calls `execSync` with a template string that interpolates user-controlled `--version` and `--notation` values into a shell command:

```js
execSync(`lilypond --svg -o "${outDir}/" "${ly}"`, { stdio: "inherit" });
```

`validateOptions` rejects bare boolean flags but does not whitelist allowed values or escape shell metacharacters. A crafted value containing quotes and command separators breaks out of the quoted `-o` argument and executes arbitrary shell commands. A probe replacing `lilypond` with `echo` shows `INJECTED_WINDOWS` executing on Windows cmd.exe when `--version` contains `" & echo INJECTED_WINDOWS & `.

## Recommended fix

1. Switch the `execSync` string call to `execFileSync("lilypond", ["--svg", "-o", `${outDir}/`, ly], { stdio: "inherit" })` so no shell interprets the arguments.
2. Alternatively, or additionally, have `validateOptions` whitelist `--version` to `"Hugh Keyte"` or `"OUP"` and `--notation` to `"early"` or `"modern"` (or null), matching the values it already documents in its error messages.

## Test plan

- Unit test that `buildScore` invokes `execFileSync` (or equivalent) with an array when building a score.
- Unit test that `validateOptions` rejects a `--version` or `--notation` value outside the allowed set.
- Probe test confirming shell metacharacters in `--version` do not execute injected commands.

wainwright1000:
Fixes #624.

`buildScores.mjs` invoked LilyPond via `execSync` with the output directory and
`.ly` path interpolated into a shell command string. `--version` and `--notation`
are user-controlled CLI flags that flow into the output path, so a crafted value
could break out of the quoted `-o` argument and run arbitrary shell commands.

## Fix

- Replace both `execSync` call sites with `execFileSync` (no shell), so every
  argument — `outDir` (from `--version`/`--notation`/`--outDir`), the `.ly` path,
  and the LilyPond binary — is passed literally and cannot inject. This is the
  actual injection guarantee.
- Whitelist `--version` (`Hugh Keyte`, `OUP`) and `--notation` (`early`, `modern`)
  values, which become output-directory path segments (defence in depth, not the
  injection barrier).
- Add a `LILYPOND_CMD` JSON-array override (`[binary, ...prefixArgs]`) for wrapped
  installs (`flatpak`, `wsl`) and the test fake — needed because `execFileSync`
  cannot run a `.bat`/script without a shell.

## Tests

- New `buildScores.security.test.ts`: proves `execFileSync` (not `execSync`) is
  used with an argument array and no `shell` option; a shell-metacharacter payload
  is rejected by the whitelist and passed literally on the unwhitelisted
  `--outDir` path; `LILYPOND_CMD` parse/rejection branches are covered.
- Integration test rewired from PATH-resolved `.bat`/`.sh` fakes to
  `LILYPOND_CMD=[node, helper]` (cross-platform, no shell), preserving all prior
  coverage. The scores suite (61 tests) and `pnpm run ci` pass.
- `doc/BUILD.md` documents the `execFileSync`/`LILYPOND_CMD` model.

Reviewed through the Vera gate (two passes); the cycle-1 findings — a version-check
stderr-leak regression, untested `LILYPOND_CMD` error paths, and comment/doc
accuracy on which defence is load-bearing — were addressed. See the VERA-624
report and synthesis comments on #624.

- Vera

### DIFF
--- a/doc/BUILD.md
+++ b/doc/BUILD.md
@@ -81,6 +81,8 @@ pnpm run build:scores -- --version="Hugh Keyte" --notation=early --choir="II B"
 
 This iterates over matching `Choir*.ly` files under `packages/scores/src/` and runs `lilypond --svg` for each, then post-processes the generated SVG with `packages/scores/build/postprocessSvg.mjs`.
 
+LilyPond is invoked via `execFileSync` (no shell), so no option value — `--version`, `--notation`, `--outDir`, the resolved `.ly` path, or `LILYPOND_CMD` — can inject shell commands: each is passed as a literal argument. Separately, `--version` and `--notation` are whitelisted to their known values because they become output-directory path segments. By default the `lilypond` binary on `PATH` is used; set the `LILYPOND_CMD` environment variable to a JSON array to point at a wrapped install — for example `LILYPOND_CMD='["flatpak","run","org.lilypond.LilyPond"]'` or `LILYPOND_CMD='["wsl","lilypond"]'` (POSIX shell syntax; in PowerShell use `$env:LILYPOND_CMD='[...]'`). The array is `[binary, ...prefixArgs]`; the build appends LilyPond's own arguments.
+
 ### Timing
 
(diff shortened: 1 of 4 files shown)


wainwmr (approved):
Excessive?  But OK.

wainwright1000:
Hm I don't think we're doing a very good job at getting the right level of information to you, (or me).  "excessive but ok" is actually a pretty good response but I can't see how you got there on the little information we sent.  let's have a chat.

PR #631, opened 2026-06-15
fix: seek canvas sections via ranges so keyboard nav stops at fractional bar boundaries

### LINKED ISSUE #598: Canvas keyboard navigation skips sections starting at fractional bars
1. Title
Canvas keyboard navigation skips sections starting at fractional bars

2. User-visible symptom
When a singer uses Ctrl/Cmd+Arrow to jump to the next or previous singing section, the playhead can skip over a section that begins between integer bar numbers and land in the wrong place.

3. Steps to reproduce
- Load the player and select any choir.
- Press Ctrl+Right (or Cmd+Right) to seek forward to the next section boundary.
- If the next section for that choir begins at a fractional bar position, the seek continues past it to the next integer-bar entry.

4. Expected behaviour
Seek stops at the first bar where the selected choir's singing state changes, even when that boundary starts at a fractional bar position.

5. Actual behaviour
Seek stops at the next integer bar index that has an entry in `notesByQuant`, missing boundaries that fall between integers.

6. Affected files
- `src/ts/MusicCanvas.ts:285-301` (`seek()`)

7. Root cause hypothesis
`seek()` increments an integer `intbar` and reads `this.notesByQuant.get(intbar)`. `notesByQuant` is keyed by fractional bar position, so entries at non-integer positions are never matched. The `ranges` Map is the correct data source for section boundaries.

8. Test code (failing)
```ts
// With notesByQuant containing an entry at bar 1.5 and a range [1.5, 3.0]
const pos = { choir: 0, part: "all", bar: 1 };
expect(canvas.seek(pos, +1)).toBeLessThan(3); // currently returns 3
```

9. Acceptance criteria
- A unit test with fractional section boundaries passes.
- Ctrl/Cmd+Arrow navigation in the full score lands at the actual singing/resting boundary, not the next integer-bar note.

10. Risk and scope
`seek()` is used only for keyboard navigation. Replacing the `notesByQuant` lookup with `ranges` may change the granularity of seek stops; clamping to `barCount` and lower bound must be preserved.

11. Notes
- Documented as unresolved in `wiki/refactor-MusicCanvas.ts.md` item 4 and `tests-local/architecture-canvas.md`.
- Related to [#244](https://github.com/wainwmr/spem-player/issues/244), which fixed a crash on undefined `notesByQuant` entries but did not change the data source.

wainwright1000:
## Summary

Canvas keyboard section-navigation (Cmd/Ctrl + Left/Right) skipped sections that begin on a fractional bar, stepping over them to the next whole-bar section.

`MusicCanvas.seek()` now derives section boundaries from the `ranges` map (which carries fractional quant positions) instead of scanning integer `notesByQuant` keys. It walks de-duplicated, sorted range edges and keeps only edges where the collective "any part of the choir singing" state flips, so it lands on fractional boundaries the old integer scan stepped over. Half-open `[from, to)` semantics and `[0, barCount]` clamping are preserved.

## Changes

- `packages/pwa/src/ts/MusicCanvas.ts` — rewrite `seek()` to use range edges; explicit `: number` return; comments on the strict-comparison "always advance" semantic and the midpoint flip-detection.
- `packages/pwa/src/test/canvas.test.ts` — tests for the fractional boundary, the collective-flip filter, the multi-part union, and shared-fixture save/restore; #244 tests re-pointed at `ranges`.

## Test plan

- Unit: new canvas seek tests (fractional boundary 1.5, collective-flip suppression, multi-part union) plus the existing suite.
- Manual: Cmd/Ctrl + Right/Left now stops at sections starting on fractional bars (verified in the dev build).

Reviewed through the Vera pre-publish gate; findings addressed or deferred (see issue comments).

Fixes #598

- Tele

### DIFF
--- a/packages/pwa/package.json
+++ b/packages/pwa/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@spem/pwa",
   "private": true,
-  "version": "2.8.4",
+  "version": "2.8.5",
   "description": "Helping singers learn Thomas Tallis Spem in alium",
   "type": "module",
   "dependencies": {
--- a/packages/pwa/src/test/canvas.test.ts
+++ b/packages/pwa/src/test/canvas.test.ts
@@ -17,6 +17,17 @@ describe("MusicCanvas custom element", () => {
     vi.clearAllMocks();
   });
 
+  // seek() tests overwrite choir ranges on the shared beforeAll canvas; snapshot
+  // and restore the map per test so a fixture never leaks into a later test (#598).
+  let savedRanges: NonNullable<typeof canvas>["ranges"];
+  beforeEach(() => {
+    savedRanges = new Map();
+    for (const [key, value] of canvas!.ranges) savedRanges.set(key, [...value]);
+  });
+  afterEach(() => {
+    canvas!.ranges = savedRanges;
+  });
+
   it("Check that the canvasent contains a canvas", async () => {
     expect(canvas).not.toBeNull();
     expect(canvas?.querySelector("canvas")).not.toBe(null);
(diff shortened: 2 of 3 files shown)


wainwmr (approved):
Ok

PR #632, opened 2026-06-15
build: resolve .worktree-offset by walking up to the worktree root

### LINKED ISSUE #620: Per-worktree dev/preview port offset not applied: .worktree-offset in wrong location after monorepo move
## Symptom
`pnpm dev` / `pnpm preview` in any non-default worktree binds the base ports (5173 / 4173) instead of the per-worktree offset ports, so multiple agent worktrees collide on the same port.

## Root cause
`packages/pwa/worktree-ports.ts` resolves the offset from (1) `SPEM_PORT_OFFSET` env, then (2) a `.worktree-offset` file **beside the module** (i.e. `packages/pwa/.worktree-offset`). After the monorepo restructure (#599/#600) moved `worktree-ports.ts` into `packages/pwa/`, the actual `.worktree-offset` file was left at the **worktree root** (`~/Spem/spem-<agent>/.worktree-offset`). So `packages/pwa/.worktree-offset` is absent, the file lookup misses, and the offset defaults to 0. Tool-spawned shells don't set `SPEM_PORT_OFFSET` (env not reliably inherited per ARCH § cwd is reliable, the environment is not), so the env path doesn't save it either.

## Reachability
Observed live in spem-vera (offset 40 at the worktree root): `pnpm dev` bound 5173, not 5213; setting `SPEM_PORT_OFFSET=40` explicitly bound 5213, confirming the file path is the broken one.

## Fix options
1. Have `worktree-ports.ts` resolve `.worktree-offset` from the worktree root (walk up from the module), keeping the file where the restructure left it; or
2. Move/relink each worktree's `.worktree-offset` to `packages/pwa/` beside the module.
Option 1 is more robust (survives future moves). Update the file-location note in `worktree-ports.ts`'s header comment either way.

Found during the #569 publish UI sign-off (Vera).

- Vera

wainwright1000:
## Summary

After the monorepo move, `packages/pwa/worktree-ports.ts` read `.worktree-offset` only from beside the module (`packages/pwa/`), but each worktree keeps its offset file at its **root**. So the offset was never found and every worktree fell back to the default ports (5173/4173), colliding across worktrees.

The resolver now **walks up** from the module directory to find `.worktree-offset`, stopping at the worktree root (the directory containing `.git`, file or dir, covering linked worktrees), with a filesystem-root backstop that yields offset 0. First match wins (nearest-wins).

## Changes

- `packages/pwa/worktree-ports.ts` — walk-up resolution with a worktree-root ceiling; header/JSDoc updated to describe the bounded walk; agent-name table removed in favour of a generic, name-free explanation.
- `packages/pwa/src/test/worktree-ports.test.ts` — hermetic tests: root-termination, present-but-unreadable, nearest-wins, root-inclusion.
- `doc/BUILD.md` — each checkout keeps its offset locally in its `.worktree-offset`; no central registry.

## Test plan

- Unit: `worktree-ports.test.ts` (root ceiling, directory-named file, nearest-wins, root-inclusion), hermetic via a `.git` marker.

Reviewed through the Vera pre-publish gate; all findings addressed or rejected with reason (see issue comments).

Fixes #620

- Tele

### DIFF
--- a/doc/BUILD.md
+++ b/doc/BUILD.md
@@ -23,19 +23,15 @@ This serves the application locally with hot module replacement. The `--host` fl
 
 ### Parallel worktrees
 
-Multiple worktrees can run dev and preview servers simultaneously by setting
-`SPEM_PORT_OFFSET` in each worktree's environment (for example, in a
-`.env.local` file). The offset is added to the default ports:
+Multiple worktrees can run dev and preview servers at once without colliding.
+Each declares a port offset, added to the base ports (dev 5173, preview 4173).
+`worktree-ports.ts` resolves it in order: `SPEM_PORT_OFFSET` if set, otherwise a
+gitignored `.worktree-offset` file (just the number) at the worktree root,
+found by walking up from the module. A checkout with neither (CI, a fork, the
+main checkout) gets offset 0 and the base ports. Each checkout keeps its own
+offset locally in its gitignored `.worktree-offset`; there is no central registry.
 
-| Worktree | Offset | Dev port | Preview port |
-| -------- | ------ | -------- | ------------ |
-| default  | 0      | 5173     | 4173         |
-| vera     | 100    | 5273     | 4273         |
-| kimi     | 200    | 5373     | 4373         |
-| copilot  | 300    | 5473     | 4473         |
-| tele     | 400    | 5573     | 4573         |
-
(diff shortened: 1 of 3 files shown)


wainwmr (approved):
Ok

PR #662, opened 2026-06-18
fix: play/pause button keyboard-activatable via <button> (#634)

### LINKED ISSUE #634: MusicControls: play/pause button (div) not keyboard-activatable when focused
**Source:** `packages/pwa/src/ts/MusicControls.ts:39-42,143` (the `playpausebutton` `<div>` and its sole `click` handler); `packages/pwa/index.ts:255-256` (the `.control` keyboard guard).

**Description:** The play/pause button is a `<div id="playpausebutton" tabindex="0" class="control">` with a `click` handler only and no `keydown` handler. A `<div>` does not synthesise a click on Enter/Space (unlike `<button>`), and the global keyboard handler in `index.ts` returns for any element with class `control` on every key except Escape (`index.ts:255-256`), before reaching the Space/Enter to play/pause path (`index.ts:306-313`). So when a keyboard user Tabs to the play/pause button (it is `tabindex="0"`, in the tab order) and presses a key while it is focused:

- **Enter** does nothing (no toggle).
- **Space** scrolls the page (the `.control` return at line 256 happens before the SCROLL_KEYS `preventDefault` at line 275) and does not toggle playback.

It is also a generic `<div>` with no `role`, so assistive technology announces it as a generic focusable element rather than a button (WCAG 2.1.1 Keyboard; 4.1.2 Name, Role, Value). Confirmed by code inspection and manual repro; the failure is structural, so it is identical in Chromium, Firefox, and WebKit.

Note: play/pause is still operable via the global Space/Enter shortcut when focus is on the page body (`index.ts:306-313`); the defect is specifically the focusable button not activating when it is focused (plus the Space-scroll and the missing button role). The other `.control` widgets (recording/score/dark/help) are `tabindex="-1"` (out of the tab order, shortcut-only by design), so play/pause is the only tabbable-but-not-activatable one.

**Recommended fix:** make `playpausebutton` a `<button>`. The existing `click` handler then receives Enter/Space natively, the element gains the button role, and it can carry an accessible name (aria-label or text). Confirm Space no longer scrolls when the button is focused, and that the `tabindex="-1"` shortcut-only intent for the other controls still holds.

**Test plan:** e2e (Playwright, all three engines): focus the play/pause button (Tab or `.focus()`), press Enter and Space, assert the play state toggles and the page does not scroll. Unit: assert the control is a `<button>` with an accessible name.

**Dependencies:** none. Resolves Workbench Item wainwright1000/spem-tools#446.

- Vera

wainwright1000:
Fixes #634

The play/pause control in `MusicControls` was a `<div>`: focusable but not
keyboard-activatable, so Enter/Space did nothing once it had focus. This swaps it
to a native `<button>`, restoring keyboard activation while preserving the
existing appearance and click behaviour.

## Tests
- `src/test/controls.test.ts`, `src/test/keyboard.test.ts`, `src/test/main.test.ts`
  — unit coverage for keyboard activation of the play/pause control.
- `e2e/playback-toggle.spec.ts` — end-to-end keyboard toggle of playback.

### DIFF
--- a/packages/pwa/e2e/playback-toggle.spec.ts
+++ b/packages/pwa/e2e/playback-toggle.spec.ts
@@ -51,6 +51,58 @@ test.describe("Playback keyboard toggle", () => {
     await expect(pauseIcon).toBeHidden();
   });
 
+  test("focused play/pause button activates on Enter (#634)", async ({ page }) => {
+    await page.goto("/");
+
+    await page.evaluate(() => {
+      HTMLAudioElement.prototype.play = function () {
+        (this as HTMLAudioElement).dispatchEvent(new Event("play"));
+        return Promise.resolve();
+      };
+    });
+
+    const controls = page.locator("music-controls");
+    const pauseIcon = controls.locator("#playpausebutton svg").nth(1);
+
+    await page.locator("#playpausebutton").focus();
+    await expect(pauseIcon).toBeHidden();
+
+    await page.keyboard.press("Enter");
+    await expect(pauseIcon).toBeVisible();
+
+    await page.keyboard.press("Enter");
+    await expect(pauseIcon).toBeHidden();
+  });
+
+  test("focused play/pause button activates on Space without scrolling (#634)", async ({
+    page,
+  }) => {
+    await page.goto("/");
+
+    await page.evaluate(() => {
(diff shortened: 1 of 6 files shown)


wainwmr (approved):
All good

PR #667, opened 2026-06-18
build: guard build:ohm against CRLF ly-grammar.ohm working tree (#648)

### LINKED ISSUE #648: Guard build:ohm against CRLF ly-grammar.ohm working tree (#611 recurrence)
**Type:** tooling bug
**Area:** Tooling
**Difficulty:** XS

**Source:** `packages/pwa/package.json` script `build:ohm`; tracked generated bundle `packages/pwa/src/ohmjs/ly-grammar.ohm-bundle.js`.

**Description:**
#611 added `*.ohm text eol=lf` to `.gitattributes`, which stops new Windows checkouts of `ly-grammar.ohm` from being CRLF. However, worktrees that already held a CRLF working copy continue to hold it: git normalizes CRLF to LF when comparing against the blob, so the file reads as clean and `git checkout --` never rewrites it. `git add --renormalize` also leaves the working tree untouched.

When an agent in such a worktree runs `pnpm run build:ohm`, the CRLF source is baked into the `source` literal of `ly-grammar.ohm-bundle.js` as `\\r\\n`, and `pnpm run ci` fails the #611 ohm-eol test locally.

Local fix: `rm packages/pwa/src/ohmjs/ly-grammar.ohm && git checkout -- packages/pwa/src/ohmjs/ly-grammar.ohm`.

**Recommended fix:**
Add a tiny normalization step to the `build:ohm` script so it ensures `src/ohmjs/ly-grammar.ohm` is LF before invoking `ohm generateBundles`. For example, prepend a Node one-liner that rewrites CRLF to LF, or extract it into a small helper script. This keeps the guard automatic and self-contained.

**Test plan:**
On a Windows worktree with a CRLF `ly-grammar.ohm`, run `pnpm run build:ohm`; confirm the generated bundle has no `\\r\\n` escape and `git status` shows no diff.

wainwright1000:
## Summary

Guard `build:ohm` against a CRLF `ly-grammar.ohm` working copy (a #611 recurrence).

`.gitattributes` (`*.ohm text eol=lf`, added in #611) only fixes *fresh* checkouts. A worktree that already held a CRLF `ly-grammar.ohm` keeps it: git reads it as clean against the LF blob, so `git checkout --` and `git add --renormalise` never rewrite it. `pnpm run build:ohm` then bakes the CRLFs into the `source` literal of the generated bundle as `\r\n`, and the #611 ohm-eol test fails locally on Windows even from a clean tree.

## Fix

`build:ohm` now normalises `packages/pwa/src/ohmjs/ly-grammar.ohm` to LF (via `packages/pwa/build/build-ohm.mjs`) before invoking `@ohm-js/cli generateBundles`, so the build is deterministic regardless of the working copy's line endings.

## Test plan

`packages/pwa/src/test/ohm-build-normalise.test.ts` covers the normalisation. On a Windows worktree with a CRLF `ly-grammar.ohm`, `pnpm run build:ohm` now produces a bundle with no `\r\n` escape and `git status` shows no diff.

`doc/BUILD.md` updated to describe the new normalisation step.

Fixes #648

### DIFF
--- a/doc/BUILD.md
+++ b/doc/BUILD.md
@@ -168,9 +168,9 @@ When releasing, update `package.json` only. The build will propagate the new ver
 
 ### Ohm Grammar
 
-`pnpm run build:ohm` regenerates `packages/pwa/src/ohmjs/ly-grammar.ohm-bundle.js` and `packages/pwa/src/ohmjs/ly-grammar.ohm-bundle.d.ts` from `packages/pwa/src/ohmjs/ly-grammar.ohm` via `@ohm-js/cli`. If you modify the grammar, rebuild before testing or deploying.
+`pnpm run build:ohm` first normalises `packages/pwa/src/ohmjs/ly-grammar.ohm` to LF (via `packages/pwa/build/build-ohm.mjs`), then regenerates `packages/pwa/src/ohmjs/ly-grammar.ohm-bundle.js` and `packages/pwa/src/ohmjs/ly-grammar.ohm-bundle.d.ts` from it via `@ohm-js/cli`. If you modify the grammar, rebuild before testing or deploying.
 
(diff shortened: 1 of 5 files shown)


wainwmr (approved):
Windows sucks

PR #680, opened 2026-06-19
perf: throttle idle canvas shimmer redraw while paused (#649)

### LINKED ISSUE #649: Throttle idle canvas redraw while paused to cut CPU/battery waste
## Description

While the player is paused (its common idle state), `MusicCanvas` redraws the full canvas scene on every `requestAnimationFrame` tick. The only visible change between frames is the breathing opacity sine on the false-relation hotspots; the static voice-part grid, bar highlight, and selection highlight are re-rendered identically roughly 60 times per second. This wastes CPU and battery on the idle path.

## Current state

`packages/pwa/src/ts/MusicCanvas.ts`

- `#startShimmerLoop()` (lines 126-134) schedules a `requestAnimationFrame` loop that calls `draw()` whenever `!this.playing`.
- `draw()` (lines 350-366) runs the full pipeline: `#clearCanvas`, `#drawBarHighlight`, `#drawSelectionHighlight`, `#drawVoiceParts` (40 choir×part rows, every range segment stroked), `#drawDev`, `#drawFalseRelationHotspot`, and `#drawFalseRelationPulses`.
- The false-relation hotspot opacity is the only paused-state animation; everything else is static.

## Root cause

The shimmer loop has no throttle and no early-exit when there is nothing to animate, so it repaints the whole scene at display refresh rate while paused.

## Scope

**In scope:**
- Throttle the paused shimmer loop so it does not redraw the full scene on every rAF tick.
- Skip the shimmer loop entirely when `frLocations` is empty (no false-relation hotspots to animate).
- Add unit tests covering the throttle and the empty-`frLocations` path.

**Out of scope:**
- Full offscreen-cache or static-layer separation. That is a larger refactor and is deferred.
- Changing the playback loop (handled in #554).

## Recommended fix

1. Add a static throttle constant and a timestamp field to `MusicCanvas`:

   ```typescript
   static readonly SHIMMER_IDLE_INTERVAL_MS = 100; // ~10 fps is enough for breathing shimmer
   #lastShimmerDraw = 0;
   ```

2. In `#startShimmerLoop()`, guard the draw call:

   ```typescript
   #startShimmerLoop() {
     if (this.frLocations.length === 0) return;
     const loop = (timestamp: number) => {
       if (!this.playing) {
         if (timestamp - this.#lastShimmerDraw >= MusicCanvas.SHIMMER_IDLE_INTERVAL_MS) {
           this.draw();
           this.#lastShimmerDraw = timestamp;
         }
       }
       this.shimmerLoopId = requestAnimationFrame(loop);
     };
     this.shimmerLoopId = requestAnimationFrame(loop);
   }
   ```

3. Use the rAF timestamp argument rather than `Date.now()` so the guard is tied to the animation frame clock.

4. Keep the existing behaviour that cancelling/restarting the loop on play/pause and disconnect/reconnect works unchanged.

## Test plan

Extend `packages/pwa/src/test/canvas.test.ts`:

1. **Paused shimmer is throttled.** Create a fresh `music-canvas`, mock `requestAnimationFrame` to collect callbacks, mock `Date.now()`/`performance.now()` if needed, and drain 12 simulated rAF ticks. Assert that the number of full-canvas clears (`fillRect(0, 0, width, height)`) is greater than 0 but less than 12 (e.g. 1-2 for a 100 ms throttle).
2. **Empty `frLocations` skips the shimmer loop.** Create a fresh element with `frLocations = []` before the loop starts (or after init, cancel the loop and reconnect), and assert that no shimmer-loop rAF callback is scheduled.
3. **Existing playback-loop and reconnect tests still pass.** Run the full `canvas.test.ts` suite.

## Dependencies

- #554 (playback-loop throttle) is related but already merged; this ticket concerns the paused shimmer loop only.

## Open questions

None.

## Follow-up items

- Full static-layer caching for the canvas can be investigated if the throttle proves insufficient. Not filed now.

## Documentation impact

None expected.

wainwright1000:
Fixes #649

Throttles the idle canvas redraw (the "shimmer") while playback is paused, cutting CPU/battery waste with no change to visible behaviour.

- `packages/pwa/src/ts/MusicCanvas.ts`: throttle the idle redraw loop while paused/idle.
- `packages/pwa/src/test/canvas.test.ts`: cover the throttled path.

### DIFF
--- a/packages/pwa/src/test/canvas.test.ts
+++ b/packages/pwa/src/test/canvas.test.ts
@@ -225,6 +225,125 @@ describe("MusicCanvas custom element", () => {
     canvas!.draw();
   });
 
+  it("paused shimmer loop throttles its idle redraws (#649)", () => {
+    // The paused shimmer loop used to call draw() on every rAF tick (the
+    // display refresh rate, 60-120Hz+) even though only the FR-hotspot breathing
+    // changes while paused. The throttle must drop most idle redraws. Mirrors
+    // the #554 play-loop test:
+    // mock rAF to collect callbacks, drive ticks manually, and count effective
+    // renders by the one full-canvas fillRect(0,0,w,h) each draw() issues.
+    //
+    // The whole body stays synchronous: a fresh canvas's #init runs to
+    // completion during appendChild (no awaits in #init), scheduling the first
+    // shimmer callback into the mock. Staying synchronous means the real rAF
+    // never fires, so the shared fixture's live shimmer loop cannot reschedule
+    // into this mock and pollute the count.
+    const rAFCallbacks: FrameRequestCallback[] = [];
+    let rafId = 0;
+    const rafSpy = vi
+      .spyOn(window, "requestAnimationFrame")
+      .mockImplementation((cb: FrameRequestCallback) => {
+        rAFCallbacks.push(cb);
+        return ++rafId;
+      });
+
+    const fresh = document.createElement("music-canvas") as MusicCanvas;
+    let fullClears: number | undefined;
+    try {
+      document.body.appendChild(fresh);
+      // Preconditions: #init completed, the loop will actually run, and draw()
+      // will not early-return on empty data (a dead fixture would report 0
+      // clears and masquerade as a throttle).
+      expect(fresh.frLocations.length).toBeGreaterThan(0);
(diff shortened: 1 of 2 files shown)


wainwmr (approved):
All good

PR #713, opened 2026-06-24
fix: pulse sub-1/16 note onsets in canvas (#704)

### LINKED ISSUE #704: Canvas pulse animation drops 16th-note onsets that land off the 1/16 quant grid
## Description

During playback the canvas overview pulses the choir/part row while a note is sounding. For notes whose onset falls between the 1/16-bar grid points used by `MusicCanvas.#updatePulses()`, the pulse is never triggered, so the note appears silently on the canvas.

The shipped scores contain the figure `a16 \ficta bes a8` (Hugh Keyte bar ~1237, OUP bar ~1228). The duration-inherited `bes` is a second 16th note that onsets at a +1/32-bar position. Because `#updatePulses()` only queries `notesByQuant` at 1/16-bar quant positions, that onset is missed.

## Reachability

1. Load the player.
2. Start playback (or scrub) so the playhead reaches the bar containing `a16 \ficta bes a8` in either edition.
3. Observe the canvas overview: the second 16th (`bes`) does not receive a pulse highlight.

The figure is present in both shipped `spem.ly` files, so the failure is reachable in production without malformed input.

## Current state

`packages/pwa/src/ts/MusicCanvas.ts`, `#updatePulses()` (around line 360).

```ts
const quant = Math.floor(this.bar * 16) / 16;
const notes = this.notesByQuant.get(quant);
```

`this.notesByQuant` is keyed by the raw fractional bar position produced in `packages/pwa/src/ts/lily.ts:processLilypond()` (lines 343-350). That producer advances position by `comp.duration.sfths / 128` per component, so a 16th note (`sfths = 4`, defined in `packages/pwa/src/ts/music-classes.ts:33`) advances by `1/32` bar. The second 16th of a pair therefore onsets at a key such as `x + 1/32`, which `Math.floor(bar * 16) / 16` can never produce.

## Root cause

The pulse consumer and the LilyPond producer disagree on the time grid. The producer stores onsets at the native `sfths/128` resolution (and at exact `pos` values for tuplets/scaled durations). The consumer rounds the current bar to a coarser 1/16 grid, so any onset not on a 1/16 boundary is invisible to the pulse lookup.

This is the same class of mismatch that #598 fixed for `seek()` by switching to the `ranges` edge walk; the pulse path was left on the old coarse grid.

## Scope

In scope:
- Change `#updatePulses()` to query `notesByQuant` at the producer's native 1/128 resolution.
- Add a unit test in `packages/pwa/src/test/canvas.test.ts` that verifies a note at a +1/32 (or other sub-1/16) position is pulsed.
- Verify no regression in existing canvas pulse tests.

Out of scope:
- Refactoring the producer to use a different keying scheme.
- Fixing the separate 1/16 quantization in `lily.ts detectFalseRelations()` (refactor-report item; affects false-relation hotspots, not pulse animation).
- General `MusicCanvas` helper extraction or lifecycle refactors.

## Recommended fix

1. In `packages/pwa/src/ts/MusicCanvas.ts`, `#updatePulses()`:
   - Replace the 1/16 quant computation with a 1/128 quant:
     ```ts
     const quant = Math.floor(this.bar * 128) / 128;
     ```
   - Leave the rest of the function unchanged. `lastNoteDuration` is already in `sfths/128` units, so the pulse fade timing remains consistent.

2. If the change makes the magic numbers `16`/`128` less self-documenting, consider adding a named constant local to the function (for example `const PULSE_QUANT = 128;`) or a short comment explaining that the quant must match the `sfths/128` resolution used by `processLilypond()`. This is optional; the surrounding code already uses `sfths/128` elsewhere.

## Test plan

Extend `packages/pwa/src/test/canvas.test.ts`.

1. **Sub-1/16 onset is pulsed.**
   - Construct a `MusicCanvas` instance and populate its `notesByQuant` with one note at a position that is not a multiple of `1/16`, for example `0.03125` (1/32 bar).
   - Set `canvas.bar` to that position.
   - Call a method that exercises `#updatePulses()` (for example `canvas.draw()` or, if the test file already accesses the private method, call `#updatePulses()` directly).
   - Assert that the corresponding `canvas.pulses[c][p]` is not the idle value `1` (or assert that `lastNoteStart[c][p]` equals the note onset and `lastNoteDuration[c][p]` is set).
   - Keep the test deterministic: use a synthetic `notesByQuant` rather than relying on the full `spem.ly` parse, so the exact onset is under the test's control.

2. **Existing 1/16 and whole-note onsets still pulse.**
   - Add or reuse a case with a note on a 1/16 boundary and a note on a whole-bar boundary to confirm the finer quant does not break previously working onsets.

3. **Regression.** Run `pnpm --filter @spem/pwa test:unit` and confirm all `canvas.test.ts` tests pass.

No Playwright case is required: the failure is a data-lookup mismatch that jsdom can exercise by inspecting `pulses`/`lastNoteStart`; no browser chrome, focus, or rAF timing is involved.

## Dependencies

- Residual of #598 (closed).
- None blocking.

## Open questions

None.

## Follow-up items

None. (The separate 1/16 quantization in `lily.ts detectFalseRelations()` is recorded in `wiki/refactor-lily.ts.md` item 18 and is not adjacent to this pulse fix.)

## Documentation impact

None expected. The change is internal to `MusicCanvas` pulse lookup and does not alter component conventions, build steps, or test commands.

- KIMI

wainwright1000:
## What

`MusicCanvas.#updatePulses()` looked up note onsets on a 1/16-bar grid (`Math.floor(this.bar * 16) / 16`), but `notesByQuant` is keyed at the producer's native 1/128 resolution. Any onset that lands between 1/16 points — the second of a 16th-note pair, e.g. `a16 bes a8` — was never matched, so its canvas pulse never fired. This changes the lookup to 1/128 so those onsets pulse.

```
- const quant = Math.floor(this.bar * 16) / 16;
+ const quant = Math.floor(this.bar * 128) / 128;
```

Fixes #704

## Tests

Two unit tests in `canvas.test.ts`: a note onsetting at 1/32 bar (off the old grid) now pulses (the regression witness — fails on `main`, passes here), and a no-regression check that 1/16 and whole-bar onsets still pulse under the finer quant. `notesByQuant` is snapshot/restored per test, mirroring the existing `ranges` guard (#598). Full `pnpm ci` green.

## Note for review (trade-off)

A Vera review flagged a second-order effect worth your eye: the 1/128 exact-match narrows the playback onset-capture window from ~8 frames to ~1 frame at peak tempo, so an on-grid pulse can be dropped if a frame is lost during the fastest passages (0% at a steady 60fps; the cost is one missing decorative highlight on the overview, not audio or score). It is a net improvement over `main` and was assessed as bounded tech debt rather than a blocking defect. The robust fix — detecting onsets by scanning the bar interval between frames (the `seek()`/`ranges` edge-walk pattern from #598) — is recorded as ride-along tech debt in the MusicCanvas refactor report (it qualifies under our tech-debt test but the harm removed doesn't earn a standalone ticket), to be applied next time this path is worked.

- Claude

### DIFF
--- a/packages/pwa/package.json
+++ b/packages/pwa/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@spem/pwa",
   "private": true,
-  "version": "2.8.6",
+  "version": "2.8.7",
   "description": "Helping singers learn Thomas Tallis Spem in alium",
   "type": "module",
   "dependencies": {
--- a/packages/pwa/src/test/canvas.test.ts
+++ b/packages/pwa/src/test/canvas.test.ts
@@ -1,6 +1,7 @@
 import { MusicCanvas } from "../ts/MusicCanvas";
 import config from "../ts/config";
 import type { Position } from "../ts/common";
+import { Duration, Note } from "../ts/music-classes";
 MusicCanvas.define("music-canvas");
 
 var canvas: MusicCanvas | null;
@@ -19,13 +20,17 @@ describe("MusicCanvas custom element", () => {
 
   // seek() tests overwrite choir ranges on the shared beforeAll canvas; snapshot
   // and restore the map per test so a fixture never leaks into a later test (#598).
+  // #704 also mutates notesByQuant, so snapshot that too.
   let savedRanges: NonNullable<typeof canvas>["ranges"];
+  let savedNotesByQuant: NonNullable<typeof canvas>["notesByQuant"];
   beforeEach(() => {
     savedRanges = new Map();
     for (const [key, value] of canvas!.ranges) savedRanges.set(key, [...value]);
+    savedNotesByQuant = canvas!.notesByQuant;
   });
   afterEach(() => {
     canvas!.ranges = savedRanges;
+    canvas!.notesByQuant = savedNotesByQuant;
   });
 
   it("Check that the canvasent contains a canvas", async () => {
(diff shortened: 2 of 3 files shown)


wainwmr (approved):
OK

PR #715, opened 2026-06-24
test: reusable Playwright computed-style assertion helper (#714)

### LINKED ISSUE #714: Reusable Playwright computed-style assertion helper (general + assertReadableInMode contrast)
### 2. Description

Provide a reusable Playwright computed-style assertion helper in `packages/pwa/e2e/` so the real-browser computed-style test mandated by the #1074 rule is cheap to write and not re-derived per ticket. This folds Workbench Item #523, which will be closed as superseded.

### 3. Reachability

Not applicable: this is a `tech debt` ticket.

### 4. Current state

- `packages/pwa/e2e/lightmode-contrast.spec.ts` already reads `window.getComputedStyle(el).color` inline (lines 10-20 and 59-61) to verify contrast in light and dark modes.
- `doc/TESTING.md` lists "CSS-triggered behaviour (no layout engine)" under "What jsdom cannot reliably test" (lines 89-96), but does not call out computed properties, the cascade, specificity, `!important`, or inheritance as a distinct class.
- `tests-local/method-sub-vera-gate.md` encodes the rule for gate reviewers (step 4, "test of record"): when a fix turns on a computed CSS property or the cascade, the proof must be a real-browser read.
- No shared helper exists; each ticket is expected to write its own `getComputedStyle` / `elementFromPoint` / `toHaveCSS` assertion.

### 5. Root cause

jsdom stores inline style strings and does not compute the cascade, so `style.*` assertions can pass while the rendered result is wrong. The only reliable proof is a real-browser read. Without a helper, every ticket that needs this proof re-derives the same Playwright boilerplate.

### 6. Scope

**In scope:**

1. A general computed-style assertion helper in `packages/pwa/e2e/helpers/computed-style.ts`:
   - `expectComputedStyle(page, selector, property, expected)` — resolves the first matching element via `page.locator(selector).evaluate(el => getComputedStyle(el)[property])` and asserts equality.
   - `expectElementFromPointStyle(page, x, y, property, expected)` — resolves `document.elementFromPoint(x, y)` and asserts the computed property.
   - Optional negation variant `expectComputedStyleNot` for "does not equal" assertions.
2. The contrast specialisation `assertReadableInMode(page, selector, mode)` from #523:
   - Toggle the requested theme (`dark` or `light`).
   - Read the foreground `color` and effective background `background-color` of the element.
   - Assert the two values differ (minimum contrast ratio check is optional but preferred).
3. Refactor `packages/pwa/e2e/lightmode-contrast.spec.ts` to import and exercise the helper, proving it works in the existing contrast tests.
4. Update `doc/TESTING.md` "What jsdom cannot reliably test" to add the computed-CSS/cascade class and reference the helper as the test-of-record pattern.

**Deferred:**

- Refactoring `e2e/splitter-cursor.spec.ts` — the file does not exist; revisit when that ticket is worked.
- Turning the helper into a shared npm package or Vitest plugin — out of scope for this ticket.

### 7. Recommended fix

1. Create `packages/pwa/e2e/helpers/computed-style.ts`:

   ```typescript
   import { expect, type Page } from "@playwright/test";

   export async function expectComputedStyle(
     page: Page,
     selector: string,
     property: string,
     expected: string
   ) {
     const actual = await page.locator(selector).evaluate(
       (el, prop) => window.getComputedStyle(el)[prop],
       property
     );
     expect(actual).toBe(expected);
   }

   export async function expectElementFromPointStyle(
     page: Page,
     x: number,
     y: number,
     property: string,
     expected: string
   ) {
     const actual = await page.evaluate(
       ({ x, y, prop }) => {
         const el = document.elementFromPoint(x, y);
         if (!el) throw new Error(`no element at (${x}, ${y})`);
         return window.getComputedStyle(el)[prop];
       },
       { x, y, property }
     );
     expect(actual).toBe(expected);
   }
   ```

2. Add `assertReadableInMode(page, selector, mode)` on top of the general helper:

   - Click `#darkswitch` if `mode === "light"` and the body does not already have `light-theme`.
   - Read `color` and `background-color` from the computed style of the selected element.
   - Assert `color !== background-color`.

3. Refactor `packages/pwa/e2e/lightmode-contrast.spec.ts`:

   - Replace inline `window.getComputedStyle(el).color` calls with `expectComputedStyle`.
   - Replace the mode-toggling and colour-equality check in the second test with `assertReadableInMode` where appropriate, or keep the existing structure if it proves the helper.
   - Keep the test fixtures that inject `#test-help-link` and reveal `#feedback-modal`.

4. Update `doc/TESTING.md` lines 89-96:

   - Add a bullet: "computed CSS properties and cascade effects (specificity, inheritance, `!important`) — use Playwright `getComputedStyle` / `elementFromPoint` / `toHaveCSS`".
   - Reference `packages/pwa/e2e/helpers/computed-style.ts` as the reusable helper.

### 8. Test plan

1. **Helper unit behaviour via e2e spec.** Add or extend a focused Playwright spec (`e2e/computed-style.spec.ts`) that:
   - navigates to `/`, injects an element with a known inline style, and asserts `expectComputedStyle` returns the resolved value;
   - asserts `expectElementFromPointStyle` returns the resolved value for the element under a given point;
   - asserts `assertReadableInMode` passes for an element whose foreground and background differ in both modes.
2. **Existing coverage preservation.** Run `pnpm run e2e` after refactoring `lightmode-contrast.spec.ts`; the two existing contrast tests must still pass.
3. **Type check.** Run `pnpm run check:types:e2e`; the helper and refactored spec must type-check.
4. **Repo-doc check.** Run `pnpm run check` or at least `pnpm --filter @spem/pwa check:types` and `npx markdownlint-cli2 doc/TESTING.md` after the doc edit.

### 9. Dependencies

- Supersedes Workbench Item #523. Close #523 when this ticket is created/moved to In Progress.
- None blocking.

### 10. Open questions

None.

### 11. Follow-up items

- Close Workbench Item #523 as superseded by Ticket #714.

### 12. Documentation impact

- `doc/TESTING.md` — update the jsdom-limitations list to name computed CSS / cascade as a distinct class and reference the helper.

wainwright1000:
## What

A reusable Playwright computed-style assertion helper in `packages/pwa/e2e/helpers/computed-style.ts`, so the real-browser computed-style proof the #1074 rule mandates is cheap to write and not re-derived per ticket. jsdom stores inline style strings but does not compute the cascade, so `style.*` assertions can pass while the rendered result is wrong; the only reliable proof is a real-browser read.

- `getComputedStyleValue` — read primitive; all assertions build on it.
- `expectComputedStyle` / `expectComputedStyleNot` — equality / negation.
- `expectElementFromPointStyle` — reads the element under a viewport point.
- `assertReadableInMode` — toggles the theme and asserts the element's `color` and (opaque) `background-color` differ.

All reads use `getComputedStyle().getPropertyValue` (kebab-case names resolve) and reject an empty (`""`) result — a misspelled/wrong-case/unset property — unless `{ allowEmpty: true }`, so a typo cannot make an assertion pass vacuously.

Refactors `lightmode-contrast.spec.ts` onto the helper; adds `computed-style.spec.ts` exercising every function including negative cases that prove the assertions can fail; documents the computed-CSS/cascade class in `doc/TESTING.md`.

Fixes #714

## Tests

`pnpm run e2e` (the new spec + the refactored contrast spec). Type-checked via `check:types:e2e`. No production code changes.

Folds and supersedes Workbench Item #523.

- Claude

### DIFF
--- a/doc/TESTING.md
+++ b/doc/TESTING.md
@@ -92,9 +92,10 @@ Choosing the right layer keeps the suite fast and deterministic.
 - `AudioContext` (not implemented)
 - canvas pixel output (the `canvas` package provides the API but not pixel-perfect browser rendering)
 - CSS-triggered behaviour (no layout engine)
+- computed CSS properties and cascade effects (specificity, inheritance, `!important`) — jsdom stores inline style strings but does not compute the cascade, so `style.*` can pass while the rendered result is wrong; use Playwright `getComputedStyle` / `elementFromPoint` / `toHaveCSS`
 - real timer execution inside DOM event handlers (`vi.useFakeTimers()` does not intercept `setTimeout` scheduled by jsdom's internal event loop)
 
-If a test needs any of the above, it belongs in the E2E layer.
+If a test needs any of the above, it belongs in the E2E layer. For the computed-CSS / cascade case, the reusable assertions in `packages/pwa/e2e/helpers/computed-style.ts` (`expectComputedStyle`, `expectElementFromPointStyle`, `assertReadableInMode`) are the test-of-record pattern — import them rather than re-deriving the `getComputedStyle` boilerplate.
 
 ## CI Behaviour
 
--- a/packages/pwa/e2e/computed-style.spec.ts
+++ b/packages/pwa/e2e/computed-style.spec.ts
@@ -1,0 +1,180 @@
+import { test, expect, type Page } from "@playwright/test";
+import {
+  getComputedStyleValue,
+  expectComputedStyle,
+  expectComputedStyleNot,
+  expectElementFromPointStyle,
+  assertReadableInMode,
+} from "./helpers/computed-style";
+
+// Exercises the reusable computed-style assertion helper (#714). The helper is
+// the test-of-record pattern for the #1074 rule: when a fix turns on a computed
+// CSS property or the cascade, the proof must be a real-browser read, because
+// jsdom does not compute the cascade.
+
+test.describe("computed-style e2e helper", () => {
+  test.beforeEach(async ({ page }: { page: Page }) => {
+    await page.goto("/");
+  });
+
+  test("getComputedStyleValue / expectComputedStyle resolve a computed property", async ({
+    page,
+  }) => {
+    await page.evaluate(() => {
+      const el = document.createElement("div");
+      el.id = "cs-fixture";
+      el.style.color = "rgb(1, 2, 3)";
+      document.body.appendChild(el);
+    });
+
+    expect(await getComputedStyleValue(page, "#cs-fixture", "color")).toBe(
+      "rgb(1, 2, 3)"
+    );
+    await expectComputedStyle(page, "#cs-fixture", "color", "rgb(1, 2, 3)");
+    await expectComputedStyleNot(
+      page,
+      "#cs-fixture",
+      "color",
+      "rgb(255, 255, 255)"
+    );
+  });
+
+  test("expectComputedStyle reads a kebab-case property", async ({ page }) => {
+    await page.evaluate(() => {
+      const el = document.createElement("div");
+      el.id = "bg-fixture";
+      el.style.backgroundColor = "rgb(4, 5, 6)";
+      document.body.appendChild(el);
+    });
+
+    // The bracket-index form would return undefined here; getPropertyValue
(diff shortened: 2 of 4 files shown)


wainwmr (approved):
OK

Issue #646, opened 2026-06-17
Type fireEvent's CustomEvent detail across the event system


wainwright1000:
## 2. Description

`packages/pwa/src/ts/MusicElement.ts` dispatches a plain, un-generic `CustomEvent` for every custom event used by the four music elements. Because the event detail is untyped, every consumer's `e.detail` is effectively `any`. A drift in the payload shape (for example renaming `Position.part`, changing `PartType`, or passing the wrong detail object) produces no compile-time error and only surfaces at runtime, if at all. This is the same silent shape-drift class that produced Bug #568.

## 3. Reachability

N/A. This is tech debt, not a user-facing bug.

## 4. Current state

Files involved:

- `packages/pwa/src/ts/MusicElement.ts` — defines `fireEvent(type: string, pos?: Position)` and dispatches `new CustomEvent(type, { detail: { position: position } })`.
- `packages/pwa/src/ts/common.ts` — defines the `Position` interface used in the detail.
- `packages/pwa/src/ts/MusicControls.ts`, `MusicCanvas.ts`, `MusicScore.ts` — subclass emitters that call `fireEvent` with 11 distinct event type strings.
- `packages/pwa/index.ts` — adds listeners for `music-score-click`, `music-controls-changed`, `music-controls-playing`, `music-controls-paused`, `music-canvas-click`, `music-canvas-touchstart`, and `music-canvas-touchend`, casting handlers with `as (e: Event) => void`.
- `packages/pwa/src/ts/MusicCanvasWatcher.ts` — listens for `music-canvas-hover` and reads `e.detail.position`, also cast with `as (e: Event) => void`.
- Unit tests in `canvas.test.ts`, `controls.test.ts`, `score.test.ts`, `canvaswatcher.test.ts`, `setbar.test.ts` construct or listen for these events.

All current event payloads use the same shape: `{ position: Position }`.

## 5. Root cause

`CustomEvent` is used without a generic type argument, so TypeScript infers `detail` as `any`. The event names are also passed as unconstrained `string`, so a misspelled event type at a call site is not caught either. There is no `HTMLElementEventMap` augmentation, so `addEventListener` cannot narrow the event object for the music-specific types and consumers must cast.

## 6. Scope

**In scope:**

- Introduce a shared `MusicEventDetail` type and a `MusicEventType` union of all event names in `packages/pwa/src/ts/common.ts`.
[...]

PR #718, opened 2026-06-24
refactor: add compile-time exhaustiveness guard for music event typing

### LINKED ISSUE #716: Add a compile-time exhaustiveness check to MusicElement's event typing
**Type:** tech debt | **Area:** Other | **Status:** Todo | **Difficulty:** S | **Concern:** `~maint`

## 2. Description

`MusicEventType` (the union of the 12 `music-*` custom-event names) and the
`HTMLElementEventMap` augmentation that narrows `addEventListener`/`e.detail` for
those events are two hand-maintained lists of the same keys, kept in sync by a
comment with no compile-time guard. A new event name added to `MusicEventType`
**without** a matching `HTMLElementEventMap` key compiles cleanly on the
`fireEvent` side and only fails to narrow at the consumer's `addEventListener`,
where the listener silently falls back to the generic `Event` and `e.detail`
loses its `MusicEventDetail` type. Nothing fails the build.

This is the finding already recorded as item 20 in `wiki/refactor-common.ts.md`
(RR-only tech debt, classified Q7 — a compiler-enforced invariant). It was
surfaced 2026-06-23 by the `musicelement.test.ts` bug-hunt and empirically
verified: removing the `music-controls-changed` map entry produced `TS2339` at
the test listener and at `index.ts:567`. This ticket promotes that RR item to a
board ticket and closes the gap with a guard.

## 4. Current state

- `packages/pwa/src/ts/common.ts:18-30` — the `MusicEventType` union (12 members).
- `packages/pwa/src/ts/common.ts:54-67` — the `HTMLElementEventMap` augmentation
  (12 keys, all `CustomEvent<MusicEventDetail>`).
- `packages/pwa/src/ts/common.ts:50-53` — the comment instructing maintainers to
  "Keep these keys in lockstep", noting the DRY `extends Record<MusicEventType,
  ...>` form is blocked by eslint `no-empty-object-type`.
- `fireEvent`'s parameter is typed `MusicEventType`
  (`packages/pwa/src/ts/MusicElement.ts:94`), so the reverse drift (a map key not
  in the union) already fails to compile if that name is ever fired; only the
  union-without-map direction is silent.

## 5. Root cause

The DRY single-source form was deliberately rejected over an eslint rule, leaving
two explicit lists with no type-level cross-check. There is no assertion that the
union is a subset of the map's keys, so drift in the dangerous direction is
invisible to `tsc`.

## 6. Scope

**In scope:** one zero-runtime, type-level assertion in `common.ts` next to the
two lists, plus a committed type-level test pinning that the guard rejects a
missing key.

**Out of scope (and why only one direction is checkable):** `HTMLElementEventMap`
is the *global* DOM event map, so `keyof HTMLElementEventMap` already contains
every standard DOM event name. Only "every `MusicEventType` is a key of the map"
is checkable from the types — which is exactly the silent-failure direction. The
reverse (a map key with no union member) cannot be isolated at type level and is
already covered by `fireEvent`'s `MusicEventType` parameter. Refactor item 26
(per-event `detail` typing) is a distinct redesign blocked by the same eslint
rule; it is **deferred**, not folded (see Follow-up items).

## 7. Recommended fix

Add, immediately after the `HTMLElementEventMap` augmentation in `common.ts`:

```ts
/**
 * Compile-time exhaustiveness guard (#716): every `MusicEventType` member must
 * have a key in the `HTMLElementEventMap` augmentation above, so a new event
 * name cannot silently untype its listeners. If a member is added without a map
 * entry, `Exclude<...>` is that member (a string literal, not `never`) and
 * `AssertNever` fails its `never` constraint, so `tsc` (check:types) errors here.
 * Type-only: emits no runtime code.
 */
type AssertNever<_T extends never> = void;
type _MusicEventMapIsExhaustive = AssertNever<
  Exclude<MusicEventType, keyof HTMLElementEventMap>
>;
```

When the lists are in sync, `Exclude<...>` is `never`, `AssertNever<never>` is
valid, and `tsc` is silent. When a union member lacks a map key, that member is
not assignable to `never`, so the constraint fails (`TS2344`) at the definition
site.

Notes for the implementer (verify at `work` time, adjust if a check objects):

- This form is preferred over the refactor report's `const _c: ... = true;`
  suggestion because the `const` emits runtime code and trips `noUnusedLocals`;
  the type-only form does neither.
- Confirm `check:lint` (`@typescript-eslint/no-unused-vars`) and `check:unused`
  (knip) do not flag the unexported type aliases. Neither should: knip reports
  unused *exports*, and no-unused-vars does not report unused type aliases. If
  either objects, fall back to the committed `@ts-expect-error` probe form as the
  guard's home.

## 8. Test plan

TDD, red first (the failing check is type-level):

1. **Red.** Temporarily add a canary member to `MusicEventType`, e.g.
   `| "music-canary-unmapped"`, with **no** matching `HTMLElementEventMap` key.
   Run `pnpm --filter @spem/pwa check:types`. It must fail at
   `_MusicEventMapIsExhaustive` / `AssertNever` (`TS2344`, the canary string does
   not satisfy `never`). This proves the guard tests what we think.
2. **Green.** Remove the canary. `check:types` passes.
3. **Durable committed test.** In `packages/pwa/src/test/musicelement.test.ts`
   (beside the #646 listener-typing test), add a self-contained type probe using
   `@ts-expect-error`, mirroring the existing #646 and #551 (`spem.test.ts`)
   compile-time-guard pattern. It exercises the `AssertNever`/`Exclude` mechanism
   against a *local* union + *local* map with a deliberately missing key, so it
   does not depend on the real lists drifting. If someone weakens the mechanism so
   a missing key stops erroring, the `@ts-expect-error` becomes unused and
   `check:types` fails. The test files are in the `tsconfig` `include`, so
   `check:types` enforces this in CI.
4. The existing #646 listener-typing test continues to pass unchanged.

No runtime test is added: the change is type-only and emits nothing to run.

## 9. Dependencies

None. Builds on #646 / PR #673 (the event typing), both merged.

## 10. Open questions

None.

## 11. Follow-up items

- **Refactor item 26** (`refactor-common.ts.md`) — the augmentation gives all 12
  events one uniform `CustomEvent<MusicEventDetail>` detail type rather than a
  per-event keyed detail. **Disposition: deferred (not folded).** It is a distinct
  redesign (a per-event keyed map or generic `fireEvent<K>`), blocked by the same
  eslint `no-empty-object-type` rule, and folding it would turn an additive guard
  into a map redesign. It stays in the refactor report; it can earn its own
  ticket via `isitabug` if it ever merits one.

## 12. Documentation impact

None expected. `common.ts` maps to no `doc/*.md` that documents the event union;
the change is internal and type-level.

- Vera

wainwright1000:
Adds a zero-runtime, compile-time exhaustiveness guard so a `MusicEventType`
union member added without a matching `HTMLElementEventMap` key fails `tsc`,
instead of silently untyping its listeners (the listener falls back to the
generic `Event` and loses `e.detail`'s `MusicEventDetail` type, with no build
error).

Changes:

- `packages/pwa/src/ts/common.ts`: `AssertNever<_T extends never>` plus
  `MusicEventMapIsExhaustive = AssertNever<Exclude<MusicEventType, keyof HTMLElementEventMap>>`.
  Both are type-only (no runtime output). When a union member lacks a map key,
  `Exclude<...>` is that member (a string literal, not `never`) and the
  `extends never` constraint fails at `tsc` (check:types), at the definition site.
- `packages/pwa/src/test/musicelement.test.ts`: a self-contained `@ts-expect-error`
  type probe pinning the guard mechanism, matching the #646 fireEvent test and the
  #551 readonly-Range test.

Only the union-member-without-map-key direction is type-checkable: `HTMLElementEventMap`
is the global DOM event map, so the reverse (an orphan map key) can't be isolated
as an assertable-empty set; it is also low-harm, since `fireEvent`'s `MusicEventType`
parameter rejects dispatching any name outside the union. This is the silent-failure
direction the guard closes.

Promotes refactor-report item 20 (previously RR-only). Verified by a throwaway
canary (an unmapped union member produced `TS2344`, then reverted) and the
committed probe. Vera gate passed; `pnpm run ci` green.

Fixes #716

- Vera

### DIFF
--- a/packages/pwa/src/test/musicelement.test.ts
+++ b/packages/pwa/src/test/musicelement.test.ts
@@ -1,6 +1,21 @@
 import { MusicElement } from "../ts/MusicElement";
 import { MusicControls } from "../ts/MusicControls";
-import type { MusicEventDetail } from "../ts/common";
+import type { AssertNever, MusicEventDetail } from "../ts/common";
+
+// Compile-time exhaustiveness guard for the music-event typing (#716).
+// `AssertNever<T>` (common.ts) errors unless `T` is `never`. Applied here to a
+// local union and a local event map with a deliberately MISSING key, the
+// `Exclude` resolves to that missing key, the assertion errors, and the
+// suppression directive below absorbs it. If the guard mechanism ever stops
+// rejecting a missing key, that directive becomes unused and `tsc`
+// (check:types) fails -- the same compile-time-guard pattern as the #646
+// fireEvent test below and the #551 readonly-Range test in spem.test.ts. This
+// pins the mechanism that protects the real `MusicEventType` /
+// `HTMLElementEventMap` lockstep in common.ts.
+type _ProbeMap = { "music-mapped": CustomEvent<MusicEventDetail> };
+type _ProbeUnion = "music-mapped" | "music-unmapped";
+// @ts-expect-error -- "music-unmapped" has no _ProbeMap key, so Exclude is "music-unmapped", which violates AssertNever's `never` constraint (#716)
+type _ProbeExhaustive = AssertNever<Exclude<_ProbeUnion, keyof _ProbeMap>>;
 
 // Create a concrete subclass for testing
 class TestElement extends MusicElement {
--- a/packages/pwa/src/ts/common.ts
+++ b/packages/pwa/src/ts/common.ts
@@ -67,6 +67,34 @@ declare global {
   }
 }
 
+/**
+ * Compile-time assertion that `T` is `never`. A non-`never` `T` fails the
+ * `extends never` constraint, so `tsc` errors at the instantiation site. It is a
+ * `type` alias (resolving to `void`), so it emits no runtime code. Shared so
+ * type-level guards (and their tests) can express "this set must be empty".
+ */
+export type AssertNever<_T extends never> = void;
+
+/**
+ * Compile-time exhaustiveness guard (#716): every `MusicEventType` member must
+ * have a matching key in the `HTMLElementEventMap` augmentation above, so a new
+ * event name cannot silently untype its listeners. If a member is added to the
+ * union without a map entry, `Exclude<...>` is that member (a string literal, not
+ * `never`) and `AssertNever` fails its `never` constraint, so `tsc`
+ * (check:types) errors here at the definition site. Type-only: no runtime output.
+ * The reverse direction (a map key with no union member) is not checkable here:
+ * `HTMLElementEventMap` is the global DOM event map, so `keyof` it spans every
+ * native event and the reverse `Exclude` is not an assertable-empty set. It is
+ * also low-harm, because `fireEvent`'s `MusicEventType` parameter rejects
+ * dispatching any name outside the union, so an orphan map key can never be fired.
+ *
+ * Exported (not a local alias) only so `noUnusedLocals` does not flag it; it is a
+ * guard, not a value to consume.
+ */
+export type MusicEventMapIsExhaustive = AssertNever<
+  Exclude<MusicEventType, keyof HTMLElementEventMap>
+>;
+
 export type Status = "playing" | "paused" | "loading";
 export type RecordingIndex = 0 | 1;
 


wainwright1000:
Cancelling this PR and returning #716 to In Progress.

#716 is a low-priority maintainability guard, and build usage is currently throttled, so it should wait in the parked queue rather than hold a Review slot and fire CI / Netlify previews on every push. It was published earlier via a `high` label used purely to bypass the throttle, which misrepresented its priority; that label has been removed.

The branch `andrew/716-event-map-exhaustiveness-guard` is preserved locally with the work intact (Vera gate already passed). It will re-park on the next `publish` and open a fresh PR automatically when the throttle clears.

- Vera

PR #720, opened 2026-06-27
refactor: deduplicate MusicCanvas #init listener registration

### LINKED ISSUE #651: Deduplicate MusicCanvas #init listener registration (test-first)
**Type:** tech debt — **Area:** Canvas — **Difficulty:** S
**Source:** `packages/pwa/src/ts/MusicCanvas.ts` — `#init()` listener registration, duplicated between the early-return (reconnect) branch (lines 182-200) and the create branch (lines 206-219).

## Description

`#init()` registers the same six event listeners in two places. The early-return branch (taken on reconnect, when `this.canvas != null`) and the create branch (first connect) each call the identical six `addEventListener` lines — click, mousemove, touchstart, touchmove, touchend, wheel — with identical handler references (the `#bound*` fields) and identical options. ~14 duplicated lines. The early-return path is load-bearing: it re-attaches listeners on reconnect **without** re-running the first-init work (canvas creation, `processLilypond`, pulse/phase array init, initial `draw()` and shimmer-loop start).

## #484 qualifier — Simplify (Q6)

A countable reduction: the six listener registrations collapse from two sites to one. Real-but-not-broken tech debt, ticketed because the duplication is a maintenance hazard — a future listener change (a new event, an options tweak, mirroring the disconnect-side teardown) must be made in two places or silently diverge.

## Root cause

`#init()` grew two entry paths (first connect vs reconnect) and the listener registration was copied into both rather than factored out. Confirmed against `origin/main`: the two blocks are **byte-identical** in handlers and options (`mousemove` uses capture `false`; touch/wheel use `{ passive: false }`; `click` uses defaults), so the extraction is behaviour-preserving.

## Recommended fix

Factor the six registrations into one private helper and call it from both paths; guard the first-init-only work behind a flag.

```typescript
async #init() {
  const firstInit = this.canvas == null;
  if (firstInit) {
    this.canvas = document.createElement("canvas");
    this.append(this.canvas);
  }
  this.#attachListeners();
  if (firstInit) {
    this.#calculateCanvasSize();
    this.#showLoadingOnCanvas();
    // processLilypond, pulse/phase array init, draw(), #startShimmerLoop() — unchanged
  }
}

#attachListeners() {
  this.canvas!.addEventListener("click", this.#boundCanvasClicked);
  this.canvas!.addEventListener("mousemove", this.#boundCanvasHovered, false);
  this.canvas!.addEventListener("touchstart", this.#boundTouchStarted, { passive: false });
  this.addEventListener("touchmove", this.#boundTouchMoved, { passive: false });
  this.addEventListener("touchend", this.#boundTouchEnded, { passive: false });
  this.addEventListener("wheel", this.#preventVerticalScroll, { passive: false });
}
```

External behaviour stays identical: reconnect re-attaches all six listeners and reloads no data; first connect does the full setup once. `disconnectedCallback` does not null `this.canvas`, so the reconnect path still finds `firstInit === false`. Do not fold in unrelated MusicCanvas refactor items.

## Test plan (test-first — load-bearing)

The early-return branch is the reconnect path, so the merge MUST be guarded by tests, red-green-refactor.

1. **(Existing — must stay green) Reconnect re-registers all six listeners.** `canvas.test.ts` "removes all event listeners on disconnect and re-adds them on reconnect (#203)" already asserts this; it is the listener half of the guard and must not break. (The ticket's original "currently untested" note is **stale**: listener re-registration is covered. The untested invariant is the once-only first-init work below.)

2. **(New — write first, red) Reconnect does not re-run first-init work.** On disconnect→reconnect, assert the observable once-only invariants a botched merge would break:
   - the `<canvas>` child element is the **same node** after reconnect (the create branch did not run again — `expect(reconnectedCanvas).toBe(originalCanvas)`);
   - `shimmerPhases` is **unchanged** across reconnect (the random phase init at first connect was not repeated).

   Assert observable state, not internals — do not spy on the private helper. Written against current code these pass (the early-return already preserves them); they then pin the contract the extraction must keep.

3. Run the full `canvas.test.ts` suite; the existing reconnect tests (#203, #245, #554) stay green. No browser-specific behaviour changes, so no new Playwright case is required — jsdom covers listener re-registration (as the #203 test does).

## Dependencies

- Distinct from #203 (disconnect-side teardown, fixed) and #245 (duplicate animation loops, fixed).
- Three open `MusicCanvas.ts` refactors from the same 2026-06-17 hunt touch **disjoint regions**: #649 (`#startShimmerLoop`, parked), #650 (draw/projection helpers, lines ~481-715), and this #651 (`#init`, lines ~180-250). No textual conflict between them; whichever merges first, the others rebase cleanly. No bundling.

## Provenance

Recovered tech-debt finding from the #593 recovery sweep, `wiki/refactor-MusicCanvas.ts.md` bh3 lens pass (item 3). Filed by Vera.

- Vera

wainwright1000:
Deduplicates the `MusicCanvas` `#init` listener registration (test-first).

## What changed

- Extracted a shared helper that registers the canvas/element listeners, used by both `#init` paths (first init and re-init), removing the duplicated listener-registration logic.
- Test-first: canvas tests cover the shared registration so the two paths cannot silently drift.

## Testing

- `pnpm run test:unit` (canvas.test.ts); full CI on the PR.

Fixes #651

- Tele

### DIFF
--- a/packages/pwa/src/test/canvas.test.ts
+++ b/packages/pwa/src/test/canvas.test.ts
@@ -1410,6 +1410,40 @@ describe("MusicCanvas custom element", () => {
     freshCanvas.remove();
   });
 
+  it("reconnect does not re-run first-init work (#651)", async () => {
+    const freshCanvas = document.createElement("music-canvas") as MusicCanvas;
+    document.body.appendChild(freshCanvas);
+    await Promise.resolve();
+
+    const originalInnerCanvas = freshCanvas.querySelector("canvas")!;
+    const originalShimmerPhases = freshCanvas.shimmerPhases;
+    const originalFalseRelationPulses = freshCanvas.falseRelationPulses;
+    // Guard against a vacuous identity probe: shimmerPhases must be a populated
+    // array so a re-run would produce a different one.
+    expect(originalShimmerPhases.length).toBeGreaterThan(0);
+
+    freshCanvas.remove();
+    // The <canvas> survives disconnect (disconnectedCallback removes listeners
+    // and cancels loops but does not null this.canvas); this is the invariant
+    // the firstInit sentinel relies on to take the reconnect path.
+    expect(freshCanvas.canvas).toBe(originalInnerCanvas);
+
+    document.body.appendChild(freshCanvas);
+    await Promise.resolve();
+
+    // The reconnect (early-return) path re-attaches listeners but must NOT
+    // re-run first-init work: the <canvas> is not recreated, and the freshly
+    // minted first-init arrays (random shimmerPhases via map, falseRelationPulses
+    // via new Array) keep their identity. These two are the right re-run probes:
+    // they are rebuilt on every #init, unlike frLocations/notesByQuant/ranges,
+    // which processLilypond memoises and would return by the same reference.
+    expect(freshCanvas.querySelector("canvas")).toBe(originalInnerCanvas);
+    expect(freshCanvas.shimmerPhases).toBe(originalShimmerPhases);
+    expect(freshCanvas.falseRelationPulses).toBe(originalFalseRelationPulses);
+
+    freshCanvas.remove();
+  });
+
   it("prevents default on vertical wheel events but not on horizontal/zero scroll (#676)", () => {
     expect(canvas).not.toBeNull();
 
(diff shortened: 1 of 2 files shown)


wainwmr (approved):
Nice

wainwmr (approved):
Ok

PR #721, opened 2026-06-27
refactor: harden colors() against partial CSS load and mutation

### LINKED ISSUE #653: Harden colors(): complete defaults, finite-hue guard, immutable return, typed choir
**Type:** tech debt — **Area:** UI
**Source:** `packages/pwa/src/ts/common.ts` (`colors()` ~45-105, `Colors.choir` ~32)

## Summary

Harden `colors()` and the `Colors` type against partial CSS load, shared-cache mutation, and incomplete defaults — one coherent unit:

- **(11) NaN hues on partial CSS load → transparent render.** `Number(getPropertyValue("--color-cN"))` returns `NaN` when a hue custom property is absent/non-numeric; `NaN` flows into `hsla(NaN, 80%, 55%, 1)` at `MusicCanvas.ts:435,473,513` and `MusicScore.ts:229`, which browsers render as **transparent, no console error**. The CSS-present gate only checks `--color-background`, so a partial-load state (background present, `--color-c5` empty) passes it. Fix: validate each hue with `Number.isFinite`, substituting `config.choirHues[i]`, or `logError` + return `defaultColors`.
- **(6) Incomplete default choir array.** `defaultColors.choir` has one hue (`[360]`) but the app expects 8; in the fallback path `colors().choir[c]` is `undefined` for `c > 0`. Fix: complete the defaults to 8 hues.
- **(19) Mutable cached object returned.** `colors()` returns the module-level `loadedColors` directly on cached calls (~79); a caller mutating it (`colors().choir.push(...)`) corrupts the cache for everyone. The fallback branch already copies; the cached branch does not. Fix: return a shallow copy or `Object.freeze` the cache.
- **(13) Type-level invariant — the compile-time half of 6/11.** Tighten `Colors.choir: number[]` → `readonly number[]`, or better a length-8 tuple, so TypeScript checks the 8-element literal in the CSS-present branch and forbids downstream mutation.
- **(5, rides along)** Remove the trailing semicolons in `defaultColors` HSL strings (~45-50) — invalid as colour values, inconsistent with `getPropertyValue` output.

## #484 qualifier — Maintainable (Q7) + Simplify (Q6)

Q7: removes a specific way the code renders wrongly (transparent choirs on partial load), enforces an immutable-data invariant (mutable cache), and adds a compiler-enforced 8-element invariant (13). Q6: the tuple type collapses the runtime 8-element check into the type system.

## Reachability note (flag for specify)

Item 11's wrong result (transparent choir colours) is reachable at **early page load / partial stylesheet load** — an uncommon-but-real state. Filed as tech debt per the report's BT-v2 classification and the #593-recovery "no re-bucket" instruction, but it has a user-facing edge; specification may judge it Bug-shaped.

## Test plan (folds in item 12)

Add a test for the realistic partial-load case (`--color-background` present but `--color-cN` missing) exercising the finite-hue guard; optionally a Playwright pre-stylesheet test.

## Notes

Recovered tech-debt finding; not yet specified. Distinct from closed #101 (dynamic hue loading — a different fix shape). Item 16 (narrow `v` to `0|1`) is already tracked by closed #369; item 17 (TempoMap struct) is marginal post-#369 and left RR-only.

Provenance: #593 recovery sweep, from `refactor-common.ts.md` (colors() robustness items 6/11/19, type item 13). Filed by Vera.

wainwright1000:
Hardens `colors()` in `common.ts` against partial CSS load and mutation.

## What changed

- Complete defaults, so a partial CSS load can't yield undefined colour values.
- Finite-hue guard (rejects a non-finite hue).
- Returns an immutable copy, so callers can't mutate shared colour state.
- Typed `choir` parameter.

## Testing

- `pnpm run test:unit` (common.test.ts extended); full CI on the PR.

Fixes #653

- Tele

### DIFF
--- a/packages/pwa/src/test/common.test.ts
+++ b/packages/pwa/src/test/common.test.ts
@@ -321,4 +321,67 @@ describe("common", () => {
     const b = colors(true);
     expect(a.choir).not.toBe(b.choir);
   });
+
+  it("colors() substitutes config.choirHues for a non-finite or empty CSS hue (#653 item 11)", () => {
+    // Partial stylesheet load: --color-background present (so the CSS-present
+    // branch is taken) but one choir hue blank. Number("") is 0 (finite), so
+    // the guard must reject empty/blank values, not only non-finite ones;
+    // otherwise the missing hue renders as hue 0 (red) instead of transparent
+    // or its intended colour.
+    document.body.style.setProperty("--color-background", "#123456");
+    document.body.style.setProperty("--color-highlight", "#abcdef");
+    document.body.style.setProperty("--color-score-highlight", "#ffffff");
+    for (let i = 1; i <= 8; i++) {
+      document.body.style.setProperty("--color-c" + i, String(i * 10));
+    }
+    document.body.style.removeProperty("--color-c5"); // blank the 5th hue
+    document.body.style.setProperty("--color-c3", "not-a-hue"); // non-numeric
+    const result = colors(true);
+    // The blank hue (empty → Number("") === 0) and the non-numeric hue
+    // (→ NaN) both fall back to config; the rest still read from CSS.
+    expect(result.choir[4]).toBe(config.choirHues[4]); // empty → fallback
+    expect(result.choir[2]).toBe(config.choirHues[2]); // non-numeric → fallback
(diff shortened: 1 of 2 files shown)


wainwmr (approved):
Ok

Issue #728, opened 2026-06-28
monitor refresh can silently overwrite the daily mergedPRs under the rebase-retry push race


wainwright1000:
**Type:** tooling bug · **Area:** Tooling · **Difficulty:** S · **Status:** Todo
**Source:** `packages/monitor/monitor-resources.mjs`, `.github/scripts/push-with-rebase.sh`

Surfaced by the Vera gate on #727 (silent-failure-hunter). This specification supersedes the creation body; the mechanism question it left open ("a value-preserving merge for the series") is settled below.

## 2. Description

`monitor-resources.mjs` accumulates a daily build-usage series in `.github/monitor-series.json` by read-modify-write (`loadSeries` → `appendOrReplaceDay(today)` → `saveSeries`), never regenerated. The merge-time `refresh()` (#699) always writes today's entry with `mergedPRs: 0` — by design, it does not fetch the PR count — while the daily `main()` run writes today's entry with the real `mergedPRs`. Since #727, concurrent pushes to `main` converge via `git rebase -X theirs` (last-writer-wins) instead of failing non-fast-forward. So a merge-triggered `refresh()` that finishes its push **second**, against the daily run that pushed first, silently overwrites today's real `mergedPRs` (and `summary.mergedPRs`) with `0`.

The run is green; there is no error. It does not self-heal: `refresh()` always writes `0` and `main()` never backfills a past day, so the count is frozen at `0` for that calendar day (recoverable only by a manual `seedBackfillSeries`). `mergedPRs > 0` feeds the burndown histogram (`main()` lines 949-955), so the loss lands on days with merge activity.

Classification: real, broken, and **not** harm-floor — the lost value is a recoverable internal-chart data point, not user state, and the throttle gate never reads `mergedPRs` (only `summary.overallStatus`). Worth-it confirmed (Andrew, 2026-06-30): fix it. → **Tooling Bug**.

## 3. Trigger (race interleaving)

Both jobs push to `main` via `.github/scripts/push-with-rebase.sh`:

1. The daily `main()` run (`monitor-run.yml`, ~02:29 UTC) writes today's entry with `mergedPRs: N` (N > 0) and pushes first.
2. A merge-triggered `refresh()` (`monitor-refresh.yml`), checked out before that push, writes today's entry with `mergedPRs: 0`, is rejected non-fast-forward, and rebases onto the fetched tip with `-X theirs` — which favours the replayed commit (refresh's `0`). Result on `main`: today `mergedPRs: 0`.

[...]

PR #736, opened 2026-06-30
ci: push regenerated files to main via a rebase-retry helper

### LINKED ISSUE #727: Monitor workflows race when pushing monitor-series.json to main (non-fast-forward failure)
**Type:** tooling bug
**Area:** Tooling (CI)
**Difficulty:** XS
**Priority:** High (Andrew, 2026-06-28 - prioritised ahead of the other parked work).

## 2. Description

Both monitor workflows end by committing `.github/monitor-series.json` to `main` and running a bare `git push`. When two runs overlap, e.g. `monitor-refresh.yml` firing on a merge to `main` while `monitor-run.yml` runs from its cron or a manual `workflow_dispatch`, they race: one pushes first, the other is rejected non-fast-forward and reports a red CI failure. The series data still lands via the winning run; this is a spurious CI failure, not a data-correctness problem.

## 3. Reachability

This is CI-facing, not user-facing. The failure is reached whenever two jobs that commit the series file are running concurrently. The observed instance was a manual `monitor-run.yml` run overlapping with an automatic `monitor-refresh.yml` run shortly after #725 merged; Refresh won and Monitor Run failed at the `git push` step.

## 4. Current state

- `.github/workflows/monitor-refresh.yml` (lines 63-74): the `Commit refreshed series` step configures git, stages `.github/monitor-series.json`, commits with `[skip ci]`, and runs `git push` with no rebase or concurrency guard.
- `.github/workflows/monitor-run.yml` (lines 40-50): the `Commit daily resource series` step does the same, using a different commit message (`chore: update daily resource series [skip ci]`).
- Neither workflow declares a `concurrency:` group, so GitHub Actions will run overlapping jobs in parallel.

## 5. Root cause

The two workflows share a single mutable resource (`main` branch state) but do not serialise access to it. A bare `git push` from a checkout that is behind `origin/main` fails non-fast-forward when another job has pushed in the meantime.

## 6. Scope

**In scope:**
- Add a shared `concurrency:` group to `.github/workflows/monitor-refresh.yml` and `.github/workflows/monitor-run.yml` so jobs that commit the series file cannot run concurrently.
- Keep the existing commit-and-push steps otherwise unchanged.

**Out of scope:**
- Rebase-before-push logic (belt-and-braces alternative in the original report). The concurrency group is sufficient and simpler.
- `actions/checkout` version drift between the two files.
- Other CI workflow changes.

## 7. Recommended fix

Add an identical `concurrency:` block to both workflow files at the top level:

```yaml
concurrency:
  group: monitor-series-push
  cancel-in-progress: false
```

Use the same group name in both files so GitHub Actions treats the jobs as a single serialised queue. Set `cancel-in-progress: false` so an already-running daily monitor is not cancelled by a merge-time refresh; it simply waits its turn.

Place the block near the top of each file, after `name:` and before `on:`.

## 8. Test plan

- Inspect both workflow files and confirm the `concurrency` group name is identical (`monitor-series-push`).
- Confirm the YAML is valid: run `actionlint .github/workflows/monitor-refresh.yml .github/workflows/monitor-run.yml` if available, or check the workflow files in the GitHub UI editor.
- Confirm the `[skip ci]` self-commit behaviour is preserved so the concurrency group does not introduce a loop.

A real push race is not unit-testable in this repo; the verification is static inspection.

## 9. Dependencies

None. Surfaced incidentally during the #724 / #725 monitor throttle merge.

## 10. Open questions

None.

## 11. Follow-up items

None.

## 12. Documentation impact

None expected. `doc/MONITOR.md` describes the monitor workflows but does not need updating for a concurrency guard.

wainwright1000:
Adds a shared push helper so the three workflows that commit regenerated files to `main` (monitor-run, monitor-refresh, scores-ci) retry with a rebase when a concurrent push moves `main` first, instead of failing non-fast-forward.

- `.github/scripts/push-with-rebase.sh`: the shared helper. It tries the push, and on a non-fast-forward it fetches and rebases onto the new tip (resolving same-file conflicts in favour of the freshly regenerated content) and retries, bounded, failing loud if it cannot converge.
- The three push sites call the helper in place of a bare `git push`.
- `.github/scripts/push-with-rebase.test.sh`: a bash regression test covering the convergence cases and the loud-fail exhaustion case.
- `.github/workflows/push-helper-test.yml`: runs the test on any change under `.github/scripts/`.
- `doc/CI.md`: documents the mechanism.

This closes the race that produced spurious non-fast-forward failures on the monitor and scores pushes, one source being our own PR merges moving `main` under an in-flight bot push.

Data semantics: for a same-file conflict the helper keeps the later, freshly regenerated content. The score SVGs self-heal (regenerated each run); `monitor-series.json` is accumulated, so a merge-time refresh racing the daily run can briefly overwrite that day's mergedPRs (low odds, recoverable). The value-preserving merge for the series is tracked separately as #728.

Fixes #727

### DIFF
--- a/.github/scripts/push-with-rebase.sh
+++ b/.github/scripts/push-with-rebase.sh
@@ -1,0 +1,86 @@
+#!/usr/bin/env bash
+# Push the current HEAD commit to main with bounded rebase-retry (#727).
+#
+# A bot commit step regenerates a file, commits it, and pushes to main. A bare
+# `git push` from a checkout that has fallen behind origin/main fails
+# non-fast-forward whenever another push landed in between -- non-fast-forward is
+# a property of the ref, not the file. The symptom is a spurious red CI run; the
+# data still lands via the winning run.
+#
+# This helper makes the push converge: on a rejection it replays our commit onto
+# the freshly-fetched main and retries, bounded.
+#
+# Caller contract: HEAD is the single commit to push (already `git add` +
+# `git commit`ed). Run it in place of the final `git push`:
+#
+#     bash .github/scripts/push-with-rebase.sh
+#
+# Why rebase, not `git reset --soft origin/main`: reset would re-stage our whole
+# tree, which still holds the OLD version of any file another push changed in
+# parallel -- re-committing that would silently REVERT their change. Rebase
+# replays only our commit's delta, so concurrent changes to other files are
+# preserved.
+#
+# Why `-X theirs`: the only files that can conflict are the one(s) our commit
+# rewrote (when a parallel run rewrote them too). `-X theirs` resolves that
+# conflict deterministically in favour of the commit being replayed -- ours -- so
+# the push always converges with no manual step. ("theirs" in a rebase means the
+# commit being replayed, i.e. ours.) Picking ours is last-writer-wins on the
+# file's contents; the consequence differs by caller:
+#   - The score SVGs (scores-ci) are fully derived from source, so a dropped
+#     parallel regeneration is re-created by a later source-driven build: it
+#     self-heals. (The score build is mtime-incremental and the bot's SVG commit
+#     does not re-trigger scores-ci, so the catch-up is the next build that
+#     rebuilds the affected score, not necessarily the immediate next run.)
+#   - The monitor series (.github/monitor-series.json) is accumulated, NOT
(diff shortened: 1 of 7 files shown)


wainwmr (approved):
OK

PR #737, opened 2026-06-30
refactor: strengthen LilypondData typing, collapse MusicCanvas fields

### LINKED ISSUE #652: Strengthen LilypondData typing: outer-dim aliases, single field, deep-readonly ranges
### 2. Description

`packages/pwa/src/ts/lily.ts` returns a `LilypondData` object that is consumed by `MusicCanvas`. The current type does not express the indexing semantics of its two main collections, `MusicCanvas` stores the returned data as four loosely-coupled instance fields, and the `ranges` collection is mutable at the Map and array levels despite the `Range` leaf fields being `readonly`. These gaps make partial or accidental mutation easier than it should be.

### 3. Reachability

N/A — this is a tech-debt / type-safety ticket, not a user-facing bug.

### 4. Current state

`packages/pwa/src/ts/lily.ts`, `LilypondData` type (lines 276-281):

```ts
export type LilypondData = {
  readonly notesByQuant: Map<number, NoteEntry[]>;
  readonly ranges: Map<string, Range[]>;
  readonly barCount: number;
  readonly frLocations: FRlocation[];
};
```

- `notesByQuant` is keyed by quantised bar position and `ranges` is keyed by `"choir-part"`, but neither shape has a named alias.
- `ranges` is `Map<string, Range[]>`; only the `Range` leaf fields are `readonly`.

`packages/pwa/src/ts/MusicCanvas.ts` (lines 24-27 and 224-228):

```ts
notesByQuant: Map<number, NoteEntry[]> = new Map();
ranges: Map<string, Range[]> = new Map();
barCount: number = 0;
frLocations: FRlocation[] = [];
```

These four fields are populated together from a single `processLilypond()` call, but the uninitialised state is implicit (empty collections and zero) and partial population is possible.

### 5. Root cause

The data shape grew incrementally: #105/#106 named the leaf `Dictionary`/`Range` types and #107 returned them from `processLilypond()`, but the dimension-carrying Map types were never named; #551 made `Range` leaves readonly but stopped at the collection level. `MusicCanvas` kept the four fields because that is how the data arrived before #107.

### 6. Scope

**In scope:**

1. Add named aliases for the `LilypondData` collection shapes in `lily.ts`.
2. Make `LilypondData.ranges` deeply readonly (`ReadonlyMap<string, readonly Range[]>`).
3. Collapse `MusicCanvas`'s four instance fields into one `lilyData: LilypondData | null = null`.
4. Update `MusicCanvas` references and `packages/pwa/src/test/canvas.test.ts` accordingly.
5. Update `wiki/Architecture-Lily.md` and `wiki/Architecture-Canvas.md` if they describe the old field names or mutable `ranges` shape.

**Out of scope:**

- Deep readonly for `notesByQuant` (tests deliberately mutate it; see comment in `lily.ts`).
- Removing the module-level `barCount` export in `lily.ts` (tracked separately).
- Any other `lily.ts` hygiene (var, loose equality, magic numbers).
- Behavioural changes to canvas rendering.

### 7. Recommended fix

**`packages/pwa/src/ts/lily.ts`**

Add two type aliases above `LilypondData` and update the type:

```ts
// Quantised bar position -> all notes/rests starting at that position.
export type NotesByQuant = Map<number, NoteEntry[]>;

// "choir-part" key -> singing ranges for that part.
export type SingingRanges = ReadonlyMap<string, readonly Range[]>;

export type LilypondData = {
  readonly notesByQuant: NotesByQuant;
  readonly ranges: SingingRanges;
  readonly barCount: number;
  readonly frLocations: readonly FRlocation[];
};
```

In `processLilypond()` (around line 317-393), build `ranges` as a `ReadonlyMap`. The simplest path is to populate a plain `Map<string, Range[]>` as today, then freeze it before assigning to `lilypondCache`:

```ts
const ranges = new Map<string, Range[]>();
// ... populate as today ...
const frozenRanges: SingingRanges = new Map(
  Array.from(ranges.entries()).map(([k, v]) => [k, Object.freeze([...v])])
);
Object.freeze(frozenRanges);

lilypondCache = {
  notesByQuant,
  ranges: frozenRanges,
  barCount: localBarCount,
  frLocations,
};
```

If `Object.freeze` is undesirable, wrap the map once after population and cast to `SingingRanges`, accepting that runtime mutation remains possible but TypeScript will catch it.

**`packages/pwa/src/ts/MusicCanvas.ts`**

1. Import `LilypondData` from `./lily`.
2. Replace the four fields with one:

```ts
lilyData: LilypondData | null = null;
```

3. In `#init()` (lines 224-228), assign the whole object:

```ts
this.lilyData = processLilypond();
```

4. Update every reference from `this.notesByQuant`, `this.ranges`, `this.barCount`, `this.frLocations` to `this.lilyData!.field`. There are 24 occurrences; no method referencing these fields runs before `#init()` completes.

5. The `draw()` early-return guard (line 352) becomes:

```ts
if (this.lilyData == null) return;
```

**`packages/pwa/src/test/canvas.test.ts`**

Update tests that mutate `canvas.notesByQuant` or `canvas.ranges` to mutate `canvas.lilyData!.notesByQuant` / `canvas.lilyData!.ranges`. The seek() save/restore helper (lines 20-28) should snapshot `canvas.lilyData!.ranges` and restore it to `canvas.lilyData!.ranges`.

### 8. Test plan

1. **Type-check:** `pnpm run check:types` passes with the new aliases and `ReadonlyMap`/`readonly` shapes.
2. **Unit tests:** `pnpm run test:unit` passes, including all updated `canvas.test.ts` cases.
3. **New test in `canvas.test.ts`:** after init, `canvas.lilyData` is non-null and `canvas.lilyData.barCount === 139` (or whatever the parsed value is), asserting the collapsed field is populated.
4. **Regression test:** `draw()` returns early when `lilyData` is null. This is already exercised by the existing "empty dict/ranges" test if updated to set `canvas.lilyData = null`.

### 9. Dependencies

- Builds on closed #105, #106, #551.
- Related but distinct from open #644 (Ohm action-dict arity typing in `lily.ts`).
- None blocking.

### 10. Open questions

None.

### 11. Follow-up items

None.

### 12. Documentation impact

- `wiki/Architecture-Lily.md` and `wiki/Architecture-Canvas.md` describe the data structures and `MusicCanvas` fields. Update them after merge via the normal post-merge doc update route (or include in the PR if the wiki is treated as repo docs; check current repo-doc mapping before merging).

wainwright1000:
Strengthens the `LilypondData` types and collapses redundant `MusicCanvas` fields. Internal typing and refactor work, no user-facing behaviour change.

- `lily.ts`: outer-dimension aliases, a single canonical field, and deep-readonly ranges, so the parsed LilyPond data is harder to misuse and the ranges cannot be mutated in place.
- `MusicCanvas.ts`: collapses fields that duplicated the same state.
- `canvas.test.ts`: updated to the strengthened types.

Fixes #652

### DIFF
--- a/packages/pwa/src/test/canvas.test.ts
+++ b/packages/pwa/src/test/canvas.test.ts
@@ -2,7 +2,21 @@ import { MusicCanvas } from "../ts/MusicCanvas";
 import config from "../ts/config";
 import { colors, type Position } from "../ts/common";
 import { Duration, Note } from "../ts/music-classes";
+import { barCount as lilyBarCount, type Range } from "../ts/lily";
 MusicCanvas.define("music-canvas");
+
+// Install fixture singing-ranges on a canvas by swapping its whole `lilyData`
+// (since #652 `ranges` is a deeply-readonly SingingRanges, so build a fresh
+// mutable map and replace the object rather than mutating it in place).
+function setRanges(
+  c: MusicCanvas,
+  build: (r: Map<string, Range[]>) => void
+): void {
+  const r = new Map<string, Range[]>();
+  for (const [k, v] of c.lilyData!.ranges) r.set(k, [...v]);
+  build(r);
+  c.lilyData = { ...c.lilyData!, ranges: r };
+}
 
 var canvas: MusicCanvas | null;
 describe("MusicCanvas custom element", () => {
@@ -18,19 +32,17 @@ describe("MusicCanvas custom element", () => {
     vi.clearAllMocks();
   });
 
-  // seek() tests overwrite choir ranges on the shared beforeAll canvas; snapshot
-  // and restore the map per test so a fixture never leaks into a later test (#598).
-  // #704 also mutates notesByQuant, so snapshot that too.
-  let savedRanges: NonNullable<typeof canvas>["ranges"];
-  let savedNotesByQuant: NonNullable<typeof canvas>["notesByQuant"];
+  // seek() tests install fixture ranges by swapping the whole `lilyData` object
+  // (its `ranges` is deeply readonly since #652, so the original is never mutated
+  // in place). A shallow snapshot of the reference therefore restores it cleanly
+  // per test, so a fixture never leaks into a later test (#598, #652). The #704
+  // pulse tests also swap lilyData (for notesByQuant), so this snapshot covers them.
+  let savedLilyData: NonNullable<typeof canvas>["lilyData"];
   beforeEach(() => {
-    savedRanges = new Map();
-    for (const [key, value] of canvas!.ranges) savedRanges.set(key, [...value]);
-    savedNotesByQuant = canvas!.notesByQuant;
+    savedLilyData = canvas!.lilyData;
   });
   afterEach(() => {
-    canvas!.ranges = savedRanges;
-    canvas!.notesByQuant = savedNotesByQuant;
+    canvas!.lilyData = savedLilyData;
   });
 
   it("Check that the canvasent contains a canvas", async () => {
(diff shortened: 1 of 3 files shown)


wainwmr (approved):
OK

PR #739, opened 2026-06-30
perf: cache scroll widths in MusicScore to avoid per-frame reflow

### LINKED ISSUE #692: MusicScore: cache frame/score widths to remove per-frame forced reflow during playback
- **Type:** tech debt
- **Area:** Score
- **Status:** Todo
- **Difficulty:** S

## 2. Description

During playback, `MusicScore.scrollSmooth()` reads `this.offsetWidth` and `this.svg.getBoundingClientRect().width` on every call. It is called from `setBar()` on every fractional-bar update (every frame), *after* `highlight()`'s `setAttribute` writes have dirtied layout, forcing a synchronous reflow of the large score SVG each frame. The two widths are invariant during playback, so the per-frame read is wasted work. No user-visible symptom; a steady main-thread cost during playback that is fully eliminable.

## 3. Reachability

N/A — tech debt, not a Bug ticket (real, not broken). Per-frame inefficiency, no wrong result.

## 4. Current state

`packages/pwa/src/ts/MusicScore.ts`:

- `scrollSmooth()` (line ~315) reads `this.offsetWidth` (line ~322, `frameWidth`) and `this.svg.getBoundingClientRect().width` (line ~323, `scoreWidth`) every call.
- Called from `setBar()` (line ~301) after `highlight()` (line ~300) on every fractional-bar update during playback.
- There is **no** resize handling in the component (no `resize` listener, no `ResizeObserver`).
- The score SVG height is fixed (`music-score` is `height: 350px`), so `scoreWidth` (= height x viewBox aspect) is constant during playback and changes only when the score changes (choir switch, `#loadScore`). `frameWidth` (the element's visible width) changes only on viewport resize.

## 5. Root cause

Read-after-write layout thrash: `highlight()` dirties layout, then `scrollSmooth()`'s width reads force a synchronous reflow of the ~25,000px-wide rendered score SVG, every frame. Because both widths are constant during playback, the reads recompute an invariant.

**Measured (headless Chromium, real score SVG at rendered scale 25186 x 342 px, VERA 2026-06-22):** read-after-write ~0.37-0.38 ms/frame forced reflow (stable across runs); cached (read once) ~0.2 ms for 200 frames; projection ~22 ms/s at 60 fps (~2% of the 16.7 ms frame budget). Desktop CPU; a slower device scales it up.

## 6. Scope

**In scope:** cache `frameWidth` and `scoreWidth`; refresh on score load and on resize; read the cached values in `scrollSmooth()`. **Out of scope:** the scroll-position math (unchanged), `highlight()` logic, and the other refactor items in `refactor-MusicScore.ts.md` (items 1-11 are in unrelated methods, not adjacent — fold: none). No behaviour change intended.

## 7. Recommended fix

In `packages/pwa/src/ts/MusicScore.ts`:

1. Add private numeric fields, e.g. `#frameWidth` and `#scoreWidth`.
2. Cache `#scoreWidth = this.svg.getBoundingClientRect().width` inside the existing post-paint `requestAnimationFrame` in `#loadScore` (the same rAF added for #92, where layout is clean), so it re-caches on every score/choir change.
3. Cache `#frameWidth = this.offsetWidth` on `connectedCallback`, and refresh it from a `ResizeObserver` observing `this` (the `music-score` element); disconnect the observer in `disconnectedCallback`.
4. In `scrollSmooth()`, replace the `this.offsetWidth` and `this.svg.getBoundingClientRect().width` reads with `#frameWidth` / `#scoreWidth`. Keep the existing `if (this.svg == null) return` guard; if the cache is not yet populated, fall back to a direct read so the first call is correct.
5. No change to the `idealPos` scroll math.

## 8. Test plan

Extend the existing MusicScore tests (`packages/pwa/src/test/setbar.test.ts` and/or `score.test.ts`). Tests assert **observable outcomes**, not call counts:

- **Scroll position preserved:** with `offsetWidth` and `svg.getBoundingClientRect` stubbed (as existing tests do), `setBar(b)` produces the same `scrollArea.scrollLeft` for a representative bar as the pre-cache behaviour.
- **`frameWidth` refreshes on resize:** drive the resize path (invoke the `ResizeObserver` callback / resize handler) with a new `offsetWidth` and assert `scrollSmooth()` uses the new value.
- **`scoreWidth` refreshes on score load:** after a choir/score change (`await music-score-ready`), the cached `scoreWidth` reflects the new SVG (guards the stale-cache-across-choir-change risk).

**Not unit-testable:** the forced-reflow elimination cannot be tested in jsdom (no real layout); it is verified by the benchmark in section 5. Do not write a reflow-count test.

## 9. Dependencies

None. Related: #92 (load-time reflow, fixed in #354 via `requestAnimationFrame`) addressed the *load* path; this is the per-frame *playback* path. Not a blocker.

## 10. Open questions

None.

## 11. Follow-up items

None. Fold: none (the other `refactor-MusicScore.ts.md` items are not adjacent to `scrollSmooth`).

## 12. Documentation impact

None expected. `packages/pwa/src/ts/*` maps to no `doc/*.md`; behaviour is unchanged. The `Architecture-Scores.md` wiki page describes the scroll mechanism, but as a post-merge archidoc and only if the description needs it (behaviour preserved, so likely none).

wainwright1000:
Caches the scroll widths in `MusicScore` so the per-frame scroll handler no longer forces a layout reflow on every frame. Performance only, no user-facing behaviour change.

- `MusicScore.ts`: compute the widths once and reuse them, instead of reading layout-dependent values each frame.
- `score.test.ts`: covers the cached path.

Fixes #692

### DIFF
--- a/packages/pwa/src/test/score.test.ts
+++ b/packages/pwa/src/test/score.test.ts
@@ -8,6 +8,55 @@ import {
 } from "./helpers";
 import { makeFixtureSvg } from "./fixtureScore";
 
+// Capture ResizeObserver callbacks so tests can drive the resize path, and
+// count disconnects so the teardown test can assert the observer is released.
+const resizeObserverCallbacks = new WeakMap<Element, () => void>();
+let resizeObserverDisconnects = 0;
+
+vi.stubGlobal(
+  "ResizeObserver",
+  class {
+    #callback: ResizeObserverCallback;
+    #target: Element | null = null;
+
+    constructor(callback: ResizeObserverCallback) {
+      this.#callback = callback;
+    }
+
+    observe(target: Element) {
+      this.#target = target;
+      resizeObserverCallbacks.set(target, () =>
+        // The entry shape is intentionally minimal: the production callback
+        // ignores its arguments and re-reads offsetWidth / getBoundingClientRect,
+        // so only `target` is real. Enrich this if the callback ever reads the
+        // rect.
+        this.#callback(
+          [
+            {
+              target,
+              contentRect: { width: 0, height: 0 } as DOMRectReadOnly,
+            } as ResizeObserverEntry,
+          ],
+          this as unknown as ResizeObserver
+        )
+      );
+    }
+
+    disconnect() {
+      resizeObserverDisconnects++;
+      if (this.#target) {
+        resizeObserverCallbacks.delete(this.#target);
+        this.#target = null;
+      }
+    }
+  }
+);
+
+function triggerResize(target: Element) {
+  const callback = resizeObserverCallbacks.get(target);
+  if (callback) callback();
+}
+
 // Polyfill DOMPoint for jsdom
 if (typeof DOMPoint === "undefined") {
   globalThis.DOMPoint = class DOMPoint {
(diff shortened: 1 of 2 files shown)


wainwmr (approved):
OK

PR #741, opened 2026-06-30
refactor: type the LilyPond parse operation

### LINKED ISSUE #700: Type the LilyPond parse operation and related parser type holes
## Description

The Ohm semantics `parse` operation in `packages/pwa/src/ts/lily.ts` is added without an explicit result type. The action dictionary returns a mix of `undefined`, `Note`, `Rest`, `BarLine`, `Component[]`, `Duration`, `number` and `string`, and the `_iter` action returns `children.map((c) => c.parse())`. TypeScript therefore infers the operation result as `any`, so every call site is unchecked:

- `getDuration` indexes `duration.parse()[0]` (line 173).
- `durationScaled` reads `.duration` / `.dotted` from `duration.parse()` (lines 232-235).
- `RelativeClause` assigns `music.parse()` to `scores[v[0]]` (line 188) and returns `scores[v]` (line 190).
- `fraction` casts `this` through `unknown` to `ohm.Node` (line 238).

A deliberately wrong usage probe passed `pnpm exec tsc --noEmit`, confirming the call sites are effectively `any`. This is distinct from closed Ticket #644, which concerned action *arity*; the generated `LilypondActionDict<T>` already arity-checks the dictionary (see `wiki/refactor-lily.ts.md` False Positive F3).

## Reachability

Not applicable. This is a Tech Debt ticket (Maintainability, #484 Q7); there is no user-facing defect on the production path.

## Current state

- `setupLilypondParser` is declared `function setupLilypondParser(): ohm.Semantics` (line 166), discarding the generated `LilypondSemantics` type from `ly-grammar.ohm-bundle.d.ts`.
- `var s = lyGrammar.createSemantics()` (line 167) widens the semantics object; `s.addOperation("parse", { ... })` spans lines 182-251.
- The action dictionary mixes return kinds:
  - `Version`, `Include`, `command`, `BarCheck` return `undefined`.
  - `barline` returns `BarLine`.
  - `repeatedNote`, `note`, `rest` return `Note` or `Rest`.
  - `duration` returns `Duration`.
  - `durationScaled` returns `Duration`.
  - `fraction` returns `number`.
  - `variable` returns `string`.
  - `_iter` returns `unknown[]` (the array of child `parse()` results).
- The dead/wrong `return scores[v]` in `RelativeClause` is already recorded as refactor report item 24.
- The `this as unknown as ohm.Node` cast is already recorded as part of refactor report item 25.

## Root cause

The operation is not modelled as a typed algebra. Because `setupLilypondParser` widens the semantics object to `ohm.Semantics`, the operation name and result type are not constrained by the generated grammar types. Ohm's `LilypondActionDict<T>` lets the dictionary be typed, but the `T` is never fixed and the semantics object is not declared `LilypondSemantics`, so call sites fall back to `any`.

## Scope

In scope:

1. Declare `setupLilypondParser` as returning `LilypondSemantics` and use the generated type for the semantics object.
2. Introduce a `ParseValue` union that covers every action's result (`Component[] | Component | Duration | number | string | undefined`).
3. Type the `parse` operation with `addOperation<ParseValue>("parse", actionDict)` and adjust the action dictionary so every action returns `ParseValue`.
4. Fix the three unsafe call sites:
   - `getDuration`: keep `duration.parse()[0]` but the result is now typed as `Duration | undefined`.
   - `durationScaled`: narrow `duration.parse()` to `Duration` (the inner `duration` rule always returns `Duration`).
   - `RelativeClause`: narrow `music.parse()` to `Component[]` and fix the dead `return scores[v]` to `return scores[v[0]]` (or omit the return since the caller ignores it).
   - `fraction`: remove the `this as unknown as ohm.Node` cast; `this` inside an action is already `ohm.NonterminalNode`, which has `sourceString`.
5. Add a type-level regression probe that deliberately wrong `.parse()` usage fails `check:types`.

Explicitly deferred (not adjacent to the parse operation):

- General `lily.ts` hygiene (`var` → `let`/`const`, loose equality, magic numbers, `for...in`).
- Orphaned `noteToPitchClass` / `romanise` deletion (Workbench Item #1043).
- `LilypondData` structural aliases / deep-readonly ranges (Ticket #652).
- Build-time precompute of the parse (Ticket #693).

## Recommended fix

1. In `packages/pwa/src/ts/lily.ts`:

   a. Add a `ParseValue` union near the top of `setupLilypondParser` (or just above it):

      ```ts
      type ParseValue = Component[] | Component | Duration | number | string | undefined;
      ```

   b. Change `setupLilypondParser`'s return type and the semantics variable:

      ```ts
      function setupLilypondParser(): LilypondSemantics {
        const s = lyGrammar.createSemantics();
        ...
      }
      ```

   c. Type the operation:

      ```ts
      s.addOperation<ParseValue>("parse", {
        ...
      });
      ```

   d. For actions that currently return nothing, explicitly return `undefined` (or rely on implicit `undefined`). Ensure `Component` actions (`barline`, `repeatedNote`, `note`, `rest`) return `Component` values; `Note`/`Rest`/`BarLine` are subclasses of `Component`, so they are assignable.

   e. In `_iter`, the implementation returns `children.map((c) => c.parse())`. This is typed as `ParseValue[]`, which is not assignable to `ParseValue`. Either:
      - cast the result: `return children.map((c) => c.parse()) as unknown as ParseValue;` with a comment that Ohm's default iteration behavior returns an array, or
      - define `ParseValue` to include arrays and narrow at every call site.
      The first option is smaller and keeps call-site typing precise; document the cast.

   f. In `getDuration`, assert the expected shape:

      ```ts
      const d = (duration.parse() as Duration[])[0];
      ```

      (The `durationScaled?` optional rule is an iteration node, so `_iter` returns `ParseValue[]`; the only element kind is `Duration`.)

   g. In `durationScaled`:

      ```ts
      const x = duration.parse() as Duration;
      const m = multiplier.parse()[0] as number;
      return new Duration(x.duration, x.dotted, m);
      ```

   h. In `RelativeClause`:

      ```ts
      const v = variable.parse() as string;
      const components = music.parse() as Component[];
      if (v != undefined) {
        scores[v] = components;
      }
      return scores[v];
      ```

      This also fixes the dead `scores[v]` array-key bug (refactor report item 24).

   i. In `fraction`, replace the cast with:

      ```ts
      const text = this.sourceString;
      ```

      Remove the `this as unknown as ohm.Node` cast (refactor report item 25).

2. Add a type regression probe in `packages/pwa/src/test/lily.test.ts` (or a dedicated `lily-types.test.ts`) that is guarded by a TypeScript `@ts-expect-error` comment. For example, a test helper that tries to index a non-array `.parse()` result or read a non-existent property. The test itself can be empty; the value is that `check:types` fails if the error disappears.

3. Run `pnpm --filter @spem/pwa run check:types` and the full fast unit suite. No runtime behaviour should change.

## Test plan

1. **Type regression probe.** Add an `@ts-expect-error` test case that exercises a deliberately wrong `.parse()` call site (e.g. `duration.parse()[0].nonexistent`). It must fail `check:types` before the fix and pass after the fix (because the error is now expected).
2. **Parser output shape tests.** Extend `lily.test.ts` to assert that `processLilypond()` still returns `notesByQuant`, `ranges`, `barCount`, and `frLocations` with the same shapes and values as before. These already exist; ensure they still pass.
3. **Duration parsing tests.** Assert `Duration` instances produced by `durationScaled` (`* fraction`) have the expected `duration`, `dotted`, and `multiplier` fields.
4. **Fraction parsing tests.** Assert `fraction` returns `number` for integer and `n/m` inputs.
5. **Full suite.** Run `pnpm --filter @spem/pwa test:unit` and `pnpm --filter @spem/pwa run check:types`. Both must pass.

## Dependencies

- Investigation Item #1044 — closed; confirmed the result-type gap is distinct from closed Ticket #644.
- Refactor report item 24 (`RelativeClause` dead return) — folded into this ticket.
- Refactor report item 25 (`this as unknown as ohm.Node` cast) — folded into this ticket.
- Ticket #652 (LilypondData typing) — related but not blocking; #652 touches data structures returned by `processLilypond`, not the Ohm operation.

## Open questions

None.

## Follow-up items

None. The two related refactor-report findings (items 24 and 25) are folded into the Recommended fix.

## Documentation impact

None expected. The change is confined to `packages/pwa/src/ts/lily.ts` and its test file.

wainwright1000:
Adds types to the LilyPond parse operation in `lily.ts`, so the parse path is type-checked rather than relying on implicit `any`. Internal typing refactor, no behaviour change.

Fixes #700

### DIFF
--- a/packages/pwa/src/test/lily.test.ts
+++ b/packages/pwa/src/test/lily.test.ts
@@ -10,6 +10,7 @@ const {
   setupLilypondParser,
   noteToPitchClass,
   semantics,
+  parseMatch,
   resetLilypondCache,
 } = exportedForTesting;
 import { Note, Duration } from "../ts/music-classes";
@@ -66,6 +67,12 @@ describe("lilypond parsing tests", () => {
     const match = lyGrammar.match("3", "fraction");
     expect(match.succeeded()).toBe(true);
     expect(semantics(match).parse()).toBe(3);
+  });
+
+  it("parse operation result is typed, not any", () => {
+    const match = lyGrammar.match("3/4", "fraction");
+    // @ts-expect-error parse() returns a union, not any; property access is now checked.
+    expect(parseMatch(match).wrongProperty).toBeUndefined();
   });
 
   it("processLilypond", () => {
--- a/packages/pwa/src/ts/lily.ts
+++ b/packages/pwa/src/ts/lily.ts
@@ -2,7 +2,9 @@
 // SPDX-License-Identifier: MIT
 
 import config from "./config";
-import lyGrammar from "../ohmjs/ly-grammar.ohm-bundle";
+import lyGrammar, {
+  type LilypondSemantics,
+} from "../ohmjs/ly-grammar.ohm-bundle";
 import * as ohm from "ohm-js";
 import { Duration, BarLine, Note, Rest, Component } from "./music-classes";
 import spem from "@lilypond/spem.ly?raw";
@@ -32,7 +34,23 @@ export var scores: { [id: string]: Component[] } = {};
 // Set up Lilypond parser
 // -----------------------------------------------------
 
-var semantics: ohm.Semantics = setupLilypondParser();
+type ParseValue =
+  | Component[]
+  | Component
+  | Duration
+  | number
+  | string
+  | undefined;
+
+interface LilypondOperations {
+  parse(): ParseValue;
+}
+
+var semantics: LilypondSemantics = setupLilypondParser();
+
+function parseMatch(match: ohm.MatchResult): ParseValue {
+  return (semantics(match) as unknown as LilypondOperations).parse();
+}
 
 function romanise(num: number) {
   var lookup: { [index: string]: number } = {
@@ -163,14 +181,14 @@ export function detectFalseRelations(
   return frLocations;
 }
 
-function setupLilypondParser(): ohm.Semantics {
-  var s = lyGrammar.createSemantics();
+function setupLilypondParser(): LilypondSemantics {
+  var s: LilypondSemantics = lyGrammar.createSemantics();
 
   // If lilypond input has no duration, use lastDuration; use lastNote if note name is missing
   var lastNote: Note, lastDuration: Duration;
 
   function getDuration(duration: ohm.Node) {
-    var d = duration.parse()[0];
+    var d = (duration.parse() as Duration[])[0];
     if (d == undefined) {
       d = lastDuration;
     } else {
@@ -179,15 +197,16 @@ function setupLilypondParser(): ohm.Semantics {
     return d;
   }
 
-  s.addOperation("parse", {
+  s.addOperation<ParseValue>("parse", {
     Version(_, _2, _v, _3) {},
     Include(_, _2, _filename, _3) {},
     RelativeClause(variable, _, _2, _3, _4, music, _6) {
-      const v = variable.parse();
+      const v = variable.parse() as string[];
+      const components = music.parse() as Component[];
       if (v[0] != undefined) {
-        scores[v[0]] = music.parse();
+        scores[v[0]] = components;
       }
-      return scores[v];
+      return scores[v[0]];
     },
     Component(comp) {
       const c = comp.parse();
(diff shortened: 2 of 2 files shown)


wainwmr (approved):
Ok

PR #743, opened 2026-06-30
fix: pin row-resize cursor during splitter drag (#709)

### LINKED ISSUE #709: Fix splitter drag cursor: no resize cursor shown during drag
## Type / Area / Difficulty

- **Type:** bug
- **Area:** UI
- **Difficulty:** XS

## 1. Symptom

When you drag the splitter between the score and the canvas to resize them, the
pointer does not show a resize cursor as it should. As you drag to the top or
bottom limit (where the splitter stops but the pointer keeps moving), the cursor
flips to a crosshair over the score area instead of staying a resize cursor. The
resize itself still works; only the cursor feedback is wrong.

## 2. Steps to reproduce

1. Open the player on a desktop browser (pointer device).
2. Press and hold the splitter bar between the score (top) and canvas (bottom).
3. Drag up or down, especially as far as it will go.
4. Watch the cursor.

## 3. Expected behaviour

Throughout the drag the cursor stays a vertical-resize cursor (`row-resize`, the
same cursor the splitter shows on hover), including when the splitter clamps at
its top/bottom limit and the pointer moves over the score.

## 4. Actual behaviour

The cursor is not pinned during the drag. Hovering the splitter shows
`row-resize` (from CSS), but once the splitter clamps or the pointer outpaces it,
the cursor reverts to the score area's `crosshair`. The JavaScript that is meant
to pin the cursor for the duration of the drag does nothing, because it assigns
an invalid value.

## 5. Affected files

- `packages/pwa/index.ts:106` — splitter `mousedown` handler:
  `document.body.style.cursor = "col.resize";`
- `packages/pwa/index.ts:118-120` — `mouseup` handler resets `document.body.style.cursor = ""`.
- `packages/pwa/src/scss/style.scss:341` — `.splitter { cursor: row-resize; }` (the intended/correct cursor).
- `packages/pwa/src/scss/style.scss:353` — the score/canvas area is `cursor: crosshair` (what shows instead).

## 6. Root cause hypothesis

Two errors in one line (`index.ts:106`):

1. **Invalid value:** `"col.resize"` uses a dot, not a hyphen. `col.resize` is
   not a valid CSS `cursor` keyword, so the browser silently ignores the
   assignment and `document.body.style.cursor` is never set during the drag.
2. **Wrong axis:** even corrected to `"col-resize"`, that is a *column* (horizontal)
   resize cursor, but the splitter resizes **height** (`newHeight = e.clientY - ...`,
   `index.ts:113-115`). The splitter's own CSS uses `row-resize`, so the JS should
   match it.

The fix is `document.body.style.cursor = "row-resize";`.

First seen: commit `ee846a52` (2026-05-04).

## 7. Test code (failing test that proves the bug)

A jsdom unit test on the splitter wiring (the cursor is set on `document.body`):

```ts
it("pins a row-resize cursor on body while dragging the splitter", () => {
  const splitter = document.querySelector(".splitter") as HTMLElement;
  splitter.dispatchEvent(new MouseEvent("mousedown", { bubbles: true }));
  expect(document.body.style.cursor).toBe("row-resize"); // currently "col.resize" (or "" if jsdom rejects it)
  document.dispatchEvent(new MouseEvent("mouseup", { bubbles: true }));
  expect(document.body.style.cursor).toBe("");
});
```

(Note: jsdom does not always validate CSS keyword values, so it may store
`"col.resize"` literally; the first assertion fails either way against the
current code.)

## 8. Acceptance criteria

- During an active splitter drag, `document.body.style.cursor` is `row-resize`.
- On `mouseup` it is cleared back to `""`.
- The resize behaviour itself is unchanged.

## 9. Risk and scope

Tiny. One-word change in one line. No behaviour change beyond the cursor. The
only adjacent code is the `mouseup` reset, which already clears the cursor and
needs no change.

## 10. Notes

- Found by the index.ts bug-hunt (2026-06-23). Two of five blind reviewers
  (comment-analyzer, silent-failure-hunter) flagged the invalid cursor
  independently; the wrong-axis point is confirmed by the splitter's own CSS
  (`row-resize`).
- Low severity (cosmetic; resize works) but a clear, correct, one-word fix.

- Tele

wainwright1000:
Pins the cursor to `row-resize` for the duration of a splitter drag, so it no longer reverts to the default pointer when the cursor briefly leaves the splitter handle mid-drag.

Changes: `src/scss/style.scss` (cursor rule) and `index.ts` (drag state), with unit (`src/test/splitter.test.ts`) and e2e (`e2e/splitter-cursor.spec.ts`) coverage.

Fixes #709


wainwmr (approved):
OK

PR #744, opened 2026-06-30
fix: stop service worker activating before user refresh (#710)

### LINKED ISSUE #710: Stop the PWA service worker activating before the user clicks Refresh
## Symptom

A user running the installed/cached PWA gets a new version applied **without** clicking the "Update available → Refresh" toast. The service worker activates as soon as the new version is fetched, instead of waiting for the user's consent. In the worst case the new worker seizes the already-open tab mid-session, mixing the old in-memory JavaScript with newly-precached assets (the classic `skipWaiting` hazard the prompt model exists to avoid). The "Update available" toast the app shows therefore becomes misleading — the update has already (partly) happened.

## Steps to reproduce

1. Deploy version A; open the installed PWA so the service worker caches it.
2. Deploy version B.
3. Revisit / leave the PWA open. The new service worker installs and **activates immediately** (and, via `clientsClaim`, claims the open tab) rather than waiting in the background for the user to click Refresh.

(The exact user-visible manifestation — silent update vs misleading toast vs a broken mid-session shell — is browser-dependent and worth confirming with a real two-deploy test; the config-vs-contract contradiction below is deterministic regardless.)

## Expected behaviour

Per the app's documented and implemented update model, a new service worker **waits in the background**; the app surfaces the "Update available" toast; the update applies **only** when the user clicks Refresh (which triggers `updateSW(true)`). `doc/BUILD.md:189` states this verbatim: *"`registerType: "prompt"` — new service workers wait in the background; the app surfaces a toast for the user to refresh."*

## Actual behaviour

The generated service worker calls `self.skipWaiting()` (from `skipWaiting: true`) in its install handler, so it activates immediately without waiting for the client's `SKIP_WAITING` message; `clientsClaim: true` then makes it claim open tabs on activation. The user's Refresh click (`updateSW(true)`) is rendered a no-op handshake.

## Affected files

- `packages/pwa/vite.config.ts:165-166` — `clientsClaim: true`, `skipWaiting: true` (inside the `VitePWA` workbox block), contradicting `:119` `registerType: "prompt"`.
- (intent sources, unchanged) `packages/pwa/src/ts/pwa-update.ts` (the `onNeedRefresh` toast + `updateSW(true)`-on-click), `packages/pwa/src/test/pwa-update-toast.test.ts` (asserts the prompt contract), `doc/BUILD.md:189`.

## Root cause hypothesis

`skipWaiting` and `clientsClaim` belong to the **autoUpdate** workbox recipe, not the **prompt** model. They were added (with `registerType: "prompt"`) in the same commit (db117a33, 2026-06-08) — a workbox-boilerplate copy-paste. The two settings silently override the wait-for-user behaviour that `registerType: "prompt"` + the toast are built to provide.

## Test plan / failing test

A true runtime assertion needs the deployed SW (Vitest cannot exercise SW activation — `pwa-update-toast.test.ts` passes because it tests the toast UI, not whether the SW waits, which is why this slipped past it and two prior config audits). Two options:

- **Static (cheap, CI-able):** after `vite build`, assert the generated service worker (`dist/sw.js` / generated SW) does **not** contain `self.skipWaiting()` / `skipWaiting()` while `registerType` is `prompt`. Fails today, passes after the fix.
- **E2E (definitive):** a two-deploy Playwright test that loads v1, swaps in v2's precache, and asserts the new SW stays in `waiting` until the Refresh handshake.

## Acceptance criteria

- A new service worker does not activate until the user clicks Refresh (the `updateSW(true)` handshake is the only path to activation).
- The "Update available" toast accurately reflects a *pending* update.
- `doc/BUILD.md:189`'s documented behaviour ("wait in the background") matches reality.

## Risk and scope

- Fix is a 2-line config change: set `skipWaiting: false` and `clientsClaim: false` (or omit them — those are the prompt-model defaults). `updateSW(true)` already performs the skip-waiting handshake on the user's click.
- Low risk: no change to first-install behaviour (first SW still activates normally with no waiting client). The doc is correct and needs no change — fix the config, not the doc.

## Notes

- Found by the 2026-06-23 bug-hunt reviewer pass on `vite.config.ts` (byte-identical to its last audit; the line-level audit and two prior passes missed it because it is a cross-file/cross-doc contradiction, not a single-line defect).
- Undermines the feature added in #503 ("Surface the PWA update prompt") and tested via #657. Not a duplicate of either (those add the toast and its test coverage; neither touches the `skipWaiting` setting).
- Residual: confirm the precise user-visible symptom with a two-deploy test before sizing; the contradiction itself is established.

- Tele

wainwright1000:
Stops a newly-installed service worker from activating before the user refreshes, so an update never takes control mid-session without the user's prompt.

Changes: `vite.config.ts` (PWA service-worker wait behaviour), with `src/test/vite-config-pwa.test.ts` coverage.

Fixes #710


wainwmr (approved):
OK

PR #803, opened 2026-07-13
perf(scores): split rebuild invalidation and parallelise LilyPond renders

### LINKED ISSUE #760: Cut full score-rebuild cost: split postprocess invalidation, parallelise renders
## Symptom

The score build's expensive path (a full LilyPond re-render of every score) fires more often and runs slower than the pipeline needs. Two independent wastes, verified on the live files, found by the 2026-07-04 spem-performance repass:

1. **A postprocess-only change forces a full re-render.** `postprocessSvg` overwrites the raw LilyPond SVG in place (`postprocessSvg.mjs:261`), so changed postprocessing can only be re-applied by re-rendering. Any edit to `postprocessSvg.mjs` (a file with active churn: #624, #555) invalidates all 16 SVGs via the `POSTPROCESS_SVG_MTIME` gate (`buildScores.mjs:220`) and re-runs `lilypond --svg` 16 times — tens of minutes serial — when re-postprocessing kept raw SVGs would take seconds. The scores-ci path filter plus its git-timestamp restore make this the CI cost of every postprocessor edit.
2. **Independent renders run serially.** Each render is independent (distinct input, distinct output) and LilyPond is single-threaded, yet the loop blocks on `execFileSync` one score at a time (`buildScores.mjs:255`, loops at `:288`/`:309`), idling 3 of 4 vCPUs on `ubuntu-latest`. A `spem.ly` edit invalidates both notations (16 renders), so this is the routine content-edit path, not an edge.

## Recommended fix

Two parts, one branch; each is independently landable if the other stalls.

1. **Split the invalidation granularity.** Render to a kept raw location (e.g. `<outDir>/.raw/<name>.svg`, gitignored) and write the postprocessed result to the final path. Split `needsRebuild` into two decisions: raw stale vs `maxLyMtime` → render + postprocess; raw fresh but final stale vs `POSTPROCESS_SVG_MTIME` → postprocess only. Cost: one extra copy of each SVG on disk. Ride-along while in `postprocessSvg`: memoise `parseVariables` (re-reads two large .ly files 16 times per build with identical arguments; recorded RR-only).
2. **Parallelise renders.** Promisified `execFile` with a small concurrency pool (~3, or `os.availableParallelism()` capped), buffering each render's output and printing per-file on completion so logs stay readable; fail-fast by aborting the pool on first error before the existing `rmSync` cleanup. Expected roughly 2.5-3x wall-time cut on the full-rebuild path.

## Test plan

The integration suite (`buildScores.integration.test.ts`) already pins mtime caching, notation isolation, and error handling; extend it: (a) a postprocessor-mtime-only change re-postprocesses without invoking lilypond (spy/fake on the binary); (b) a `.ly` change still renders; (c) parallel path preserves per-file failure cleanup and non-zero exit on first error; (d) output SVGs byte-identical to the serial build for a fixed fixture.

## Dependencies

None open. Touches the same `needsRebuild` region as the report's RR item 4 (the strict-`>` tie gate) — cheap to fold that one-character `>=` fix in while restructuring the decision.

## Notes

Found by the spem-performance reviewers-only repass (bug hunt, #1060 backfill). Full evidence and the cleared no-finding list (per-file process startup, findPartIndex scan, sort/statSync costs all below the bar) are in wiki/refactor-buildScores.mjs.md. The reviewer also independently re-derived the dormant OUP-edition postprocessing mismatch, already recorded as refactor-lilypond.md Tech Debt item 1; no change there.

- Tele

wainwright1000:
Fixes #760

A full score rebuild re-rendered and re-postprocessed every SVG even when only the postprocess step had changed, and the LilyPond renders ran serially.

Two cuts to that cost: invalidation is split so a postprocess-only change no longer forces a LilyPond re-render (the render cache keys the two stages separately), and the renders that are needed run in parallel. `packages/scores/build/buildScores.mjs` and `postprocessSvg.mjs` carry the change, with unit, integration and security tests updated alongside; `doc/BUILD.md` describes the split.

Note: this PR is `expensive` (it touches the score build, so merging busts the SVG cache and forces one full re-render).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,10 @@
 .worktree-offset
 node_modules
 dist
+# Kept raw LilyPond renders (#760): the un-postprocessed source of each SVG,
+# used to re-postprocess without re-rendering. An intermediate build artifact,
+# not committed — only the postprocessed SVGs under scores/ are tracked.
+packages/pwa/src/scores/**/.raw/
 *.ipynb
 *.pdf
 # Local Netlify folder
--- a/doc/BUILD.md
+++ b/doc/BUILD.md
@@ -58,9 +58,11 @@ Serves the contents of `dist/` locally.
 ## Regenerate SVG Scores
 
 The SVG files in `packages/pwa/src/scores/` are generated from LilyPond source files in
-`packages/scores/src/`. They are committed to git as source assets. `pnpm run build`
-regenerates them automatically via the `prebuild` step when `.ly` sources are
-newer than the generated SVGs.
+`packages/scores/src/` and committed to git as source assets. Regenerate them with
+`pnpm run build:scores` locally, or let the **Scores CI** workflow
+(`.github/workflows/scores-ci.yml`) rebuild and commit them when a push changes
+`packages/scores/**`. The app build (`pnpm run build`, and Netlify) uses the committed
+SVGs and does not regenerate them.
 
 To build scores manually:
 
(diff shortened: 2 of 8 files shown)


wainwmr (approved):
Ok

PR #804, opened 2026-07-13
perf: precompute the LilyPond parse at build time (removes the cold-load stall)

### LINKED ISSUE #693: lily.ts: precompute the LilyPond parse at build time (remove ~500ms cold-load stall)
- **Type:** tech debt
- **Area:** Lily

## What and why

`processLilypond()` parses `spem.ly` with Ohm at runtime, in `MusicCanvas.#init()` on cold load. The input is static and the output invariant, so every fresh page load re-derives a build-time constant. The module-level cache (#376) only spares repeats within a process.

## Measured (VERA, 2026-06-22)

Cold `processLilypond()`, 8 runs with the cache reset each time (Node/V8, desktop): **median 488 ms, mean 500 ms, first cold-JIT run 528 ms, range 420–676 ms.** Blocking, main-thread, on every cold load. A slower device is worse.

## Proposed direction

Parse at build time → ship `lilyData.json` → import/fetch it at runtime, removing the parse from cold load. May also let ohm-js and the grammar bundle drop from the production bundle.

## Classification

Real, not broken; qualifies under #484 **Q5 Performance** (large, measured, eliminable cold-load cost). This is an **architecture change** (build step, data format, runtime load path, the `barCount`/`Config` plumbing, and it touches `buildScores.mjs`), so it needs a design pass at specify time.

## Design decisions (resolved 2026-07-01)

Specified: see the [Specification comment](https://github.com/wainwmr/spem-player/issues/693#issuecomment-4851450304). Settled with Mark (question comment) and Andrew:

1. **Data model** — serialise the full `LilypondData`; the runtime rebuilds the two `Map`s only (it reads plain fields, so no class reviver is needed).
2. **Package ownership** — move the LilyPond parsing domain (parser, grammar, `music-classes`) into `packages/scores`; `pwa` consumes the emitted `lilyData.json`. One-way `pwa → scores`.
3. **Load path** — bundle `lilyData.json` and load it synchronously, preserving the current init path and `barCount` timing.
4. **Drop Ohm at runtime** — yes; `lily.ts` is the only runtime importer of `ohm-js` and the grammar.
5. **Dev regen / cache-bust** — regenerate `lilyData.json` in the score build alongside the SVGs, via the existing trigger.
6. **Scope** — Hugh Keyte only; OUP deferred.

Sequenced after #703 (dead-helper removal). Full detail, options, and test plan are in the Specification comment.

## Source

`wiki/refactor-lily.ts.md` item 20; measured by VERA 2026-06-22.

wainwright1000:
Fixes #693

The LilyPond note data was parsed in the browser on every cold load, stalling the app for roughly half a second before the score could respond.

That parse now runs at build time. `@spem/scores` gains a `build:lilydata` step that parses `spem.ly` through the Ohm grammar and writes `packages/scores/src/lily/lilyData.json`; the app loads that committed file instead of re-deriving it per visit. The Ohm grammar and its build move into `@spem/scores` alongside the data they produce, so the pwa package no longer carries the parser toolchain. A `lily-precompute` test fails if the committed data goes stale against `spem.ly`.

`doc/BUILD.md` and `doc/CONTRIBUTING.md` describe the new step. Note this PR is `expensive` (it touches the score build, so merging busts the SVG cache and forces one full re-render).


wainwmr (approved):
Wow

PR #805, opened 2026-07-13
fix: honour control changes during the audio load window

### LINKED ISSUE #764: Control changes during the audio-load window are silently ignored (play sets playing=false while awaiting)
## 2. Description

While a new choir, part, or recording MP3 is loading, a further control change is silently dropped. The audio keeps playing the file that was loading when the window opened, so playback no longer matches the displayed selection until the next control change.

## 3. Reachability

1. Start playback on any choir/part/recording other than the default, so the MP3 must be loaded.
2. While the loading spinner is visible (during `audio.load()` / `await audio.play()`), change choir, part, or recording again. Triggers:
   - holding ArrowUp/ArrowDown (auto-repeat; `REPEAT_EXEMPT_KEYS` in `index.ts`) while an MP3 is still loading;
   - rapid taps on the choir/part/recording widgets while a load is in flight.
3. The UI updates to the new selection, but when the original `play()` promise resolves the audio resumes on the file that was loading at step 1.

## 4. Current state

`packages/pwa/src/ts/MusicControls.ts` (`origin/main` @ `7f314e3`).

- `play()` (lines 235-283):
  - Line 240: `this.playing = false` is set for the whole load+await window whenever a new file is loaded.
  - Lines 245-247: `this.audio.src = newfile; this.audio.load(); this.audio.currentTime = ...`
  - Line 251: `await this.audio.play()`.
- `setChoir()` (lines 294-300):
  - Line 299: reload guard is `if (this.isPlaying()) this.play();`.
- `setRecording()` (lines 302-307):
  - Line 306: reload guard is `if (this.isPlaying()) this.play();`.
- `setPart()` (lines 309-315):
  - Line 314: reload guard is `if (this.isPlaying()) this.play();`.
- `isPlaying()` in `MusicElement.ts` (line 90-92) returns `this.playing`.

Because `playing` is `false` during the await, a control change arriving mid-load sees `isPlaying() === false` and skips its reload. When the original `play()` promise resolves, playback continues on the file loaded at the start of the window.

## 5. Root cause

The `playing` boolean is overloaded: it represents both "audio is currently audible" and "the user wants playback to continue". During the load window the audio is not yet audible, so `playing` is set to `false`, but that falsely tells the reload guards that the user has stopped playback. There is no separate signal for "a `play()` call is in flight".

## 6. Scope

**In scope:**
- Add a pending-load signal inside `MusicControls.ts` so reload guards fire while `play()` is awaiting.
- Make `play()` safe against concurrent calls (rapid repeated control changes).
- Keep the `#isLooping` guard intact.
- Add unit-test coverage for the mid-load control change and the loading-icon state.
- Replace the unnecessary `self` closure in `play()` with an arrow function (adjacent refactor fold).

**Out of scope:**
- Broader audio-state machine refactor (e.g. adopting the unused `Status` type from `common.ts`).
- Fixing the separate `isSameAudio` percent-encoding issue (#757) or the bar-input seek issue (#762).
- Adding `disconnectedCallback` or listener-reference cleanup.

## 7. Recommended fix

In `packages/pwa/src/ts/MusicControls.ts`:

1. Add private fields to track the in-flight `play()` call:

   ```typescript
   #playGeneration = 0;
   #pendingPlay: Promise<void> | null = null;
   ```

2. Replace the `self` closure in `play()` with an arrow function so the generation capture is lexical.

3. Rewrite `play()` to:
   - Capture the current generation at entry.
   - Compute the target file and, if different, start loading it, show the loading icon, and fire `music-controls-loading`.
   - Store `this.audio.play()` in `#pendingPlay`.
   - After `await`, update state only if the generation is still current.
   - In the `catch` branch, reset the icon only if the generation is still current.

   Example shape:

   ```typescript
   async play() {
     const gen = ++this.#playGeneration;
     const newfile = this.getMP3filename();

     if (!this.isSameAudio(newfile)) {
       this.playing = false;
       this.#showIcon("loading");
       this.fireEvent("music-controls-loading");
       this.audio.src = newfile;
       this.audio.load();
     }

     const playPromise = this.audio.play();
     this.#pendingPlay = playPromise;

     try {
       await playPromise;
     } catch {
       if (gen !== this.#playGeneration) return;
       this.#showIcon("play");
       this.playing = false;
       this.#pendingPlay = null;
       return;
     }

     if (gen !== this.#playGeneration) return;
     this.playing = true;
     this.#showIcon("pause");
     this.fireEvent("music-controls-playing");
     this.#pendingPlay = null;

     if (this.#isLooping) return;
     this.#isLooping = true;

     const loop = () => {
       this.bar = getBarFromTime(this.audio.currentTime, this.recording);
       const intbar = Math.floor(this.bar);
       if (this.barinput && Number(this.barinput.value) != intbar) {
         this.barinput.value = String(intbar);
       }
       this.fireEvent("music-controls-changed");

       if (this.isPlaying()) {
         this.#loopId = window.requestAnimationFrame(loop);
       } else {
         this.#isLooping = false;
       }
     };
     this.#loopId = window.requestAnimationFrame(loop);
   }
   ```

4. Add a helper that reports whether a load is in flight:

   ```typescript
   #isLoading(): boolean {
     return this.#pendingPlay !== null;
   }
   ```

5. Update the reload guards in `setChoir`, `setRecording`, and `setPart`:

   ```typescript
   if (this.playing || this.#isLoading()) this.play();
   ```

6. Update `pause()` to invalidate any in-flight `play()`:

   ```typescript
   pause() {
     this.playing = false;
     this.#playGeneration++;       // stale play() resolutions will no-op
     this.#pendingPlay = null;
     this.#isLooping = false;
     window.cancelAnimationFrame(this.#loopId);
     this.#showIcon("play");
     this.audio.pause();
     this.fireEvent("music-controls-paused");
   }
   ```

## 8. Test plan

Extend `packages/pwa/src/test/controls.test.ts`:

1. **Mid-load control change reloads the latest selection.**
   - Mock `HTMLAudioElement` so `audio.play()` returns a promise that does not resolve immediately.
   - Call `controls.setChoir(1)` while playing, which enters `play()` and leaves `#pendingPlay` set.
   - While the promise is pending, call `controls.setChoir(2)`.
   - Assert that `audio.src` ends with the choir-2 filename.
   - Resolve the first promise and assert it does not update `playing` or the icon (generation mismatch).
   - Resolve the second promise and assert `playing === true` and the pause icon is visible.

2. **Rapid part change during recording load lands on the latest part.**
   - Start `play()` for a recording change with a pending promise.
   - Call `setPart(0)`, `setPart(1)`, `setPart(2)` in quick succession.
   - Assert the final `audio.src` matches part 2.

3. **Loading icon is shown during the load window.**
   - Start `play()` with a pending `audio.play()` promise.
   - Assert `svgLoading.style.display === "block"` and `svgPlay.style.display === "none"`.
   - Resolve the promise and assert the pause icon is shown.

4. **Pause during load cancels the pending play.**
   - Start `play()` with a pending promise.
   - Call `controls.pause()`.
   - Resolve the original promise and assert `playing` remains `false` and the play icon is shown.

**Browser verification:** this fix is about the internal audio-loading state machine. The unit tests above exercise it with a mocked `Audio` element; a Playwright case is not required because the defect is not in browser chrome, focus, or cross-browser playback differences.

## 9. Dependencies

- #757 (`isSameAudio` percent-encoding reload) and #762 (bar-input seek) are related `MusicControls.ts` bugs but independent; no sequencing required.
- PR #743 touches `packages/pwa/index.ts` only for splitter-cursor code, so it does not conflict with this change.

## 10. Open questions

None.

## 11. Follow-up items

None.

## 12. Documentation impact

None expected. The change is confined to `MusicControls.ts` and its unit tests; no repo doc (`doc/*.md`) maps to this file.

- ZIMI

wainwright1000:
Fixes #764

Control changes made while the audio was still loading were silently discarded. `play()` set `playing = false` while awaiting the load, so a choir, part or bar change during that window (and a second press of play) was overwritten when the load resolved, leaving the UI showing one state and the player in another.

The load window is now guarded: state changes made during it are honoured when the load completes rather than clobbered. `MusicControls.ts` carries the fix, with the cases pinned in `controls.test.ts`.

### DIFF
--- a/packages/pwa/package.json
+++ b/packages/pwa/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@spem/pwa",
   "private": true,
-  "version": "2.8.9",
+  "version": "2.8.10",
   "description": "Helping singers learn Thomas Tallis Spem in alium",
   "type": "module",
   "dependencies": {
--- a/packages/pwa/src/test/controls.test.ts
+++ b/packages/pwa/src/test/controls.test.ts
@@ -655,4 +655,224 @@ describe("MusicControls custom element", () => {
 
     cancelSpy.mockRestore();
   });
+
+  // #764: while a new file is loading, play() holds playing=false across the
+  // await, so a control change arriving mid-load sees isPlaying()===false and is
+  // dropped; audio then resumes on the stale file. The fix adds an in-flight
+  // signal (reload guards fire during a load) and a generation counter (a stale
+  // play() resolution no-ops).
+  describe("play() load-window race (#764)", () => {
+    // A deferred play() mock we settle by hand, so the load window can be held
+    // open across further control changes. In a real browser a play() promise
+    // whose load is interrupted by a new load()/pause() REJECTS with AbortError,
+    // so both a resolve and a reject handle are exposed per call.
+    let playResolvers: Array<() => void>;
+    let playRejecters: Array<(reason?: unknown) => void>;
+
+    beforeEach(() => {
+      playResolvers = [];
+      playRejecters = [];
+      vi.spyOn(HTMLMediaElement.prototype, "play").mockImplementation(
+        () =>
+          new Promise<void>((resolve, reject) => {
+            playResolvers.push(resolve);
+            playRejecters.push(reject);
+          })
+      );
+    });
+
+    afterEach(() => {
+      // Re-apply the suite-wide immediate stub for later tests. mockRestore()
+      // would expose the real (jsdom-unimplemented) play(), so re-stub instead.
+      vi.spyOn(HTMLMediaElement.prototype, "play").mockReturnThis();
+    });
+
+    // Put the element in a loaded, playing choir-part state without opening a
+    // load window (a choir-part file, not default.mp3, so a later change reloads).
+    function loadedAndPlaying(elem: MusicControls): void {
+      elem.setPart(0); // Soprano
+      elem.setChoir(0); // Choir 1
+      elem.audio.src = "http://localhost/audio/ALC/Choir%201-Soprano.mp3";
+      elem.playing = true;
+    }
+
+    const last = <T>(a: T[]): T => a[a.length - 1];
+
+    it("a control change during the load window reloads the latest selection", async () => {
+      const elem = document.querySelector("music-controls") as MusicControls;
+      loadedAndPlaying(elem);
+      let playingEvents = 0;
+      elem.addEventListener("music-controls-playing", () => playingEvents++);
+
+      // Choir 2: a new file, so play() opens the load window and awaits.
+      elem.setChoir(1);
+      await Promise.resolve();
+      expect(playResolvers.length).toBe(1);
+
+      // Choir 3 arrives mid-load. Without the fix, playing===false drops it and
+      // audio stays on Choir 2.
+      elem.setChoir(2);
+      await Promise.resolve();
+      expect(decodeURIComponent(elem.audio.src)).toContain("Choir 3-Soprano");
+
+      // Resolve the STALE load alone: the generation guard must swallow it, so
+      // no state flip while the newer call is still pending. (Resolving both
+      // would pass whether or not the guard exists, so this isolates it.)
+      playResolvers[0]();
+      await Promise.resolve();
+      expect(elem.playing).toBe(false);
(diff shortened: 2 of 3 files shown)


wainwmr (approved):
OK

PR #807, opened 2026-07-13
fix: derive feedback status at read time instead of a stored mirror

### LINKED ISSUE #772: Feedback status is always 'paused' (State.status mirror is never written); derive at read time
## 2. Description

`current.status` (`packages/pwa/index.ts`) is a stored copy of playback state, read only by `updateFeedbackContext()` into the hidden feedback payload. It has exactly one writer: the `current` initialiser (`index.ts:88`, `status: "paused"`). No transition ever updates it. The `music-controls-playing` / `music-controls-paused` handlers (`handleAudioPlaying` / `handleAudioPaused`, `index.ts:540-548`) set only the canvas and score `playing` attributes, never `current.status`, and there is no `music-controls-loading` listener at all. So the diagnostic status submitted with feedback is **always `"paused"`**: wrong whenever the user is actually playing, and wrong during the audio-load window (never `"loading"`). The original ticket understated this as a missing `"loading"` case; the mirror is in fact dead on every submission during playback or loading.

The deeper issue is the shape: `current.status` is a hand-maintained mirror (the third, alongside the canvas/score `playing` attributes), so correctness depends on every present and future playback transition remembering to write it, and the type system cannot enforce that.

## 3. Reachability

Trigger an audio reload while playback is active or requested, then open the feedback modal before `audio.play()` resolves. Reload happens when the selected choir/part/recording maps to a not-yet-cached MP3 (`MusicControls.play()`, `packages/pwa/src/ts/MusicControls.ts:235-248`). The wrong value is visible only in the hidden diagnostic `context` field submitted with feedback; no visible UI breaks.

## 4. Current state

- `packages/pwa/src/ts/common.ts:70` defines `export type Status = "playing" | "paused" | "loading"`.
- `packages/pwa/src/ts/common.ts:73-81` defines `State` with a stored `status: Status`.
- `packages/pwa/index.ts:81-89` initializes `var current: State` with `status: "paused"`.
- `packages/pwa/index.ts:437-455` `updateFeedbackContext()` copies `current.status` into the hidden context input.
- `packages/pwa/index.ts:568-569` registers `music-controls-playing` -> `handleAudioPlaying` and `music-controls-paused` -> `handleAudioPaused`; those handlers (`index.ts:540-548`) set only the canvas/score `playing` attributes and never write `current.status`. No `music-controls-loading` handler exists. So after the `index.ts:88` initialiser, `current.status` has **no writer at all** (confirmed: a repo grep finds no assignment to `current.status` outside the initialiser).
- `packages/pwa/src/ts/MusicControls.ts:235-248` `play()` sets `this.playing = false`, shows the loading icon, fires `music-controls-loading`, assigns `audio.src`, calls `audio.load()`, then awaits `audio.play()`.
- No other production code reads or writes `State.status`.

## 5. Root cause

`State.status` is a hand-maintained mirror of playback state with **no writer**: none of the playing, paused, or loading transitions update it, so it is fixed at its initial `"paused"` for the life of the page. The mirror shape is the root cause: correctness depends on transitions writing a field that nothing writes, and the type system cannot enforce it. Deriving the status at read time from the media element removes the mirror entirely.

## 6. Scope

In scope:

- Derive the feedback status at read time inside `updateFeedbackContext()`.
- Remove the stored `status` field from `State` and from the `current` object.
- Keep the `Status` type (it remains the return type of the derivation).
- Add one unit test asserting `"loading"` is reported during the audio-load window.
- Preserve the #765 playing/paused test cases.

Out of scope:

- Refactoring the other playback-state representations inside `MusicControls` (icon argument, `this.playing`, `audio.paused`, `#isLooping`). Those are tracked in the refactor reports for `MusicControls.ts` and in tickets #764 and #773.
- Any visible change to the feedback UI.

## 7. Recommended fix

1. In `packages/pwa/src/ts/common.ts`, remove `status: Status` from the `State` interface. Keep `export type Status = "playing" | "paused" | "loading"`.
2. In `packages/pwa/index.ts`:
   - Remove `status: "paused"` from the `current` initializer.
   - In `updateFeedbackContext()`, replace `status: current.status` with `status: deriveFeedbackStatus()`.
   - Add a helper near `updateFeedbackContext()`:

     ```ts
     function deriveFeedbackStatus(): Status {
       const audio = controls.audio;
       if (audio.ended) return "paused";
       if (
         audio.readyState < HTMLMediaElement.HAVE_FUTURE_DATA ||
         audio.networkState === HTMLMediaElement.NETWORK_LOADING
       ) {
         return "loading";
       }
       return audio.paused ? "paused" : "playing";
     }
     ```

     This uses `controls.audio` as the single source of truth because the media element is the actual playback source and can distinguish loading and ended states. The canvas/score `playing` attributes are UI-derived and would reintroduce a mirror.
3. Leave `handleAudioPlaying()` and `handleAudioPaused()` unchanged; they continue to update the canvas and score attributes.
4. In `packages/pwa/src/test/feedback.test.ts`, add a test that mocks `controls.audio.readyState` and `networkState` to loading values, opens the feedback modal, and asserts `JSON.parse(hiddenContext.value).status === "loading"`.

## 8. Test plan

- Extend `packages/pwa/src/test/feedback.test.ts`:
  - Add `"reports loading status while audio is loading"`: mock `controls.audio.readyState = HTMLMediaElement.HAVE_NOTHING` (or `HAVE_METADATA`) and `controls.audio.networkState = HTMLMediaElement.NETWORK_LOADING`, call `showFeedback()`, and assert the hidden `context` input serializes `status: "loading"`.
  - Keep the existing #765 playing/paused assertions green by ensuring those tests set `controls.audio.paused` and `readyState` appropriately.
- Run `pnpm --filter @spem/pwa test` and confirm all feedback tests pass.
- No Playwright case is required: the value under test is hidden diagnostic data and no visible UI or browser-chrome behaviour changes.

## 9. Dependencies

- Blocked by #765 (the playing/paused telemetry fix must land first; this ticket removes the `current.status` field that #765 writes).
- Related: #764 (control changes during the load window) and #773 (play failure swallowing) touch adjacent `MusicControls` state but do not change the specification here.

## 10. Open questions

None.

## 11. Follow-up items

None.

## 12. Documentation impact

None expected. The change is internal to feedback telemetry; no repo docs map to `packages/pwa/index.ts`, `src/ts/common.ts`, or `src/test/feedback.test.ts`.

- JACKIE

wainwright1000:
Fixes #772

The feedback payload always reported the playback status as `paused`, whatever the player was actually doing. `State.status` was declared and read, but nothing ever wrote to it, so the field sat at its initial value and every feedback report carried it.

The fix removes the stored mirror rather than adding a write to it. `State` no longer carries `status`; the value is derived at read time from the audio element, in `deriveFeedbackStatus()`, at the moment the feedback payload is built. A value that cannot be stored cannot go stale. The `Status` union type is kept as the derivation's return type.

`doc/CONTRIBUTING.md` is updated to describe the derivation, since it documented the removed field.

### DIFF
--- a/doc/CONTRIBUTING.md
+++ b/doc/CONTRIBUTING.md
@@ -70,7 +70,11 @@ The central `State` interface (`packages/pwa/src/ts/common.ts`) tracks:
 - `choir` — 0 to 7
 - `part` — `"all"` or 0 to 4
 - `bar` — 0 to 139
-- `status` — `"playing"`, `"paused"`, or `"loading"`
+
+Playback status is not stored on `State`. The `Status` type
+(`"playing" | "paused" | "loading"`) is still exported, but the value is derived
+at read time from the audio element (`deriveFeedbackStatus()` in `index.ts`) when
+the feedback payload is built, so it cannot drift stale.
 
 State changes flow through `index.ts` helpers (`setChoir()`, `setPart()`, `setBar()`), which set
 attributes on elements. Components react via `attributeChangedCallback` and fire custom events
--- a/packages/pwa/index.ts
+++ b/packages/pwa/index.ts
@@ -9,6 +9,7 @@ import {
   MusicEventDetail,
   PartType,
   State,
+  Status,
   colors,
   toNum,
   toRecordingIndex,
@@ -85,7 +86,6 @@ var current: State = {
   choir: 0,
   part: "all",
   bar: 0,
-  status: "paused",
 };
 
 // TODO: Change dark mode to moon/sun icons
@@ -449,6 +449,25 @@ function showFeedbackResult(message: string) {
   setTimeout(() => showFeedback(false), 1500);
 }
 
+// Derive the playback status from the audio element at read time, rather than
+// mirroring it into a stored State field that every handler must remember to
+// update (the field was the stale-status bug this replaces). networkState is
+// checked first: it reports NETWORK_LOADING through the whole fetch, including
+// the load window where play() has already flipped audio.paused to false, so a
+// paused-first order would misreport a loading track as playing. The trade is
+// that mid-play buffering (a not-yet-fully-fetched track that is audibly
+// playing) also reads "loading"; the audio element alone cannot tell that from
+// the pre-play load window. Distinguishing them needs MusicControls' own state
+// (a status() getter) — deferred to a follow-up. Guard the controls read so a
+// missing element degrades to a default rather than throwing and dropping a
+// feedback submit, matching updateFeedbackContext's other null-guards.
+function deriveFeedbackStatus(): Status {
+  const audio = controls?.audio;
+  if (!audio) return "paused";
+  if (audio.networkState === HTMLMediaElement.NETWORK_LOADING) return "loading";
+  return audio.paused ? "paused" : "playing";
+}
+
 function updateFeedbackContext() {
   if (!hiddenFeedbackForm) return;
   const contextInput = hiddenFeedbackForm.querySelector<HTMLInputElement>(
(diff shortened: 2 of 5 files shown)


wainwmr (approved):
OK

PR #808, opened 2026-07-13
test: fail e2e tests on captured page errors

### LINKED ISSUE #775: Fail e2e tests on page errors: masked crashes currently pass
## Symptom

Uncaught page exceptions, unhandled promise rejections, and console errors never fail a Playwright test, and no spec in `packages/pwa/e2e/` installs a `page.on("pageerror")` or console listener (verified by grep). The cost is a false-green channel across the whole e2e layer: the SUT updates the asserted UI before doing the dependent work — `play()` sets the icon (`MusicControls.ts:259`) before `fireEvent("music-controls-playing")` (`:260`), `pause()` likewise (`:289` before `:290-291`) — so an exception in any `music-controls-*` listener (`index.ts:568-569`) or in the rAF loop (`getBarFromTime`, `MusicControls.ts:268`) leaves the playback pipeline (score highlight, canvas) dead while every icon assertion stays green. `index.ts:314/318` also invoke `controls.play()` as a floating promise, so a throw outside its internal try is an invisible unhandled rejection. Separately, the un-mocked media path emits zero signal: a 404 or decode failure on `/audio/...` sets `audio.error` and fires an unlistened error event while the mocked `play()` resolves regardless, so a dead audio asset pipeline passes.

## Steps to reproduce

1. Inject `throw new Error("x")` into the `music-controls-playing` handler in `packages/pwa/index.ts`.
2. `pnpm run e2e` (or just `playback-toggle.spec.ts`).
3. The suite passes.

## Expected behaviour

An uncaught page exception, an unhandled rejection, a console error, or a failed `/audio/` request during a test fails that test.

## Actual behaviour

All pass; the errors are invisible to the runner.

## Affected files

- `packages/pwa/e2e/helpers/` (new fixture module)
- `packages/pwa/e2e/*.spec.ts` (wire-in; 8 specs)

## Root cause hypothesis

Playwright's default semantics ignore page errors, and the e2e layer grew spec-by-spec with no shared fixture (only `helpers/computed-style.ts`, #714). Nothing established the error-capture convention when the first spec was written.

## Test code

Characterisation experiment, not a unit test: the injected-throw reproduction above must flip from green to red once the fixture lands. N/A as a failing unit test.

## Acceptance criteria

- A page exception thrown in any spec's page fails that test with the exception message.
- A `console.error` fails the test (with an explicit, commented allowlist if any spec legitimately logs one).
- A failed (>=400 or aborted) request to a `/audio/` URL fails the playback spec.
- All existing e2e specs remain green across chromium/firefox/webkit (any newly surfaced masked error is triaged, not suppressed).

## Risk and scope

Test-only; no production code. The fixture may surface currently-masked errors in other specs — that is the point; each surfaced error is triaged (possible new tickets). Small flake surface if a browser logs incidental console noise; the allowlist handles it explicitly.

## Notes

From the bug-hunt of `playback-toggle.spec.ts` (2026-07-07; silent-failure-hunter findings F1/F2, verified). Closes the throw-half of the #398-test vacuity noted in the same hunt (the reflow-observable assert is in the companion spec-hardening ticket). Related but orthogonal: #726 (workflow-level failure notification; this ticket is test-level detection). Companion ticket: harden playback-toggle spec (filed together).

- Claude

wainwright1000:
Fixes #775

An e2e test could pass while the page it was driving threw errors. Playwright reports assertion failures, but an uncaught page error, a `console.error`, or a mis-served audio file left no mark on the run, so a genuine crash could hide behind a green suite.

Every spec now imports `test` and `expect` from a shared fixture, `e2e/helpers/page-errors.ts`, which captures three channels (uncaught page errors, `console.error`, and wrong content-type responses under `/audio/`) and asserts they are empty at teardown. A new `page-errors.spec.ts` covers the fixture itself. An ESLint rule enforces the import convention, so a future spec cannot quietly opt out by importing Playwright directly. Service workers are blocked suite-wide in `playwright.config.ts`, since their console noise is not the application's.

An allowlist is provided for the errors we knowingly tolerate, so the assertion stays honest rather than being switched off.

`doc/TESTING.md` documents the fixture and the import convention.

### DIFF
--- a/doc/TESTING.md
+++ b/doc/TESTING.md
@@ -124,6 +124,27 @@ Coverage reports are written to the `coverage/` directory. This directory is git
 
 End-to-end tests run in a real browser using Playwright. They live in `e2e/` and follow the naming convention `*.spec.ts`.
 
+### Page-error capture
+
+Every spec imports `test` and `expect` from `e2e/helpers/page-errors.ts`, not
+from `@playwright/test` (enforced by a scoped ESLint `no-restricted-imports`
+rule). That module wraps `test` with an automatic per-test fixture that
+collects uncaught page exceptions and unhandled rejections (`pageerror`),
+console errors, and `/audio/` requests that fail, return >=400, or whose 2xx
+response lacks an `audio/*` content-type (SPA hosting serves a missing file as
+200 text/html, so content-type, not status, is the reliable signal), and fails
+the test at
+teardown if any arrived. Playwright's default semantics ignore all three
+channels, so without the fixture a test can pass while the page under it is
+broken. `e2e/page-errors.spec.ts` is the fixture's self-test: each positive
+case injects one error class and polls the capture array, and a final
+`test.fail()`-annotated case pins that a non-empty capture fails at teardown.
+If a spec ever legitimately produces a captured message, add a narrow,
+commented entry to the module's `ALLOWLIST` (the current entries cover
+third-party script hosts that are absent or DNS-blocked in test environments)
+rather than reverting the import. Service workers are blocked suite-wide
+(`playwright.config.ts`) so page-level network events stay observable.
+
 ### E2E Prerequisites
 
 The production build must exist before e2e tests run:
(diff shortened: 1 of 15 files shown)


wainwmr (changes requested):
Fails CI build

wainwmr (approved):
Nice

PR #812, opened 2026-07-13
fix: toggle play/pause on NumpadEnter

### LINKED ISSUE #711: Numpad Enter does not toggle play/pause
## 2. Description

Pressing the Enter key on the numeric keypad does not start or stop playback, while the Enter key on the main keyboard does. A user with a full-size keyboard expects both keys to behave the same way.

## 3. Reachability

1. Load the app in any browser.
2. Press the Enter key on the numeric keypad (the key whose `KeyboardEvent.code` is `NumpadEnter`).

Playback does not toggle. This is a real user-facing defect: the numpad Enter is a standard key on full-size keyboards and OS-level accessibility setups that remap Enter to the numpad.

## 4. Current state

`packages/pwa/index.ts`, `keyboardTapped` (lines 313-320):

```typescript
if (e.code == "Enter") {
  controls.isPlaying() ? controls.pause() : controls.play();
  return;
}
if (e.code == "Space") {
  controls.isPlaying() ? controls.pause() : controls.play();
  return;
}
```

Only `Enter` and `Space` are recognised as play/pause toggles. The DOM `KeyboardEvent.code` for the numpad Enter key is `NumpadEnter`, so it falls through to the switch statement and does nothing.

## 5. Root cause

The handler tests `e.code == "Enter"` literally and omits the numpad variant. Both keys produce the same `e.key` value (`"Enter"`) but different `e.code` values, and the app uses `code` for shortcut dispatching.

## 6. Scope

**In scope:**
- Recognise `NumpadEnter` as a play/pause toggle in `keyboardTapped`.
- Merge the duplicate `Enter` and `Space` branches while touching the same code (RR-fold item 11 from `refactor-index.ts.md`).
- Add a unit test for `NumpadEnter` toggling playback.

**Out of scope:**
- The bar input allowed-key list in `packages/pwa/src/ts/MusicControls.ts:143` also lists `Enter` but not `NumpadEnter`. That affects input-field behaviour, not the global play/pause shortcut reported here; it is left for a separate ticket if it surfaces as a user issue.

## 7. Recommended fix

In `packages/pwa/index.ts`, replace the two `if` blocks with one:

```typescript
if (e.code === "Enter" || e.code === "NumpadEnter" || e.code === "Space") {
  controls.isPlaying() ? controls.pause() : controls.play();
  return;
}
```

Use strict equality (`===`) for the new condition to match the rest of the handler.

## 8. Test plan

Extend `packages/pwa/src/test/keyboard.test.ts`:

1. Add a test mirroring the existing Space test: with focus on `document.body`, dispatch a `keydown` event whose `code` is `NumpadEnter`; assert `controls.isPlaying()` toggles from `false` to `true`, and a second press toggles back to `false`.
2. Add a test mirroring the existing "held Enter does not toggle play/pause" test: a `NumpadEnter` event with `repeat: true` leaves `controls.isPlaying()` unchanged.

Browser verification: the bug is purely a missing `KeyboardEvent.code` branch; jsdom dispatches events by `code` exactly as the production handler reads it, so the unit test is the authoritative regression guard. No Playwright case is required beyond a one-time manual check with a physical numpad Enter key.

## 9. Dependencies

None.

## 10. Open questions

None.

## 11. Follow-up items

- Folded: `refactor-index.ts.md` item 11 (merge identical `Enter` and `Space` branches) is folded into this change.
- Deferred: `MusicControls.ts` bar input allowed-key list does not include `NumpadEnter`; if that proves to be a user issue, it can be ticketed separately.

## 12. Documentation impact

None expected.

- Kimi

wainwright1000:
Fixes #711

The Enter key on the numeric keypad did not toggle play/pause. The main Enter key and the space bar did.

The keyboard handler matches on the event's `code`, and the keypad's Enter reports `NumpadEnter`, not `Enter`, so it fell through unhandled. Someone driving playback from the keypad had no way to start or stop it from there.

`NumpadEnter` now joins `Enter` and `Space` in the play/pause branch, and the two duplicated branches collapse into one. The keyboard test covers the keypad path.

### DIFF
--- a/packages/pwa/index.ts
+++ b/packages/pwa/index.ts
@@ -325,11 +325,7 @@ function keyboardTapped(e: KeyboardEvent) {
     return;
   }
 
-  if (e.code == "Enter") {
-    controls.isPlaying() ? controls.pause() : controls.play();
-    return;
-  }
-  if (e.code == "Space") {
+  if (e.code === "Enter" || e.code === "NumpadEnter" || e.code === "Space") {
     controls.isPlaying() ? controls.pause() : controls.play();
     return;
   }
--- a/packages/pwa/package.json
+++ b/packages/pwa/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@spem/pwa",
   "private": true,
-  "version": "2.8.11",
+  "version": "2.8.12",
   "description": "Helping singers learn Thomas Tallis Spem in alium",
   "type": "module",
   "dependencies": {},
--- a/packages/pwa/src/test/keyboard.test.ts
+++ b/packages/pwa/src/test/keyboard.test.ts
@@ -31,6 +31,23 @@ describe("keyboard shortcuts", () => {
     expect(controls.isPlaying()).toBe(false);
   });
 
+  it("NumpadEnter toggles play/pause when focus is on the document body (#711)", async () => {
+    const controls = document.querySelector("music-controls") as MusicControls;
+    controls.playing = false;
+
+    document.body.dispatchEvent(
+      new KeyboardEvent("keydown", { code: "NumpadEnter", bubbles: true })
+    );
+    await new Promise((resolve) => setTimeout(resolve, 0));
+    expect(controls.isPlaying()).toBe(true);
+
+    document.body.dispatchEvent(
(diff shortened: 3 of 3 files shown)


wainwmr (approved):
OK

PR #814, opened 2026-07-13
ci: guard the app version on every pull request

### LINKED ISSUE #810: CI does not guard the app version: two builds can ship under one version string
**Type:** tooling bug
**Area:** Tooling
**Status:** Todo
**Difficulty:** S

**Source:** `.github/workflows/` (no version check exists);
`packages/pwa/src/test/config.test.ts:34` (the only version assertion, and it
cannot catch this).

**First seen:** 2026-07-13, commit `7ad3a8b` (PR #804), which shipped 2.8.10 after
`7b640dc` (PR #805) had already shipped 2.8.10.

## 2. Description

Two different builds shipped to production under the same version string, and
nothing noticed.

The version bump is computed against `origin/main` alone, so two PRs in flight both
read main, both compute the same next number, and both claim it:

- `7b640dc` (PR #805, the audio-load-window fix) bumped
  `packages/pwa/package.json` to **2.8.10** and merged.
- `7ad3a8b` (PR #804, the LilyPond build-time precompute) merged 12 minutes later,
  still declaring **2.8.10**.

Main stayed at 2.8.10, so two distinct builds are released under one version. The
version is injected into `index.html` at build time (`doc/CONTRIBUTING.md`, Key
Conventions), so a user's reported version no longer identifies the build they are
running.

**Nothing catches it.** There is no CI check on the version. The only test that
reads it (`config.test.ts:34`) asserts the string merely *matches* a semver shape
(`/^\d+\.\d+\.\d+/`), which a duplicate passes trivially.

The authoring rule has since been repaired locally (the next version is now computed
against a floor of main plus every other open PR), but that rule is prose and prose
can be skipped. This ticket adds the guard that cannot be.

### Why this must be in CI, not in the authoring tooling

This is the design premise, and the check is unsound without it.

At the moment PR #804 was published it was **correct**: it bumped to 2.8.10 while
main was still 2.8.9. It became wrong only later, when #805 merged first and took
that number. Nothing on the authoring side is looking at the branch at that moment,
so no amount of care at publish time can catch it.

CI is what looks again. Mark merges `main` into a branch before merging the branch
(he did exactly that on #804), which fires the `pull_request` `synchronize` event
and re-runs the checks against the refreshed branch. That re-run is where the
collision becomes visible, and it is the only place it can be caught reliably.

## 3. Recommended fix

A pull-request check that enforces **both halves** of the version rule. A check that
enforces only "the version must be bumped" would fail correct PRs.

**The rule:**

1. A PR whose commits include a `feat:`, a `fix:`, or a **user-facing** `perf:` must
   carry a version **strictly above** main's, and **equal to no other open PR's**.
2. A PR whose commits are internal only (`build:` / `ci:` / `docs:` / `refactor:` /
   `test:`, or a build-only `perf:`) must carry **no bump**: its version equals
   main's.

**Worked counter-example, currently open.** PR #808 (`test: fail e2e tests on
captured page errors`) is a test-only change sitting at 2.8.10, the same as main.
That is **correct** under rule 2, and any implementation must pass it unchanged. A
naive "version must be greater than main's" check fails it.

**Shape.** Put the decision in a **pure function** that takes the PR's commit
subjects, the PR's version, main's version, and the other open PRs' versions, and
returns pass/fail plus a reason. Unit-test that function; keep the workflow a thin
caller. A check that can only be tested by running CI is a check nobody trusts.

**Files:**

- A new job (or a new small workflow) triggered on `pull_request`, default types, so
  `synchronize` re-runs it when main is merged into the branch. `pwa-ci.yml` already
  reads `package.json` and is the natural host; a separate workflow is also fine.
- A script under `.github/scripts/` holding the pure decision function, with tests.
- Reading the other open PRs needs `GITHUB_TOKEN` with `pull-requests: read`.

**Risks, and the one decision the implementer must not duck:**

- **Fail closed when the open-PR list cannot be read.** A guard that fails open
  disappears silently on a GitHub hiccup, which is the exact failure class this
  ticket exists to remove. Fail, with a message that distinguishes "could not read
  the open PRs" from "your version collides with PR #NNN", so it is re-runnable and
  an outage is never mistaken for a violation.
- Classifying a `perf:` commit as user-facing or build-only cannot be done from the
  commit type alone. Keep it simple: treat `perf:` as user-facing (patch bump owed)
  unless its scope marks it as build-side (for example `perf(scores):`, the score
  build). Both live examples fit: #804 (`perf:`, a cold-load stall removed) bumped;
  #803 (`perf(scores):`) did not.
- The Review queue is WIP-capped at two or three, so the "every other open PR" read
  is a couple of API calls, not a sweep.

## 4. Test plan

Unit tests against the pure decision function:

1. A `fix:` PR at main's version **fails** (no bump when one is owed).
2. A `fix:` PR whose version equals another open PR's **fails** (the collision that
   shipped today; use #804 and #805's real versions as the fixture).
3. A `fix:` PR strictly above main and unique among open PRs **passes**.
4. A `test:`-only PR at main's version **passes**. This is the #808 regression, and
   it is the case that stops the check failing good PRs.
5. A `test:`-only PR carrying a bump **fails** (bumped when none was owed).
6. A user-facing `perf:` PR is treated as owing a patch bump; a build-scoped
   `perf(scores):` PR is treated as internal.
7. An unreadable open-PR list **fails** with the distinct "could not read" reason,
   not with a collision reason.

Live proof: the check must pass against PR #808 as it stands, and must fail against
a reconstruction of #804 at 2.8.10 with #805 open at 2.8.10.

## 5. Dependencies

None. The authoring-side rule is already repaired in the local publish method. This
is the independent guard and it stands alone.

## Provenance

Andrew decided on 2026-07-13 that the mechanism is a CI check and that it is wanted.
The design call is settled and does not need re-surfacing.

- CLAUDE

wainwright1000:
Fixes #810

On 2026-07-13, PR #804 and PR #805 each computed their version against `main` alone. Both claimed 2.8.10, and the second to merge shipped a different build under an already-published version. Nothing caught it: there is no CI check and no test on the version, and the only test that reads it asserts the string merely *matches* a semver shape, which a duplicate passes.

This adds a CI guard so that cannot recur.

## One thing I need from you

**The check does not block a merge yet, and it cannot until you change a repository setting.**

This adds a status check named `version`. The "Main should be golden" ruleset currently requires exactly one check, `test`:

```console
$ gh api repos/wainwmr/spem-player/rulesets/15878992 \
    --jq '[.rules[] | select(.type=="required_status_checks") | .parameters.required_status_checks[].context]'
["test"]
```

So a red `version` check annotates a PR and the merge button stays green, and `dependabot-auto-merge.yml` (which merges as soon as the *required* checks pass) cannot see it at all. Had this branch been in place on 2026-07-13, it would have turned PR #805 red and #805 would have merged anyway.

**Please add `version` to that ruleset's required status checks.** It is safe to require: the workflow has no path filter, so it runs on every PR to `main` and can never become a permanently-pending required check (which is the reason `test-noop.yml` exists for `test`). `doc/CI.md` states the gap explicitly, so the code is honest about it until you do.

## The rule

A version bump is owed when **both** signals fire:

- the PR changes **app source** (something that can alter what a user downloads), **and**
- at least one commit subject is not plainly internal.

Such a PR must carry a version strictly above main's and equal to no other open PR's. Any other PR must carry **no** bump. It may sit *below* main, which happens whenever main moves under a branch that correctly did not bump; #808 was exactly that shape, so requiring equality would have failed it.

**App source** is everything under `packages/pwa/` (bar the version file, tests, e2e and the tooling configs, enumerated by exact path), plus `packages/scores/src/` (which `vite.config.ts` aliases as `@scores` and bundles) and `pnpm-lock.yaml` (the PWA declares `"dependencies": {}`, so `vite`, `vite-plugin-pwa`, `sass-embedded` and `workbox-window` are all devDependencies whose output *is* the bundle).

**Why the path, and not the commit type alone.** The type signal on its own does not work in this repo, and that is measured rather than asserted. A type-based classifier, replayed against `main`, demanded a bump on 11 of 21 user-facing-typed commits that correctly shipped without one, because the common real shape here is an unscoped `fix:` on the monitor package, on lint config, or on a build script. A guard that reds a third of legitimate PRs gets switched off, and then the bug comes back.

**Which subjects it reads.** The repo squash-merges with `squash_merge_commit_title: COMMIT_OR_PR_TITLE`, so the landed subject is the branch commit's on a single-commit PR and the PR title otherwise. Either can be what ships, so the guard reads both. That is also why `edited` is in the trigger list: a retitle fires `edited`, not `synchronize`, so without it an author could pass green as `chore:` and retitle to `feat:` before merging.

## Evidence

Replayed against **all 128 commits merged since `packages/pwa/package.json` was created**: it agrees with what actually happened **122 times, with zero false greens**, and reds 6.

Five of the six are the guard working, including **PR #804 itself**. Three others (`8467e3b`, `8acb303`, `d0127fc`) changed app source and shipped under an unchanged version, so this has happened more than once.

The sixth is a **false red**, named here rather than buried: `f2207fb fix: restore ESLint coverage` touches only `pnpm-lock.yaml` and ships nothing to a user. It is red because the lockfile counts as app source under a `fix:` subject. That is the measured cost of closing the dependency hole, one in 128, and the remedy is honest typing (a lint-dependency change is `build:`, not `fix:`), never a version burn.

## Two known residuals, stated rather than hidden

Both are deliberate. Closing either is a rule change rather than a bug fix, so I have not made it unilaterally.

1. **An internally-typed change to app source owes no bump.** Measured: 12 of the 122 passing commits changed app source, shipped no bump and were green, and they are ordinary honest `refactor:` PRs. Roughly one in nine. Closing it means "app source changed implies a bump", which would have flagged 18 of 128.
2. **A Dependabot dev-dependency bump owes no bump**, even though `vite` is a devDependency whose output is the bundle. Without the exemption every Dependabot PR goes red, and Dependabot can neither bump the app version nor clear the check, and its PRs auto-merge.

Both are documented in `doc/CI.md` and in the module header.

## It fails closed

A guard that passes when it cannot do its job is worse than no guard, because it manufactures the confidence it was built to earn. It reports a failure, never a silent pass, when: either version is not a plain `x.y.z`; the open-PR list cannot be read, is not a list, or comes back full and so may be truncated; the list comes back **without this PR in it** (the PR under test is itself an open PR, so its absence proves the list cannot be trusted, and an empty list would otherwise read as "nothing to collide with"); a peer's version will not parse; or git cannot be read. A peer whose `packages/pwa/package.json` 404s is skipped only on proof that its ref resolves, and a **fork** peer is never skipped, because a dropped peer could hide a collision.

Exit `1` means the version is wrong. Exit `2` means the guard could not tell. A crash is not a version violation.

## Structure and cost

`decideVersion` is a pure function with **62 unit tests**, run by `push-helper-test.yml`. `version-check-cli.mjs` is a separate entry point so the module can be imported by its tests without asking "was I invoked directly?", which would be a fail-open guard on a fail-closed tool.

The job checks out with `filter: blob:none` and `sparse-checkout: .github`, because it reads a 1,905-byte `package.json` while the tip tree is 791.5 MiB, of which 775.8 MiB is choir audio. The two options are a pair: dropping the sparse checkout while keeping the filter re-introduces the full download. `fetch-depth: 0` is kept and required, because the commit walk and the diff both need the merge base.

This branch passes its own guard.

- VERA

### DIFF
--- a/.github/scripts/version-check-cli.mjs
+++ b/.github/scripts/version-check-cli.mjs
@@ -1,0 +1,28 @@
+#!/usr/bin/env node
+/**
+ * The entry point CI runs for the app-version guard (ticket #810).
+ *
+ * This file exists so `version-check.mjs` can be imported by its tests without
+ * running, WITHOUT the module needing to ask "was I invoked directly?". That question
+ * used to be answered by comparing `process.argv[1]` against `import.meta.url`
+ * (realpath'd on both sides, after a symlinked-checkout bug). It was a fail-OPEN guard
+ * on a fail-CLOSED tool: had the comparison ever returned false in CI, `main()` would
+ * simply not have run, the process would have exited 0, and the check would have been a
+ * silent no-op on every PR forever. A separate entry file cannot fail that way. There
+ * is no condition to get wrong.
+ */
+import { main } from "./version-check.mjs";
+
+let code;
+try {
+  code = main();
+} catch (e) {
+  // An uncaught throw would exit 1, and 1 is reserved for "the version is wrong". A
+  // crash is not a version violation; report it as one the contributor cannot fix.
+  console.error(`version-check: crashed: ${e?.stack || e}`);
+  code = 2;
+}
+// `process.exitCode`, never `process.exit()`: stdout to a pipe (which is what an
+// Actions step gives us) is async, and `process.exit()` terminates without draining it,
+// so the verdict we just printed could be thrown away and the guard would blame nothing.
+process.exitCode = code;
--- a/.github/scripts/version-check.mjs
+++ b/.github/scripts/version-check.mjs
@@ -1,0 +1,587 @@
+/**
+ * CI guard for the app version (ticket #810).
+ *
+ * Two builds must never ship under one version string. On 2026-07-13 PR #804 and
+ * PR #805 each computed their version against `main` alone, both claimed 2.8.10,
+ * and the second to merge shipped a different build under an already-published
+ * version. That is the failure this guard exists to prevent, and the reason to
+ * resist relaxing the rules below.
+ *
+ * THE RULE
+ *
+ *   1. A PR that changes APP SOURCE with a user-facing intent must carry a version
+ *      strictly above main's, and equal to no other open PR's.
+ *   2. Any other PR must carry no bump: its version must not be above main's. (It
+ *      may sit BELOW main, which happens whenever main moves under a branch that
+ *      correctly did not bump. Requiring equality would fail such a PR; #808 was
+ *      exactly that shape.)
+ *
+ * WHY THE PATH, AND NOT THE COMMIT TYPE ALONE
+ *
+ * The first cut of this guard classified on the commit type alone (with a list of
+ * "internal" scopes). Replayed against the real history of `main`, it demanded a
+ * bump on 11 of 21 user-facing-typed commits that correctly shipped without one,
+ * because the dominant real shape is an UNSCOPED `fix:` on the monitor package, on
+ * lint config, or on a build script. Only 14 of 129 user-facing commits carry any
+ * scope at all, so scope could never have carried this weight.
+ *
+ * Worse, it red-flagged its own review process: the Vera gate appends
+ * `fix: ... (Vera NNN-NN)` commits to the branch it reviews, so any CI-only PR that
+ * went through a gate would have been told to bump the app version. The escape
+ * route (bump to go green) burns a version number a real in-flight PR is racing
+ * for. That is a self-inflicted #810.
+ *
+ * The signal the history actually carries is the PATH: every commit that bumped touched
+ * app source; none of the false reds did. So we ask that question directly, and the scope
+ * list is gone. (Andrew's ruling, 2026-07-13, on the Vera gate for this branch.)
+ *
+ * Replayed against all 128 commits merged since `packages/pwa/package.json` was created,
+ * the path rule agrees with what actually happened 122 times, with ZERO false greens in
+ * either direction, and reds 6.
+ *
+ * FIVE of those reds are the guard working: #804 itself (`7ad3a8b`), and four commits
+ * (`8467e3b`, `8acb303`, `d0127fc`, `73dbee2`) that changed app source and shipped under an
+ * unchanged version.
+ *
+ * The SIXTH is a FALSE red, and it is named here rather than buried: `f2207fb fix: restore
+ * ESLint coverage for packages/monitor .mjs files` touches only `pnpm-lock.yaml`, ships
+ * nothing to a user, and is red solely because the lockfile counts as app source (see
+ * `isAppSource`) under a `fix:` subject. That is the measured cost of closing the
+ * dependency hole: one in 128. The remedy is honest typing, not a version burn: a
+ * lint-dependency change is `build:` or `tooling:`, not `fix:`. Do NOT clear such a red by
+ * bumping the version, which burns a number a real in-flight PR is racing for.
+ *
+ * The replay is on the Vera report attached to #810.
+ *
+ * WHAT SHIPS
+ *
+ * The version guards `packages/pwa/package.json`, which vite injects into
+ * `index.html` at build time. `packages/scores/src` is app source too: `vite.config.ts`
+ * aliases `@scores` to it and `lilyData.ts` and `config.ts` import from it, so it is
+ * bundled. `packages/scores/build` is NOT: it is the LilyPond pipeline that produces
+ * the assets, and it does not ship.
+ *
+ * WHICH SUBJECTS
+ *
+ * The repo squash-merges with `squash_merge_commit_title: COMMIT_OR_PR_TITLE`, which
+ * means the landed subject is the branch COMMIT's when the PR has exactly one commit,
+ * and the PR TITLE otherwise. EITHER can be the thing that ships, so the guard reads
+ * the union: the title AND the branch commits, with a user-facing intent in any one of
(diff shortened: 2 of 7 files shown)


wainwright1000:
Mark, this PR needs one repo setting from you before it can actually protect anything, and only you can make the change.

**The problem.** The new check reports as `version`, but the `Main should be golden` ruleset currently requires only `test`. So a **red `version` check still leaves the merge button green**, and Dependabot's auto-merge cannot see it at all. Had this guard been live on 13 July it would have turned #805 red, and #805 would have merged anyway.

**The ordering matters.** Please **merge this PR first**, and only then make the check required. If `version` becomes a required check while the workflow exists only on this branch, every other open PR will block forever, waiting for a check that never reports.

**Then, to make it required:**

1. **Settings** → **Rules** → **Rulesets** → **Main should be golden**
2. Scroll to **Require status checks to pass**
3. **Add checks**, type `version`, and select it (GitHub will suggest it, because it has now run for real on this PR)
4. **Save changes**

The check's exact name is `version`. Afterwards the ruleset should list both `test` and `version`.

**Why it is worth the setting.** The guard exists because the app has shipped more than one build under a single version string. #804 and #805 both claimed 2.8.10, and a replay across all 128 commits since `packages/pwa/package.json` was created found three further merged commits that changed app source under an unchanged version. The version is injected into `index.html` at build time, so the collision is visible in the released artefact.

The check has been replayed over that full history: 122 of 128 commits agree with it, with **zero false greens**. The six reds are five genuine defects (one of them #804 itself) and one named false red, documented in the PR body.

- TELE

wainwmr (approved):
blimey

PR #815, opened 2026-07-13
fix: stop a Telegram failure suppressing dispatchAlert's tracking issue

### LINKED ISSUE #748: Fix dispatchAlert: Telegram failure suppresses the tracking issue
**Symptom.** When the monitor dispatches a failure alert while Telegram is down or misconfigured, the GitHub tracking issue is never opened: `dispatchAlert` runs `sendTelegram` and `openIssue` in ONE try block, so the Telegram throw skips the issue leg and the whole alert degrades to a single `console.error` line in a CI log nobody reads. The durable record is lost exactly when an incident is in progress.

**Steps to reproduce.**

1. Make Telegram fail (invalid `TELEGRAM_BOT_TOKEN`, or the fork's live HTTP 404 state, see #745).
2. Make a primary API fail so `main()` takes an alert branch (e.g. Netlify 500).
3. Observe: no GitHub issue is opened; the only trace is `Failed to send alert for Netlify failure: Telegram HTTP 404` in the workflow log.

Unit-level: stub fetch so `/sendMessage` returns 500 and `/issues` returns 201; call `dispatchAlert({telegramText: "x", issueTitle: "T", issueBody: "B", onErrorLog: "p"})`; the `/issues` route is never called.

**Expected behaviour.** Each leg is attempted independently: a Telegram failure is logged and swallowed, and `openIssue` still runs (the issue is the durable channel; the existing issue-only mode shows that intent). The function's own doc ("A failure in either is logged... and swallowed") already reads as promising this.

**Actual behaviour.** A Telegram-leg throw aborts the try before `openIssue`.

**Affected files.** `packages/monitor/monitor-resources.mjs:551-563` (`dispatchAlert`). Test: `packages/monitor/monitor-resources.test.mjs` (the swallow test at ~904 fails only the issue leg; no test fails the Telegram leg).

**Root cause hypothesis.** The two legs were wrapped in one try when `dispatchAlert` was extracted (#679); the coupling was invisible because the happy path and the issue-leg-failure path both behave correctly.

**Test code.**

```js
test("dispatchAlert still opens the issue when Telegram fails", async () => {
  const m = startMock([
    ["/sendMessage", () => res("down", { ok: false, status: 500 })],
    ["/issues", () => res({ id: 1 })],
  ]);
  try {
    await dispatchAlert({ telegramText: "x", issueTitle: "T", issueBody: "B", onErrorLog: "p" });
    assert.ok(m.calls.some((c) => /\/issues$/.test(c.url))); // FAILS today
  } finally { m.restore(); }
});
```

**Acceptance criteria.** With Telegram failing and GitHub up, `dispatchAlert` opens the issue and logs the Telegram failure; with both failing, both failures are logged; existing swallow behaviour (never masking the caller's in-flight error) is preserved.

**Risk and scope.** Four-line change (two try blocks) plus one test. No behaviour change on the happy path.

**Notes.** Finding F8 of the 2026-07-04 bug-hunt (reviewer: silent-failure-hunter). Live corroboration: #659/#660 were opened by the CHART alert branch during the fork's Telegram-404 spell only because that branch is issue-only; the three telegramText-bearing branches would have lost their issues under the same conditions. Related observation for specify: `openIssue` has no dedup (#659/#660 are identical), out of scope here unless trivial. Related: #745.

- Vera

wainwright1000:
Fixes #748

`dispatchAlert` does two things when the monitor detects a problem: it sends a Telegram message, and it files a tracking issue. They ran in sequence, so a Telegram failure threw before the issue was ever created, and the problem the alert existed to report went unrecorded. The louder channel silenced the durable one.

The two are now decoupled: each is attempted independently, and a failure in one is logged without preventing the other. The tracking issue is filed whether or not the Telegram send succeeds.

`packages/monitor/monitor-resources.test.mjs` covers the case that mattered: a failing Telegram send, and the issue still filed.

### DIFF
--- a/packages/monitor/monitor-resources.mjs
+++ b/packages/monitor/monitor-resources.mjs
@@ -536,9 +536,11 @@ export async function openIssue(title, body) {
 
 /**
  * Dispatch a monitor alert: optionally send a Telegram message, then open a
- * tracking issue. A failure in either is logged with `onErrorLog` and
- * swallowed, so a secondary alert failure never masks the caller's own in-flight
- * error.
+ * tracking issue. The two legs are attempted independently: a failure in
+ * either is logged with `onErrorLog` (tagged with the failing leg) and
+ * swallowed, so a Telegram outage
+ * cannot suppress the durable tracking issue (#748) and a secondary alert
+ * failure never masks the caller's own in-flight error.
  *
  * @param {object} alert
  * @param {string} [alert.telegramText] - Telegram message. Omit it (or pass any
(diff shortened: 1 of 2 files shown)


wainwmr (approved):
OK

Issue #817, opened 2026-07-13
Declare canvas as a @spem/pwa devDependency (test 2D backend)


wainwright1000:
## 1. Original title

> @spem/pwa depends on node-canvas for its tests but does not declare it (it resolves via jsdom's optional peer, satisfied by packages/monitor)

## 2. Description

`@spem/pwa`'s unit tests render through the real node-canvas 2D backend, but `@spem/pwa` does not declare `canvas` as a dependency. It works today only as a side effect of an unrelated package. Verified against source:

- `canvas` is declared only in `packages/monitor/package.json`, as a devDependency (chart rendering).
- `jsdom` declares `canvas` as an optional peer dependency.
- pnpm satisfies that optional peer from the single workspace copy `packages/monitor` provides; the lockfile records `jsdom@29.1.1(canvas@3.2.3)` with `optionalDependencies: canvas: 3.2.3`.

If `packages/monitor` ever drops `canvas` (a realistic tidy-up, it is only a chart-rendering devDependency there), jsdom's optional peer goes unsatisfied, `getContext("2d")` returns `null`, and every canvas test dies on a non-null assertion. The failure is loud, but the diagnosis is non-obvious: a maintainer would look at `packages/pwa` and jsdom, not at a deleted line in a sibling package.

## 4. Current state

- `packages/pwa/package.json` (devDependencies, lines 8-31): no `canvas` entry.
- `packages/monitor/package.json:13`: `"canvas": "^3.2.0"`.
- `pnpm-lock.yaml`: `canvas` appears under the `packages/monitor` importer only; `jsdom@29.1.1(canvas@3.2.3)` carries `optionalDependencies: canvas: 3.2.3`.
- `packages/pwa/src/test/canvas.test.ts` (lines 106, 158, 203, 259, 275, among others): `canvas!.canvas!.getContext("2d")!` non-null assertions against the real backend.
- `packages/pwa/knip.json:5`: `"ignoreDependencies": ["autoprefixer", "lint-staged"]`.

## 5. Root cause

jsdom declares `canvas` as an optional peer, and pnpm satisfies it from whichever workspace package provides a copy. `packages/pwa` never declares `canvas` itself, so the peer satisfaction its tests rely on is incidental to an unrelated package's devDependency.

## 6. Scope

In scope: declare `canvas` in `packages/pwa/package.json`; add the matching `knip` ignore entry; regenerate `pnpm-lock.yaml`.

[...]

Issue #816, opened 2026-07-13
Light-mode false-relation hotspot: the selected part is near-invisible, the unselected one is dark (the % 100 wrap inverts the emphasis)


wainwright1000:
## 2. Description

In **light mode**, the false-relation hotspot for the **selected** voice part is almost invisible, while the hotspot for an **unselected** part is dark and obvious. The emphasis is inverted: the part the user has selected is the one they cannot see. Dark mode is correct, which is why this went unnoticed.

## 3. Reachability

Light mode is a first-class, user-toggleable theme in the shipped app (`light-theme` body class; light-mode background `hsl(210, 65%, 100%)`, dark-mode `hsl(210, 65%, 20%)`, `packages/pwa/src/scss/style.scss:12,43`). Any user in light mode who selects a choir/part containing a false relation hits this on the normal play-and-practise path; false relations are a permanent fixture of the canvas in both themes. This is canvas HSL arithmetic, not a computed-CSS property, so no real-browser computed-style check is required to establish it (the what-is-a-bug evidence caveat does not apply).

## 4. Current state

`packages/pwa/src/ts/MusicCanvas.ts` `#getHotspotLightness` (`:675-686`), sole caller `#drawFalseRelationHotspot` (`:621`):

```ts
const baseLight = isSelected
  ? MusicCanvas.SELECTED_BASE_LIGHTNESS - 3 * p   // 45 - 3p
  : dullBase - 3 * p;                             // light: 80 - 3p, dark: 38 - 3p
return (baseLight + 50) % 100;
```

Worked through (`p` is the part index, 0 to 4), with prominence as distance from the page-background lightness:

- **Light mode, selected:** `(45 - 3p + 50) % 100` = 95, 92, 89, 86, 83. Distance from white: 5 to 17. Near-invisible.
- **Light mode, unselected:** `(80 - 3p + 50) % 100` = 130 to 118, wrapped by `% 100` to 30, 27, 24, 21, 18. Distance from white: 70 to 82. Dark and obvious.
- **Dark mode, selected:** 95 down to 83. Distance from 20: 75 to 63. Correct.
- **Dark mode, unselected:** 88 down to 76. Distance from 20: 68 to 56. Correct.

The wrap only ever bites the light-mode unselected case, and it is what inverts the emphasis.

## 5. Root cause

[...]

PR #819, opened 2026-07-13
test: pin MusicCanvas rendering/pulse/colour behaviours (#706)

### LINKED ISSUE #706: Pin MusicCanvas rendering/pulse/colour behaviours with assertions (covered-but-unasserted)
## Description

The `MusicCanvas` overview's rendering and pulse/quantisation maths is covered-but-unasserted: a ~250-line band of `packages/pwa/src/ts/MusicCanvas.ts` runs under `draw()` calls in `packages/pwa/src/test/canvas.test.ts` but no test checks the computed output. A regression in colour, geometry, quantisation, or theme constants still renders a frame and throws nothing, so the suite stays green. This is the blind spot that let the 16th-note pulse-drop in #704 ship undetected.

## Current state

- `packages/pwa/src/ts/MusicCanvas.ts` (760 lines) renders the Canvas 2D overview.
- `draw()` calls the private helpers listed below; all inputs are public fields and all outputs are observable through public arrays or a spied 2D context.
- `packages/pwa/src/test/canvas.test.ts` (1473 lines) already exercises `draw()` in several smoke tests, but most assert only that no exception is thrown.

Key helper locations at `origin/main` (`f141869`):

- `#updatePulses` — lines 388-434. Quantises `this.bar`, looks up `notesByQuant`, updates `pulses[c][p]` with an ease-out-cubic envelope, and updates `falseRelationPulses[i]` with a `sqrt(1 - t)` envelope.
- `#drawBarHighlight` — lines 441-458. Guards with `this.bar <= 0 || this.bar > this.barCount`, then strokes a vertical playhead line.
- `#drawSelectionHighlight` — lines 460-490. Computes `startY` and `width` differently when `voicePart == "all"` vs a numeric part.
- `#drawVoiceParts` — lines 492-544. Four-branch colour cascade that sets `saturation`, `lightness`, and `transparency` before each `stroke()`.
- `#getHotspotLightness` — lines 650-661. Returns `(baseLight + 50) % 100` for light/dark and selected/unselected cases.
- `#getMousePos` — lines 663-685. Derives `part` from the fractional part of the normalised Y coordinate.

## Root cause

The existing `draw()` tests were written as smoke tests: they verify the element initialises and `draw()` can be called without throwing. Because Canvas 2D output is pixel-based, earlier tests avoided brittle pixel comparisons; the project now has enough `ctx` spy patterns (see `#554` and `#649` tests) to assert on colour strings, geometry, and public arrays instead.

## Scope

In scope:

- Add focused assertions in `packages/pwa/src/test/canvas.test.ts` for the behaviours listed below.
- Use public fields and `CanvasRenderingContext2D` spies; do not change `MusicCanvas.ts` visibility or signatures.

Explicitly deferred:

- The `#updatePulses` quant line `Math.floor(this.bar * 16) / 16` is the subject of #704 and is folded into that ticket's TDD fix. #706 asserts the easing envelope and the `falseRelationPulses` envelope, but not the 1/16 vs 1/128 grid mismatch.
- General test-file refactor items from `wiki/refactor-canvas.test.ts.md` (helper extraction, `var` cleanup, setTimeout removal, seek precision, touch precision) are not in scope unless a listed behaviour directly touches them.

## Recommended fix

1. In `packages/pwa/src/test/canvas.test.ts`, add one `describe` block or a set of focused `it` blocks that pin the behaviours below.
2. For each test, set the public input fields, call `draw()`, and assert the observable output. Show each new assertion failing when its behaviour is regressed (TDD).
3. Use `vi.spyOn(ctx, "stroke")` / `fillRect` / `moveTo` / `lineTo` / `lineWidth` to capture geometry and colour strings where the public arrays do not expose the value.

## Test plan

1. **`#updatePulses` easing envelope.** Set `bar` so a known note is active, call `draw()`, and assert `pulses[c][p]` equals the value produced by `#easeOutCubic(elapsed, 1.6, -0.6, duration)` in dark mode and `(0.4, 0.6, duration)` in light mode. Toggle `document.body.classList.add("light-theme")` to cover the light-mode branch.

2. **False-relation pulse `sqrt(1 - t)` envelope.** Set `bar` to `loc.from + elapsed` for a populated `frLocations[i]` where `elapsed < FR_PULSE_FADE_BARS`, call `draw()`, and assert `falseRelationPulses[i]` equals `Math.sqrt(1 - elapsed / FR_PULSE_FADE_BARS)`.

3. **`#drawVoiceParts` four-branch colour state.** Create a fixture with a known range and vary `bar`, `choir`, and `voicePart` to hit each branch. Spy on `ctx.strokeStyle` and assert the captured HSL string matches the expected hue, saturation, lightness, and alpha for:
   - current bar inside the range;
   - current choir/part selected but bar outside the range;
   - `bar === 0` or `bar > barCount`;
   - unselected, bar inside piece.

4. **Light/dark mode constants.** Toggle `document.body.classList.contains("light-theme")` and assert that `#drawDev` text colour, `#drawVoiceParts` dull base lightness, and `#getHotspotLightness` output differ between modes. (The light-mode path is currently untested.)

5. **`#getHotspotLightness` wrap.** For selected and unselected positions in light and dark modes, assert the result equals `(baseLight + 50) % 100`.

6. **`#drawBarHighlight` guard boundaries.** At `bar = 0` and `bar = barCount + 1`, spy `ctx.stroke` and expect no call. At `bar = 1` and `bar = barCount`, expect `ctx.moveTo`/`lineTo` with the playhead x-coordinate `(bar + 0.5) * barWidth + canvasPadding`.

7. **`#drawSelectionHighlight` all-vs-single-part geometry.** With `voicePart = 2`, spy `ctx.lineWidth` and `ctx.moveTo`/`lineTo` y-coordinate and assert the narrow part highlight. With `voicePart = "all"`, assert the wider, choir-centred highlight.

8. **`#getMousePos` exact part derivation.** Create a `MouseEvent` with a known `offsetY` in a known `getBoundingClientRect` fixture and assert the returned `part` equals the expected index.

9. **Regression guard.** For each new assertion, temporarily introduce the regression it guards (e.g. change the easing constants, remove the guard, invert the wrap) and confirm the test fails.

## Dependencies

- #704 — the pulse-quant grid fix. The quantisation subset of `#updatePulses` folds into #704's TDD fix; #706 depends on #704 for that specific assertion.
- #676 — predecessor ticket that closed the uncovered-line gaps; #706 closes the covered-but-unasserted gaps.

## Open questions

None.

## Follow-up items

- Folded into this ticket: `wiki/refactor-canvas.test.ts.md` item 13 (covered-but-unasserted rendering/pulse maths) and item 15 (smoke `draw()` tests assert nothing).
- Deferred: other `refactor-canvas.test.ts.md` and `refactor-MusicCanvas.ts.md` items that are not adjacent to the listed behaviours.

## Documentation impact

None expected. `doc/TESTING.md` already covers the Vitest setup; no new build or workflow behaviour is introduced.

- Kimi

wainwright1000:
Pins the `MusicCanvas` rendering, pulse and colour behaviours that `draw()` exercises but no existing test asserted: the covered-but-unasserted band that let the #704 pulse-drop ship.

Fixes #706

## What is pinned

- The `#updatePulses` easing envelope, in both modes, at onset and mid-envelope, plus the backwards-seek guard and the note-ended boundary.
- The four-branch `#drawVoiceParts` colour state, including **both ends of the half-open singing range** and the `bar > barCount` disjunct.
- The false-relation `sqrt(1 - t)` pulse envelope and its fade window.
- `#getHotspotLightness`, `#drawBarHighlight`'s guard boundaries and playhead position, `#drawSelectionHighlight`'s geometry, and `#getMousePos`.

Test-only: no production code changes, so no version bump is owed.

## Verification

Every assertion was checked by mutation rather than by inspection. **Fifteen mutations of `MusicCanvas.ts`, all killed:** both easing coefficients, both easing bases, the backwards-seek guard, the strict `<` at duration 0, the `else` rest write, both ends of the half-open range, both dull-base lightness constants, the selected-base constant, the false-relation fade window, the `sqrt(1 - t)` envelope, and the hotspot `+ 50` brighten.

Two mutants are documented in the code as **equivalent** (unobservable by construction, not coverage gaps): `<` versus `<=` at `elapsed === duration`, because the ease lands on exactly 1.0 there in both modes; and the false-relation window end at `frLocations[0]`, for a floating-point reason recorded in the comment.

Rest-value assertions (`toBe(1)`, `toBe(0)`) are pre-poisoned with a NaN sentinel, because their expected value is also the value the field already holds, so without the sentinel a `draw()` that did nothing would satisfy them.

## Two defects found while writing these tests

**#816 (bug, Canvas).** In light mode the false-relation hotspot for the **selected** part lands at ~95% lightness (near-invisible on a light background) while an **unselected** one lands at ~30% (dark and obvious). The emphasis is inverted, because `(baseLight + 50) % 100` wraps only in light mode, where the dull base is 80 and `80 + 50 = 130` wraps to 30. Dark mode is correct, which is why it has gone unnoticed.

This PR **pins the current (buggy) behaviour deliberately**, with a WARNING comment naming #816 and stating that the test must be updated when the bug is fixed, and that this is the intended outcome rather than a regression. Pinning it silently would have cemented a visual defect and made it harder to notice later; deleting the assertion would have lost the coverage.

**#817 (tech debt, Tooling).** `@spem/pwa`'s tests render through the real node-canvas 2D backend but the package never declares it. It resolves only because jsdom declares `canvas` as an optional peer and `packages/monitor` happens to provide a copy.

## Review notes

This branch was parked, conflicted on rebase, and the conflict resolution rewrote how every test in the block installs its fixture (#652 collapsed `notesByQuant` / `ranges` / `frLocations` / `barCount` into a single readonly `lilyData` object). That resolution is new composite code, so the Vera reviewer panel was re-run over three passes; the full record, including the findings against my own work, is in the ticket comment.

### DIFF
--- a/packages/pwa/src/test/canvas.test.ts
+++ b/packages/pwa/src/test/canvas.test.ts
@@ -3,12 +3,53 @@ import config from "../ts/config";
 import { colors, type Position } from "../ts/common";
 import { Duration, Note } from "@scores/lily/music-classes";
 import { barCount as lilyBarCount } from "../ts/lilyData";
-import type { Range } from "@scores/lily/lily";
+import type { LilypondData, NoteEntry, Range } from "@scores/lily/lily";
 MusicCanvas.define("music-canvas");
 
-// Install fixture singing-ranges on a canvas by swapping its whole `lilyData`
-// (since #652 `ranges` is a deeply-readonly SingingRanges, so build a fresh
-// mutable map and replace the object rather than mutating it in place).
+// Install fixture score data by swapping the whole `lilyData`. Since #652 all
+// parsed data lives on one `lilyData` object, so there is no `canvas.notesByQuant`
+// to assign; and `LilypondData`'s fields are readonly (`ranges` deeply so, also
+// #652), so a fixture patches by replacing the object rather than mutating it. The
+// file-level afterEach restores `lilyData` per test, so callers need no cleanup.
+//
+// This is the only place the #706 block installs a fixture, and new code should use
+// it. It is not the file's only writer: the file-level afterEach restores `lilyData`
+// raw, and the older tests above still assign it raw. Those are legacy, not the
+// pattern to copy.
+//
+// `frLocations` is deliberately NOT patchable. MusicCanvas sizes `shimmerPhases`
+// and `falseRelationPulses` from `frLocations.length` once, in #init, and never
+// rebuilds them; patching it here would leave `shimmerPhases` short, so
+// #drawFalseRelationHotspot would read undefined, Math.sin would yield NaN, and
+// the element would emit an hsla() with a NaN alpha that node-canvas silently
+// ignores. `Omit` alone would not prevent it: TypeScript's excess-property check
+// fires only on a fresh object literal, so a patch hoisted into a variable would
+// slip straight through. The `frLocations?: never` intersection rejects it
+// structurally, literal or not, because an FRlocation[] is not assignable to never.
+function setLily(
+  c: MusicCanvas,
+  patch: Partial<Omit<LilypondData, "frLocations">> & { frLocations?: never }
+): void {
+  const base = c.lilyData;
+  // Not an assertion: spreading null is legal JS and yields {}, so a null base
+  // would silently produce a LilypondData missing every unpatched key (barCount
+  // undefined, guards comparing against undefined). Fail loudly instead.
+  if (base == null)
+    throw new Error("setLily: canvas has no lilyData (not initialised)");
+  // An explicitly-undefined value spreads through and would blank a real field
+  // while still typing as LilypondData. Same malformed-object class as the null
+  // base above, so it gets the same loud treatment.
+  for (const [k, v] of Object.entries(patch))
+    if (v === undefined)
+      throw new Error(`setLily: patch key '${k}' is undefined`);
+  c.lilyData = { ...base, ...patch };
+}
+
+// Install fixture singing-ranges. Routes through setLily so there is exactly one
+// place that writes `lilyData`. The arrays are frozen because the real data is:
+// serialise.ts freezes every ranges/notesByQuant array at load (#652), so an
+// unfrozen fixture would let an in-place mutation added to MusicCanvas pass here
+// while throwing in production, which is a false negative rather than a lax test.
 function setRanges(
   c: MusicCanvas,
   build: (r: Map<string, Range[]>) => void
@@ -16,7 +57,8 @@ function setRanges(
   const r = new Map<string, Range[]>();
   for (const [k, v] of c.lilyData!.ranges) r.set(k, [...v]);
   build(r);
-  c.lilyData = { ...c.lilyData!, ranges: r };
+  for (const v of r.values()) Object.freeze(v);
+  setLily(c, { ranges: r });
 }
 
 var canvas: MusicCanvas | null;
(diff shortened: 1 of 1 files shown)


wainwmr (approved):
OK

wainwmr (approved):
OK

wainwmr (approved):
OK

PR #824, opened 2026-07-14
fix: keep the user's feedback when the send fails

### LINKED ISSUE #798: Failed feedback send wipes the typed message while asking for retry
## Type

bug (Area: UI)

## Symptom

A user whose feedback fails to send sees "Couldn't send, please try later"
for 1.5 seconds, then the modal closes and everything they typed is gone.
The message tells them to retry; retrying means retyping the whole message
and re-selecting the rating.

## Steps to reproduce

1. Open the feedback modal (F key or the feedback icon).
2. Type a message; optionally pick a rating.
3. Go offline (or block the POST to `/`), press Send.
4. Watch the error flash, the modal close, and the form reset.

## Expected behaviour

On a failed send the modal stays open with the message and rating intact,
showing the error inline, so the user can retry without retyping.

## Actual behaviour

The failure path shares the success path's teardown: the catch routes into
`showFeedbackResult`, which hides the form and after 1500 ms calls
`showFeedback(false)`, whose `resetFeedbackForm()` runs
`feedbackForm.reset()`. The pending timeout fires even if the user reopens
the modal within the window. The catch also discards the error object, and
the message-less `throw new Error()` on `!res.ok` erases the HTTP status,
so an unregistered Netlify form (permanent 404) is indistinguishable from
a network blip when debugging.

## Affected files

- `packages/pwa/index.ts:428-434` (`showFeedbackResult`: shared teardown)
- `packages/pwa/index.ts:615-643` (submit handler and catch)
- `packages/pwa/index.ts:401-426` (`showFeedback`/`resetFeedbackForm`)
- `packages/pwa/src/test/feedback.test.ts:163-184` (pins only the error text)

## Root cause hypothesis

`showFeedbackResult` was written for the success case (confirm, then tidy
up) and the failure case was routed through it unchanged, inheriting the
auto-close and reset.

## Test code

Extend `feedback.test.ts`: stub fetch to reject, submit with a typed
message, advance timers 1500 ms, assert the textarea still holds the text
and the modal is still open. Fails on current code (the form resets).

## Acceptance criteria

- After a failed send, the modal remains open, the message text and rating
  are unchanged, and the error shows inline (no auto-close on failure).
- The failure is logged with the HTTP status or the caught error.
- Success keeps the current confirm-then-close behaviour.
- A test pins message preservation on failure and fails if the reset returns.

## Risk and scope

One file plus a test extension. The success path is untouched. Low risk.

## Notes

- Found by the 2026-07-13 `packages/pwa/index.html` bug hunt (VERA), finding
  F6; raised by a spawned silent-failure reviewer, byte-verified by the
  hunting agent, cold-verified (defect-verifier: defect, with the
  reopen-within-1500ms wrinkle; defect-weigher: fix/bug, **data-loss floor**,
  destruction certain once the path runs).
- The harm floor (data loss of user-typed work) means this is never accepted
  as a known defect regardless of cost (`doc/CONTRIBUTING.md` § Harm floor).
- Related: the rr `refactor-feedback.test.ts.md` records the declined #752
  cluster (fixture drift); the diagnostics half of this path (discarded
  error, message-less throw) is rr item 15 in `refactor-index.ts.md` and is
  fixed in passing here.

- VERA

wainwright1000:
## The problem

When a feedback send failed, the user saw "Couldn't send, please try later" for
1.5 seconds, then the modal closed and everything they had typed was gone. The
message asked them to retry. Retrying meant retyping the whole message and
re-selecting the rating.

The failure path shared the success path's teardown. The catch routed into
`showFeedbackResult`, which hid the form and, after 1500 ms, called
`showFeedback(false)`, whose `resetFeedbackForm()` ran `feedbackForm.reset()`.
The pending timeout fired even if the user reopened the modal inside that
window.

Losing the user's typed work puts this on the harm floor
(`doc/CONTRIBUTING.md` § Harm floor), so it is fixed rather than accepted.

## What changes

- On a failed send the modal stays open, the message and the rating are
  untouched, and the error shows inline. The user can press Send again without
  retyping a word.
- Success is unchanged: confirm, then close.
- The failure is now logged with its cause, and the thrown error carries the
  HTTP status. An unregistered form (a permanent 404) is no longer
  indistinguishable from a network blip when debugging.
- Success and failure are handled by two sibling handlers rather than
  `then`/`catch`, so an exception raised while rendering the confirmation is no
  longer reported as a network failure.

## Tests

- `src/test/feedback.test.ts` pins the failure path: submit a typed message with
  `fetch` stubbed to reject, advance the timers past 1500 ms, then assert the
  modal is still open and the text is still there. It fails on the old code.
- `e2e/feedback-modal.spec.ts` drives the same path in a real browser.
- The e2e failure is simulated by stubbing `window.fetch` in the page rather
  than by aborting the route. An aborted route makes the browser log its own
  resource error, and allowlisting that needs a port-dependent entry which is
  not portable between worktrees. The reasoning is recorded in
  `doc/TESTING.md`.

Full suite green on this branch: 352 tests across 21 files, plus `check`.

Fixes #798

### DIFF
--- a/doc/TESTING.md
+++ b/doc/TESTING.md
@@ -139,11 +139,27 @@ channels, so without the fixture a test can pass while the page under it is
 broken. `e2e/page-errors.spec.ts` is the fixture's self-test: each positive
 case injects one error class and polls the capture array, and a final
 `test.fail()`-annotated case pins that a non-empty capture fails at teardown.
-If a spec ever legitimately produces a captured message, add a narrow,
-commented entry to the module's `ALLOWLIST` (the current entries cover
-third-party script hosts that are absent or DNS-blocked in test environments)
-rather than reverting the import. Service workers are blocked suite-wide
-(`playwright.config.ts`) so page-level network events stay observable.
+If a spec legitimately produces a captured message, add a narrow, commented entry
+to the module's `ALLOWLIST` rather than reverting the import. The current entries
+cover third-party script hosts that are absent or DNS-blocked in test
+environments, and the feedback submit handler's `console.error` on a failed send
+(#798), which is an expected byproduct of the failure path
+`e2e/feedback-modal.spec.ts` drives; that log is asserted in
+`src/test/feedback.test.ts`, not in the spec. Service workers are blocked
+suite-wide (`playwright.config.ts`) so page-level network events stay observable.
+
+**Simulate a network failure by stubbing `fetch`, not by aborting the route.**
+`route.abort()` makes the browser log its own `Failed to load resource:
+net::ERR_FAILED`, which the fixture captures. Allowlisting it means adding an
+entry for a failed load of the site root `/`, and the fixture's convention is to
+keep an entry as narrow as the URL: that URL is the port-dependent `baseURL`
+(`worktree-ports.ts` gives each worktree its own offset), so a conforming entry
+is not portable, and the broad text-only alternative would mask a genuine failure
+to load the document itself. Reject the call in the page instead
+(`page.addInitScript` replacing `window.fetch`, bound to `window`), which issues
+no request and raises the same `TypeError` a genuinely offline fetch raises.
+`failTheSend` in `e2e/feedback-modal.spec.ts` is the implementation; it points
+here for the reason rather than restating it.
 
 ### E2E Prerequisites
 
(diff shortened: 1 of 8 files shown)


wainwmr (approved):
OK

PR #823, opened 2026-07-14
test(e2e): stub Cloudflare Web Analytics beacon so it can't fail the suite (#822)

wainwmr:
Fixes #822.

## Problem

The nightly **PWA E2E** run on `main` failed **97/123 tests** ([run 29306369458](https://github.com/wainwmr/spem-player/actions/runs/29306369458)) — all from one external cause, not a regression. The app's Cloudflare Web Analytics beacon (`cloudflareinsights.com/cdn-cgi/rum`) is CORS-blocked under the preview server (`localhost:4173`), flooding `console.error`, and the page-error gate ([page-errors.ts](../blob/main/packages/pwa/e2e/helpers/page-errors.ts)) fails any test that emits a non-allowlisted `console.error`. The existing allowlist only covered `static.cloudflareinsights.com` (the script), not the RUM POST endpoint.

## Change

Intercept the whole `cloudflareinsights.com` domain in the auto page-error fixture and **fulfil with an empty 204** — not `abort()`, which would itself log `ERR_FAILED`. The beacon never reaches the network, so it produces no console error and the console-error gate stays strict with **no allowlist entry**. Removes the now-redundant `static.cloudflareinsights.com` allowlist line.

## Verification

- Ran previously-failing specs locally on chromium — **smoke (9), bar-navigation, pwa-update-toast all green**.
- Two `playback-toggle` #634 tests still fail locally, but I confirmed they **also fail on clean `main`** without this change: it's a local headless audio-autoplay quirk (`toBeVisible` on the pause icon), unrelated to the beacon. In CI those tests were only red due to beacon noise, which this removes.

## Acceptance criteria (from #822)

- [x] No E2E test issues a live request to `cloudflareinsights.com` (intercepted).
- [x] Console-error gate stays strict; no new allowlist entry for the beacon (one removed).
- [x] Real analytics still loads in production (test-only interception).
- [ ] Nightly PWA E2E on `main` green — verify on next scheduled run / a manual dispatch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #822, opened 2026-07-14
E2E: block Cloudflare Web Analytics beacon in Playwright so it can't fail the suite


wainwmr:
## Problem

The nightly scheduled **PWA E2E** run on `main` failed: **97 of 123 tests failed** ([run 29306369458](https://github.com/wainwmr/spem-player/actions/runs/29306369458)). All failures share one root cause — not a product regression.

The app loads Cloudflare Web Analytics (the RUM beacon `https://cloudflareinsights.com/cdn-cgi/rum`). Under the Playwright preview server (`http://localhost:4173`) that request is blocked by CORS and fails:

> Access to XMLHttpRequest at 'https://cloudflareinsights.com/cdn-cgi/rum' … blocked by CORS policy
> Failed to load resource: net::ERR_FAILED

The E2E harness `packages/pwa/e2e/helpers/page-errors.ts` fails **any** test that emits a non-allowlisted `console.error`. So every test that happens to trigger the beacon fails — one flaky external dependency multiplied across the whole suite. Preview builds / local runs are affected the same way whenever the beacon misbehaves.

## Fix (option 2, preferred)

Tests should never hit third-party analytics. In `packages/pwa/playwright.config.ts` (or a shared fixture), **route-block the beacon** so it never fires during tests:

```ts
// abort all requests to Cloudflare Insights during E2E
await context.route(/cloudflareinsights\.com/, (route) => route.abort());
```

Cleaner than allowlisting the errors in `page-errors.ts` (option 1), because it removes the network call entirely rather than tolerating its failure output, and keeps the console-error gate strict.

## Acceptance criteria

- [ ] No E2E test issues a request to `cloudflareinsights.com`.
- [ ] `page-errors.ts` console-error gate stays strict (no new allowlist entry for the beacon).
- [ ] Nightly PWA E2E on `main` is green.
- [ ] Real analytics still loads in production (test-only blocking).

## Context

Diagnosed from the failed nightly run; the 97 failures were entirely CORS/`ERR_FAILED` noise from the beacon, no underlying app breakage.

PR #826, opened 2026-07-14
test(e2e): fix allowlist self-test probe broken by beacon removal (#822)

wainwmr:
Follow-up to #823 (both under #822).

## Problem

The reran nightly E2E dropped from **97 failures to 7**, confirming the beacon fix worked. Of the remaining 7:

- **3× `page-errors.spec.ts › does not capture an allowlisted message`** (chromium/firefox/webkit) — a **regression from #823**. That self-test probes the allowlist by emitting `console.error("allowlist probe at https://static.cloudflareinsights.com/beacon.min.js")` and asserting it's suppressed. #823 removed that allowlist entry (the beacon is now intercepted instead), so the probe is captured and the test fails.
- **4× `playback-toggle` on webkit only** — a **separate, pre-existing** webkit audio-autoplay issue (same class as the chromium #634 failures that also fail on clean `main`), unrelated to the beacon. Not addressed here.

## Change

Repoint the self-test probe at `https://www.googletagmanager.com/gtag/js`, which is still in the allowlist, so the suppression mechanism is again tested against a live entry.

## Verification

`playwright test --project=chromium e2e/page-errors.spec.ts` → **7/7 passed** locally (was failing on the allowlist case).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/pwa/e2e/page-errors.spec.ts
+++ b/packages/pwa/e2e/page-errors.spec.ts
@@ -91,7 +91,7 @@ test.describe("page-error capture fixture (self-test)", () => {
     await page.goto("/");
     await page.evaluate(() => {
       console.error(
-        "allowlist probe at https://static.cloudflareinsights.com/beacon.min.js"
+        "allowlist probe at https://www.googletagmanager.com/gtag/js"
       );
       console.error("page-errors self-test: sentinel after allowlisted");
     });


wainwmr (commented):
OK

PR #832, opened 2026-07-14
test(e2e): exempt cancelled /audio/ loads from the page-error gate (#830)

### LINKED ISSUE #830: E2E: webkit playback-toggle fails — page-error gate flags cancelled /audio/ load (Load request cancelled)
## Problem

After the beacon fixes (#823, #826) the nightly PWA E2E is down to **4 failures, webkit-only** ([run 29319404481](https://github.com/wainwmr/spem-player/actions/runs/29319404481) — 135 passed / 4 failed):

- `playback-toggle.spec.ts:4` Space toggles play and pause
- `playback-toggle.spec.ts:32` Enter toggles play and pause
- `playback-toggle.spec.ts:54` focused play/pause button activates on Enter (#634)
- `playback-toggle.spec.ts:77` focused play/pause button activates on Space without scrolling (#634)

All four fail the same way — the page-error gate ([page-errors.ts](../blob/main/packages/pwa/e2e/helpers/page-errors.ts)) captures a **cancelled** audio load:

```
audio response 0 (no content-type): http://localhost:4173/audio/ALC/default.mp3
audio request failed: http://localhost:4173/audio/ALC/default.mp3 (Load request cancelled)
```

## Root cause

The tests toggle play → pause in quick succession. **#805 ("honour control changes during the audio load window", merged after the last green nightly)** now cancels the in-flight audio load when a control change lands during loading. On **webkit** that cancellation surfaces as a `requestfailed` (status 0, "Load request cancelled"); chromium/firefox don't report it. The fixture's `/audio/` channel treats any failed/`<audio content-type` `/audio/` request as a test failure, so webkit goes red.

Cancelling a media load when the user pauses mid-load is **correct behaviour**, not a defect — this is a webkit-specific **false positive** in the strict audio gate exposed by #805, not a user-facing regression.

## Proposed fix

Exempt **cancelled** `/audio/` requests from the audio channel in `page-errors.ts`, the same way 3xx/304 are already exempt. A load cancelled by a deliberate control change (webkit reports `errorText` "Load request cancelled" / status 0) is a legitimate outcome, not a broken audio path.

- In the `requestfailed` handler, skip when `req.failure()?.errorText` indicates cancellation (e.g. matches `/cancel/i`).
- In the `response` handler, skip `status() === 0` (cancelled) responses — they carry no meaningful content-type.

Keep the genuine failure classes (404 / SPA-fallback text-html / decode failure) flagged.

### Alternative considered

Have the specs await the audio load settling before toggling. Rejected: it wouldn't reflect real usage (users do toggle mid-load) and would mask the exact interaction #805 was written to handle.

## Acceptance criteria

- [ ] webkit `playback-toggle` specs pass; no other browser regresses.
- [ ] page-error gate still flags a real missing/404/wrong-content-type `/audio/` response (self-test `page-errors.spec.ts` covers this).
- [ ] Nightly PWA E2E on `main` fully green.

## Context

Confirmed webkit-only and post-#805 via CI history; chromium equivalents pass. Diagnosed during the E2E-stabilisation work (#822).

wainwmr:
Fixes #830. Final step of the E2E stabilisation (after #823, #826) — takes the nightly from 4 failures to green.

## Problem

Since #805 ("honour control changes during the audio load window"), toggling pause mid-load aborts the in-flight audio fetch — correct behaviour. On **webkit** that cancellation surfaces as a `requestfailed` ("Load request cancelled") and a **status-0 response**; the strict `/audio/` channel in [page-errors.ts](../blob/main/packages/pwa/e2e/helpers/page-errors.ts) flagged both, failing all 4 webkit `playback-toggle` tests. chromium/firefox don't report the cancellation.

CI confirmed all 4 failed via the gate (`audio response 0 (no content-type)`), not a product assertion.

## Change

Exempt **cancellations only**:
- `requestfailed`: skip when `errorText` matches `/cancel/i`.
- `response`: skip `status() === 0` (guarded `status() > 0`), like the existing 3xx/304 exemptions.

Genuine failures still record — the self-test's `route.abort()` (ERR_FAILED), DNS errors, 404s, and wrong-content-type `/audio/` responses.

## Verification (local)

- **page-errors self-test — 7/7 on chromium, firefox, and webkit.** The "captures a network-failed /audio/ request" case (deliberate `route.abort()`) still fails red as intended, proving the gate isn't over-loosened.
- webkit `playback-toggle` no longer trips the audio gate.

### Note on #634 tests

`playback-toggle:54/:77` (#634) fail *locally* on all three browsers with a `toBeVisible` timeout on the pause icon — the local headless environment doesn't autoplay audio, so the icon never flips. This is unrelated to this change (and to the beacon work); those tests are green in CI, where audio plays. Their CI failure was the audio-cancel gate, which this PR fixes.

## Acceptance criteria (#830)

- [x] webkit playback-toggle no longer trips the audio gate; no other browser regresses.
- [x] Gate still flags a real missing/404/wrong-content-type/aborted `/audio/` response (self-test green on all browsers).
- [ ] Nightly PWA E2E on `main` fully green — verify on next run / manual dispatch.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/packages/pwa/e2e/helpers/page-errors.ts
+++ b/packages/pwa/e2e/helpers/page-errors.ts
@@ -105,11 +105,18 @@ export const test = base.extend<{ pageErrorCapture: readonly string[] }>({
         }
       });
       page.on("requestfailed", (req) => {
-        if (isAudioUrl(req.url())) {
-          record(
-            `audio request failed: ${req.url()} (${req.failure()?.errorText})`
-          );
-        }
+        if (!isAudioUrl(req.url())) return;
+        const errorText = req.failure()?.errorText ?? "";
+        // A load cancelled by a deliberate control change during the audio
+        // load window (#805) is a legitimate outcome, not a broken audio path:
+        // toggling pause mid-load aborts the in-flight fetch. Webkit surfaces
+        // this as a requestfailed with "Load request cancelled" (and a
+        // status-0 response, exempted below); chromium/firefox do not report
+        // it at all. Exempt cancellations only — genuine failures (the
+        // self-test's route.abort() → ERR_FAILED, DNS errors, decode failures)
+        // still record. See #830.
+        if (/cancel/i.test(errorText)) return;
+        record(`audio request failed: ${req.url()} (${errorText})`);
       });
       page.on("response", (res) => {
         if (!isAudioUrl(res.url())) return;
@@ -120,11 +127,16 @@ export const test = base.extend<{ pageErrorCapture: readonly string[] }>({
         // 304 revalidations legitimately carry no content-type (probe-verified
         // against vite preview, which serves audio with ETag + no-cache) and
         // are exempt — the followed-to or originally-cached response is the
-        // one that gets content-type-checked.
+        // one that gets content-type-checked. A status-0 response is a
+        // cancelled load (webkit emits one when a control change aborts the
+        // in-flight fetch, #805/#830) — a legitimate outcome, exempt like the
+        // matching requestfailed above.
         const contentType = res.headers()["content-type"] ?? "";
         if (
           res.status() >= 400 ||
-          (res.status() < 300 && !contentType.startsWith("audio/"))
+          (res.status() > 0 &&
+            res.status() < 300 &&
+            !contentType.startsWith("audio/"))
         ) {
           record(
             `audio response ${res.status()} (${contentType || "no content-type"}): ${res.url()}`


wainwright1000:
I duplicated this by mistake. I picked up #830 without noticing you already had a PR
open on it, and only found #832 when I went to publish. Your fix stands and I am not
pushing a competing PR. Apologies for the noise.

The work did turn up three things about the exemption that I think are worth having,
and one of them will bite. Evidence below; all of it is yours to take or leave.

## 1. `/cancel/i` misses the cancellation strings the other two engines emit

`/cancel/i` matches WebKit's `Load request cancelled`, which is why it clears the
current four failures. It does not match:

- **chromium** `net::ERR_ABORTED`
- **firefox** `NS_BINDING_ABORTED`
- **firefox** `NS_ERROR_PARSED_DATA_CACHED`

The first two are what a cancelled media load reports on those engines (measured with
a probe that cancelled an in-flight `<audio>` load; both `resourceType: "media"`).

The third is the one that will bite. Firefox reports a load it had **already finished
fetching** when the supersede landed as `NS_ERROR_PARSED_DATA_CACHED`, not as an abort.
I only saw it because I wrote a spec that drives a supersede (press play, then pick a
voice part, which re-points the mp3 URL). It failed on firefox immediately:

```text
audio request failed: http://localhost:4213/audio/ALC/default.mp3 (NS_ERROR_PARSED_DATA_CACHED)
```

No spec on `main` supersedes an audio load today, which is why this has never appeared.
The first one that does will turn firefox red, and the failure will look exactly like
the one you have just fixed.

Suggest matching an exact list rather than a pattern:

```ts
const CANCELLED_LOAD_ERRORS = [
  "net::ERR_ABORTED",            // chromium
  "NS_BINDING_ABORTED",          // firefox, abandoned in flight
  "NS_ERROR_PARSED_DATA_CACHED", // firefox, already fetched when superseded
  "Load request cancelled",      // webkit
];
```

An exact list is also safer than a pattern in the other direction. `/cancel/i` is fine
today, but the near-miss strings are real: `net::ERR_CONNECTION_ABORTED` is a genuine
chromium connection failure, and a `/cancel|abort/i` widening (the obvious next edit
when someone hits the firefox strings above) would swallow it.

## 2. The exemption has no test, so deleting it leaves the suite green

I checked this by mutation on my own copy of the same fix: remove the `requestfailed`
guard, run the whole suite, everything still passes. The self-test cases all drive
error texts that are *not* exempt, so the branch's presence or absence changes nothing
they assert.

Worth pinning, because the failure mode is a silent WebKit nightly flake reappearing
months later. It can be pinned on chromium alone, since the wiring question ("does the
handler consult the exemption?") is engine-independent: `route.abort("aborted")` maps
to `net::ERR_ABORTED` there, so it can impersonate a real cancellation. (It cannot on
WebKit, which reports both `route.abort()` codes as `Blocked by Web Inspector`.) Pair it
with a second URL routed to `route.abort("failed")` that must still be captured, so the
absence of the first is proved to be a filter rather than a late event.

Note the trap that implies: **never use `route.abort("aborted")` to simulate an audio
*failure*** in a spec. On chromium it is indistinguishable from a real cancellation and
would be silently exempt. The no-argument `route.abort()` defaults to `"failed"` and is
safe, which is what the existing self-test relies on.

## 3. Once cancellation is exempt, nothing proves the audio ever loads

This is the one I would most want you to look at. Every playback spec stubs
`HTMLAudioElement.prototype.play`, so the icon assertions pass whether or not a byte of
mp3 was fetched. The `/audio/` channel was therefore the only thing asserting the audio
path worked at all, and exempting cancellation removes that. An app bug that manifests
*as* a cancelled load (the app aborting a fetch it should have kept, so the user presses
play and hears silence) now fails nothing.

The cheap compensating control is one positive spec: press play, assert the media
element actually reaches `readyState >= 2`. I verified it catches what the exemption
blinds: with `/audio/` aborted as `"aborted"` (exempt, so the fixture stays silent), the
fixture passes and that spec goes red.

It has to assert `readyState` rather than a network response, because **WebKit surfaces
no network events at all for media-element loads**. A probe pressed play under WebKit,
reached `readyState` 4, and Playwright saw zero `/audio/` request, response or failure
events, while chromium and firefox both reported a 206. (`fetch()`-issued `/audio/`
requests *are* visible on WebKit, which is how the existing network-failure self-test
works there.) That, rather than a timing race alone, is why none of this reproduces on a
developer machine.

## Where the code is

All of the above is implemented and green on
`andrew/830-exempt-cancelled-audio-loads` (pushed to the fork, not to origin): the exact
list, the fixture-level test, the two positive audio-load cases, and a table test for
the status-0 arm. Take any of it, or none. If it is useful I am happy to open it as a
follow-up PR **after** this one merges, so it rebases onto your fix rather than
competing with it.

## Two unrelated bugs the review turned up

- **#833** (Controls): a user cannot cancel a play while it is still loading.
  `playpause()` and the key handler branch on `playing`, which is false during the load
  window, so a second press calls `play()` again rather than `pause()`. #764 wired
  `#intendsToPlay()` into `setChoir`/`setPart`/`setRecording` but not into the toggle.
  There is a unit test for the behaviour, but it calls `pause()` directly and so bypasses
  the button.
- **#835** (UI): `deriveFeedbackStatus()` checks `networkState` before `paused`, so a
  feedback payload reports `"loading"` while the user was playing or paused.

- VERA

wainwmr (commented):
OK

Issue #828, opened 2026-07-14
FOUC fallback does not cover a hung stylesheet request (neither error nor load fires)


wainwright1000:
## Description

The #801 FOUC-guard fallback reveals the app when the stylesheet **fails**. It does not cover the case where the stylesheet request **hangs** and never settles.

A `<link rel="stylesheet">` whose request is black-holed (captive-portal wifi, a proxy that swallows the asset host, a dead CDN that never sends a reset) fires **neither** `load` **nor** `error` on the link, and it blocks `window.load` for as long as it hangs. So neither of #801's two reveal paths runs, and the page stays blank exactly as it did before #801 — until the browser's own network timeout eventually gives up.

This is not a regression: #801 is strictly better than the prior behaviour here and never worse. It is the residual case #801 knowingly does not cover, filed so it is not forgotten.

## Source

`packages/pwa/index.html` — the inline FOUC-guard fallback script added by #801.

## Raised by

The Vera gate on #801 (code-reviewer and silent-failure-hunter, independently).

## The design question, which is why this is a ticket and not a fix

The obvious remedy is a bounded timer: reveal if still hidden after N seconds. #801 deliberately declined a bare timeout, and the reason is sound — a timer that fires during a *slow but successful* stylesheet load produces exactly the flash of unstyled content the guard exists to prevent.

But the objection weakens with a long bound plus a still-hidden check:

```js
setTimeout(function () {
    if (stillHidden()) reveal('timeout');
}, 10000);
```

On a slow-but-successful load the user sees a brief flash of unstyled content instead of a further blank wait. On a hung load they see the app instead of nothing. Ten seconds of blank page is already a failure by any measure.

**This is a user-visible contract change on slow networks, so it is Andrew's and Mark's call, not one to bank inside a review gate.** The trade is: a possible FOUC flash on the slow-but-successful tail, bought against a bounded worst case on the hung tail.

## Recommended fix

Decide the trade above. If accepted, add the bounded third path with a generous bound and a still-hidden guard, and pin it with an e2e case that holds the CSS request open (never fulfilling the route) and asserts the reveal happens.

## Test plan

Playwright, real browser, computed style (the #801 evidence rule):

- Route the stylesheet to a handler that never calls `fulfill`/`abort`/`continue`, so the request hangs.
- Assert `.viewportDiv` computes `visibility: hidden` before the bound.
- Assert it computes `visible` after the bound.
[...]

PR #834, opened 2026-07-14
fix: reveal the app when the stylesheet fails to load (#801)

### LINKED ISSUE #801: Blank page when the stylesheet fails: add FOUC-guard fallback
## Type

bug (Area: UI)

## Symptom

A first-time visitor whose stylesheet request fails sees a permanently
blank page: no content, no message, no recovery hint, indistinguishable
from the site being down. The HTML arrived and the app JavaScript runs;
only the CSS failed.

## Steps to reproduce

1. First visit (no controlling service worker), block or drop the
   `/assets/*.css` request while allowing the HTML (DevTools request
   blocking reproduces it).
2. Load the site: the page stays blank indefinitely.

## Expected behaviour

If the stylesheet fails, the app either reveals itself unstyled or shows a
plain reload message; it never stays silently blank.

## Actual behaviour

The inline FOUC guard (`index.html:13-18`) sets
`.viewportDiv { visibility: hidden }`, and the only release is
`visibility: visible` at `style.scss:88`, shipped inside the very CSS whose
failure triggers the guard. The injected stylesheet link has no `onerror`,
no timeout fallback, no `noscript`. Verified in a real browser: with the
CSS request aborted, computed visibility stays `hidden` three seconds
after load with the app JS executing and an empty body; the control run
renders normally. Workbox precaches CSS, so repeat visits with an
installed service worker are defended; the exposure is first visits.

## Affected files

- `packages/pwa/index.html:13-18` (the guard; the fallback lands here)
- `packages/pwa/src/scss/style.scss:88` (the sole release)

## Root cause hypothesis

The FOUC guard was added for the styled path only; the failure path of its
release channel was never considered.

## Test code

Playwright: serve the built app with the CSS request aborted; assert
`.viewportDiv` becomes visible (or a fallback message appears) within the
fallback window. Fails on current code (stays hidden; the cold verifier's
probe is the template).

## Acceptance criteria

- With the stylesheet request failing on a first visit, the user sees
  either the unstyled app or an explicit reload message within a few
  seconds; never an indefinitely blank page.
- A slow-but-successful CSS load does not flash unstyled content (prefer
  `link onerror` over a bare timeout; weigher note).
- A real-browser test pins the fallback.

## Risk and scope

A few inline lines in `index.html`. No effect on the styled path. Low
risk; XS-S.

## Notes

- Found by the 2026-07-13 `packages/pwa/index.html` bug hunt (VERA),
  finding F5; raised by a spawned silent-failure reviewer, cold-verified
  by a defect-verifier that ran a real-Chromium probe (CSS aborted:
  computed visibility hidden 3 s after load, empty body; control renders),
  and weighed fix/bug (harm-justified: total silent unusability at
  low-but-real likelihood against a near-zero fix).
- Vite injects the stylesheet link at build; the fallback must live in the
  source `index.html` inline script so it ships in `dist/`.

- VERA

### LINKED ISSUE #829: The FOUC fallback reveals elements the stylesheet is the only thing hiding
## Description

When the #801 FOUC-guard fallback fires, it reveals `.viewportDiv` without the bundled stylesheet. The stylesheet is the only thing hiding three groups of elements, so the fallback screen currently shows:

- the full `#help` panel expanded,
- the full `#feedback-modal` dialog expanded,
- every header tooltip string (`#info .tooltiptext`, `#feedback-trigger .tooltiptext`, etc.) rendered inline down the page.

This only happens when the bundled CSS fails to load, so it is a quality-of-fallback issue rather than a normal-path defect.

## Current state

- `packages/pwa/index.html` (HEAD) has a single inline `<style>` block (lines 13-18) that hides `.viewportDiv` until `style.scss` releases it.
- `packages/pwa/src/scss/style.scss` is the only source that hides:
  - `#help` (around line 210),
  - `#feedback-modal` (around line 579),
  - `header .tooltip .tooltiptext` (around line 187).
- The #801 fix in open PR #834 removes the guard node when the stylesheet fails, which would leave the three elements above visible in normal flow.

## Root cause

The fallback reveals the app by removing the FOUC guard, but it does not replace the CSS rules that were keeping panels and tooltips hidden. Revealing the app and undressing it are coupled because both depend on the same missing stylesheet.

## Scope

- In scope: keep `#help`, `#feedback-modal`, and header tooltips hidden on the fallback screen.
- Explicitly deferred: adding a user-facing "stylesheet failed" banner (ticket candidate 2) is a product call for Mark and is not part of this specification.

## Recommended fix

Add a second, persistent inline `<style>` block in `packages/pwa/index.html` that mirrors the three hiding rules from `style.scss`. The fallback script must remove only the FOUC guard node (`#fouc-guard`), not this new node.

1. In `packages/pwa/index.html`, after the existing FOUC guard `<style>` block, add:

   ```html
   <style id="fallback-hide">
       #help {
           display: none;
       }
       #feedback-modal {
           display: none;
       }
       header .tooltip .tooltiptext {
           visibility: hidden;
       }
   </style>
   ```

2. Give the existing FOUC guard block `id="fouc-guard"` so the fallback script can remove it by id.

3. In `packages/pwa/src/scss/style.scss`, add mirror comments beside the three source rules noting that they are duplicated in `index.html`'s `#fallback-hide` block, so future edits keep both in step.

4. Keep the selectors identical between the two files. Do not broaden `.tooltiptext` beyond `header .tooltip .tooltiptext`; a broader selector would hide tooltips that the header hover rule cannot reveal.

5. Do not merge the two `<style>` blocks into one. Merging them would cause the fallback script to remove both nodes, undoing the hide.

### Implementation note

Open PR #834 (the #801 fix) already contains the exact changes described above, plus tests. The simplest path is therefore to ensure PR #834 also closes #829. If #834 is rejected or split, apply the same diff as a follow-on PR after #801 lands.

## Test plan

Failing-test-first, using Playwright computed-style assertions:

1. On a stylesheet-aborted page:
   - assert `.viewportDiv` computes `visibility: visible` (fallback fired),
   - assert the `#fouc-guard` node is removed,
   - assert `#help` computes `display: none`,
   - assert `#feedback-modal` computes `display: none`,
   - assert a header `.tooltiptext` computes `visibility: hidden`,
   - assert only the expected FOUC diagnostic is logged.

2. On a healthy page:
   - assert `#info` click opens `#help` (`toBeVisible`),
   - assert `#feedback-trigger` click opens `#feedback-modal`,
   - assert hovering a header tooltip reveals its `.tooltiptext`.

3. Add a unit test that reads `index.html` and pins the two-node structure: `#fouc-guard` contains `.viewportDiv`, `#fallback-hide` contains the three selectors, and the fallback script calls `getElementById('fouc-guard')`.

PR #834 already adds `e2e/fouc-fallback.spec.ts` cases for #829 and `src/test/fouc-guard-markup.test.ts` for the unit check. Verify those tests pass on top of the PR branch.

## Dependencies

- Blocked by #801 (native `blockedBy`).
- PR #834 is the open implementation of #801 and already contains the #829 fix; #829 should land with or immediately after #834.

## Open questions

- Mark: do you want a visible "stylesheet failed, reload to try again" banner on the fallback page? The ticket mentions it as a candidate but it is deferred here pending your decision.

## Follow-up items

- None.

## Documentation impact

None expected. The #801 testing notes in `doc/TESTING.md` already cover failure-injection patterns; no new pattern is introduced.

- ZIMI

wainwright1000:
The inline FOUC guard in `index.html` hides `.viewportDiv` until the bundled stylesheet releases it (`style.scss`, `.viewportDiv { visibility: visible }`). A visit whose CSS request failed never got that release, so the page stayed permanently blank.

This reveals the app instead, by removing the guard node.

## Why there are two reveal paths, and why neither can be removed

The fix listens on two things: the stylesheet link's `error` event, and a `window.load` backstop. That looks like belt and braces. It is not, and the reason is measured rather than argued.

For the most likely real failure, a hashed asset that has gone because a deploy landed under an open tab, hosting answers with the index page (200 `text/html`). The engines disagree about which event the link then fires:

- **Chromium** refuses to apply the sheet but fires `load`, so the **backstop** reveals the app.
- **Firefox** fires `error`, so the **fast path** reveals it.
- **WebKit** fires `error`, so the **fast path** reveals it.

So neither path alone covers that failure everywhere: delete the backstop and Chromium stays blank, delete the error path and Firefox and WebKit stay blank.

This is worth stating plainly because reading the HTML link-processing algorithm suggests the opposite (a non-CSS Content-Type fails the sheet, therefore fires `error`, therefore the fast path suffices). Four independent reviewers reached that conclusion during review, and the browser disagreed with all of them. `e2e/fouc-fallback.spec.ts` pins it, and the per-engine table is recorded in the `index.html` comment so the next person to reach the same tidy conclusion is stopped by evidence.

## How the reveal works

`reveal()` removes the guard `<style id="fouc-guard">` node rather than writing an inline style on `.viewportDiv`. That element lives in `<body>` and the script runs in `<head>`, so when a stylesheet error fires it may not exist yet (a pending stylesheet blocks the inline `gtag` script, which blocks parsing before `<body>`). A query for it would silently find nothing and the reveal would be lost. Removing the node is element-independent, order-independent and idempotent, and leaves no inline-style specificity residue.

The `window.load` backstop is armed **first**, before anything that can throw, so a failure in the fast-path setup cannot take the universal path down with it. The fallback logs a `console.error` and marks the document (`data-fouc-fallback`), because an unstyled app that reports nothing is a silent failure and this app collects no errors anywhere else.

## The tests

Six cases, and each pins one thing so a regression cannot hide behind another path:

- an aborted stylesheet reveals via the error path;
- a 404 reveals via the error path (what our host actually returns: the SPA `[[redirects]]` block in `netlify.toml` is commented out);
- a stylesheet served as HTML reveals via whichever path the engine takes (the divergence above);
- a stylesheet that loads but carries no rules reveals via the load backstop;
- a healthy stylesheet releases the guard via the **cascade**, with the fallback not firing;
- the built page arms the MutationObserver, not the initial scan.

Two of those need highlighting.

The healthy-stylesheet case asserts the guard node **survives** and the fallback did **not** run. That re-pins `.viewportDiv { visibility: visible }` in `style.scss`, which reads like a redundant no-op (`visible` is the initial value) and is exactly the kind of line a tidy-up deletes. Before this change, deleting it blanked the page instantly, which is the loudest possible signal. The fallback would otherwise have silently masked that.

The injection cases **count** their route hits and assert the injection actually fired, so a future change to Vite's asset path cannot quietly turn the tests into no-ops that pass while exercising nothing.

## Other changes

`e2e/helpers/page-errors.ts`: the Cloudflare beacon stub moves from `page.route` to `context.route`, so a spec that opens its own page inherits it. The allowlist filter is exported as `isAllowlisted` so an escaping spec filters through the same list rather than a private copy that drifts.

`doc/TESTING.md`: documents when to stub `fetch` versus intercept the route, and how to escape the page-error fixture without going blind (filter the console channel, do not drop it).

## Not covered

A stylesheet request that **hangs** raises neither event until the network stack times out, so the page stays blank in the meantime. That is out of scope here: a bounded timer would fix it but changes the user-visible contract on slow networks, so it is filed as #828 for a decision rather than settled unilaterally. #829 covers the quality of the unstyled fallback page, which reveals `#help` and `#feedback-modal` because `style.scss` is the only thing hiding them.

## Verification

`pnpm run ci` green. Unit: 352 passed. E2E: 157 passed, 0 failed, across Chromium, Firefox and WebKit. Note that `pwa-e2e.yml` runs on a schedule and `workflow_dispatch` only, never on `pull_request`, so the new spec does not gate this PR; it was run locally on all three engines.

Fixes #801
Fixes #829 (the fallback-hide block: the fold agreed 2026-07-14)

### DIFF
--- a/doc/TESTING.md
+++ b/doc/TESTING.md
@@ -148,18 +148,60 @@ environments, and the feedback submit handler's `console.error` on a failed send
 `src/test/feedback.test.ts`, not in the spec. Service workers are blocked
 suite-wide (`playwright.config.ts`) so page-level network events stay observable.
 
-**Simulate a network failure by stubbing `fetch`, not by aborting the route.**
-`route.abort()` makes the browser log its own `Failed to load resource:
-net::ERR_FAILED`, which the fixture captures. Allowlisting it means adding an
-entry for a failed load of the site root `/`, and the fixture's convention is to
-keep an entry as narrow as the URL: that URL is the port-dependent `baseURL`
-(`worktree-ports.ts` gives each worktree its own offset), so a conforming entry
-is not portable, and the broad text-only alternative would mask a genuine failure
-to load the document itself. Reject the call in the page instead
-(`page.addInitScript` replacing `window.fetch`, bound to `window`), which issues
-no request and raises the same `TypeError` a genuinely offline fetch raises.
-`failTheSend` in `e2e/feedback-modal.spec.ts` is the implementation; it points
-here for the reason rather than restating it.
+#### Injecting a failure: which technique, and why they differ
+
+Two specs inject failures and appear to reach opposite conclusions about
+`route.abort()`. They are not in conflict. The question is not whether aborting is
+allowed; it is **what is actually failing**.
+
+**Simulating an app-level network call: stub `fetch`, do not abort the route.**
+Here the *call* is what must fail, and the browser's own resource error is an
+artefact of the injection rather than part of the failure. `route.abort()` makes
+the browser log `Failed to load resource: net::ERR_FAILED`, which the fixture
+captures. Allowlisting it means adding an entry for a failed load of the site root
+`/`, and the fixture's convention is to keep an entry as narrow as the URL: that
+URL is the port-dependent `baseURL` (`worktree-ports.ts` gives each worktree its
+own offset), so a conforming entry is not portable, and the broad text-only
+alternative would mask a genuine failure to load the document itself. Reject the
+call in the page instead (`page.addInitScript` replacing `window.fetch`, bound to
+`window`), which issues no request and raises the same `TypeError` a genuinely
+offline fetch raises. `failTheSend` in `e2e/feedback-modal.spec.ts` is the
+implementation; it points here for the reason rather than restating it.
+
+**Injecting a failed resource load: intercept the route.** Here the *resource* is
+what fails, so there is nothing to stub, and the browser's resource error is part
+of the condition rather than an artefact. Intercept it: `route.abort()` for a
+request that never arrives, or `route.fulfill()` with a bad body for one that
+arrives wrong. `e2e/fouc-fallback.spec.ts` (#801) does both, because a stylesheet
+can fail in either shape and they take different paths through the fix.
+
+**Escape to the spec's own page when the expected errors cannot be allowlisted
+narrowly.** This is a separate decision from the one above, and it is easy to get
+the reason wrong. It is not simply "because the browser logs a resource error": the
+#801 backstop case fulfils a valid 200 and produces no resource error at all, and
+it still needs the escape, because on that path *the app itself* deliberately logs
+a `console.error`. Allowlisting either message suite-wide would blind every other
+spec to a real regression on it. So open the page with `context.newPage()` and keep
+the expected noise off the shared capture.
+
+An escaped page inherits none of the fixture's listeners, nor its `auto` teardown
+assertion, so it is not a free pass. Put a comment at the open site saying why the
+page escapes, and re-attach the channels the spec still needs.
+
+**Filter the console channel; do not drop it.** Dropping it wholesale is the easy
+mistake, and it blinds the one page in the suite whose whole purpose is to exercise
+the failing path. Re-attach `pageerror` and `console`, then assert positively: the
+message the failure is *supposed* to produce must be present, the injection's own
+resource error is allowed by URL, and anything else fails the test. A deliberate
+diagnostic then cannot be deleted without a test going red, and an unrelated error
+cannot hide. `e2e/fouc-fallback.spec.ts` shows the pattern
+(`expectOnlyTheFoucDiagnostic`). Its injection pages do **not** re-attach the two
+`/audio/` listeners, because they never touch audio and every other spec covers
+that channel; drop a channel only when you can say that of it.
+
+The fixture's *routes* (the Cloudflare beacon stub) are registered on the context,
+not the page, so another page in the same context does still inherit those; only
+the listeners are lost.
 
 ### E2E Prerequisites
 
--- a/packages/pwa/e2e/fouc-fallback.spec.ts
+++ b/packages/pwa/e2e/fouc-fallback.spec.ts
@@ -1,0 +1,374 @@
+import { test, expect, isAllowlisted } from "./helpers/page-errors";
+import type { BrowserContext, Page } from "@playwright/test";
+
+// #801 — FOUC-guard fallback.
+//
+// index.html hides .viewportDiv behind an inline <style id="fouc-guard">, and
+// the bundled stylesheet releases it (style.scss, `.viewportDiv { visibility:
+// visible }`). A visit whose CSS request fails never got that release: the page
+// stayed permanently blank. The fix removes the guard node instead, by either of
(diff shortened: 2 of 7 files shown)


wainwmr (approved):
OK

PR #857, opened 2026-07-15
fix: resolve canvas hit-test one bar too low across the right half (#790)

### LINKED ISSUE #790: Canvas hit-test resolves one bar too low across the right half of the overview
## 2. Description

`MusicCanvas` draws the overview as **140** equal bar slices, but hit-tests clicks against `barCount`, which `lily.ts` documents as the **index of the last bar** (139), not a count. The forward and inverse mappings are therefore not inverses of one another: they diverge by a factor of 140/139.

Two user-visible symptoms follow from that one misreading.

**Symptom 1 (the serious one): a click resolves one bar too low across the right half of the overview.** Clicking the visual centre of bar `b` returns `floor((b + 0.5) * 139 / 140)`. That equals `b` for `b <= 69`, and `b - 1` for `70 <= b <= 139`. Drawn bar 139 is unreachable except at a single edge pixel. The playhead is drawn at the centre of the current bar (`MusicCanvas.ts:446`), so hovering the playhead while it sits in bar 100 makes `MusicCanvasWatcher` report **Bar 99**; clicking it seeks a bar early. Verified at 200, 800 and 1400 CSS px; 70 of 140 bars mis-hit in every case.

**Symptom 2: the playhead vanishes and the grid lights up inside the final bar.** `#drawBarHighlight` and `#drawVoiceParts` both test `this.bar > barCount` as "past the end of the piece". `bar` is fractional and is clamped only at `>= 140`, so the whole of bar 139 (`139 < bar < 140`) takes the past-the-end branch. The vertical playhead disappears, and all 40 rows brighten to their bar-0 colouring, as though playback had stopped.

## 3. Reachability

Normal canvas interaction. Symptom 1 is reached by hovering or clicking anywhere in the right-hand half of the overview at any viewport size. Symptom 2 is reached by pressing `Alt+ArrowRight` (the documented 1/16-bar fine seek in `packages/pwa/index.ts:365`) while on bar 139. Playback alone cannot reach it because `getBarFromTime` clamps at `barno[last] === 139` (`packages/pwa/src/ts/common.ts:249`).

## 4. Steps to Reproduce

1. Open the player and start any recording.
2. Hover or click the canvas overview near the visual centre of bar 100 (right-hand half).
3. Observe `MusicCanvasWatcher` report **Bar 99** and, on click, the audio seek to bar 99.
4. Navigate to bar 139 and press `Alt+ArrowRight` once.
5. Observe the vertical playhead disappear and all 40 rows brighten.

## 5. Expected result

- A click at the visual centre of bar `b` returns bar `b` for every `b` in `0..139`.
- Bar 139 is reachable by clicking inside its drawn body.
- The playhead and normal row colouring remain visible throughout bar 139.

## 6. Actual result

- Bars 70 to 139 return one bar too low at their visual centres.
- Bar 139 is reachable only at its extreme right edge.
- Inside bar 139 the playhead vanishes and every row takes the bar-0 "ready" colouring.

## 7. Current state

Verified against `origin/main` (`packages/pwa/src/ts/MusicCanvas.ts`):

- Line 272 (`#calculateCanvasSize`): `this.barWidth = (this.canvas.width - 2 * this.canvasPadding) / 140;` — the forward map uses the literal count 140.
- Line 730 (`#projectToPosition`): `bar: Math.floor(((clampedX - cssPaddingX) * this.lilyData!.barCount) / drawableWidth)` — the inverse map uses `barCount` (139).
- Line 442 (`#drawBarHighlight`): `if (this.bar <= 0 || this.bar > this.lilyData!.barCount) return;` — treats any `bar > 139` as past the end.
- Line 531 (`#drawVoiceParts`): `else if (this.bar === 0 || this.bar > barCount)` — same mis-boundary.
- Line 482 (`#drawSelectionHighlight`): `this.canvasPadding + 140 * this.barWidth - this.barWidth` — another literal 140, currently consistent with the forward map but not robust to a piece with a different bar count.

`packages/pwa/src/ts/lily.ts:341` documents `barCount` as "index of the last bar"; `lily.test.ts:81` pins it to 139.

## 8. Root cause

PR #526 (commit `a8071d9`, the fix for #204) correctly scaled the canvas padding into CSS pixels and, in the same edit, replaced the literal `140` in the inverse map with `this.barCount`. It fixed the padding-scale bug but introduced a count-vs-index bug because `barCount` is the last bar index, not the number of bar slots. The forward map and the end-of-piece tests were left using the literal `140` or the wrong comparison, so the two directions no longer invert.

## 9. Scope

**In scope:**

- Fix the inverse map in `#projectToPosition` so it uses the bar-slot count (140), clamped so the right edge cannot return 140.
- Fix the end-of-piece tests in `#drawBarHighlight` and `#drawVoiceParts` so bar 139 is rendered normally.
- Replace remaining literal `140` expressions in `#calculateCanvasSize` and `#drawSelectionHighlight` with a single named `barSlots` accessor so the count cannot drift from the inverse map again.
- Update the six existing unit tests that currently assert the wrong mapping.
- Add the round-trip and final-bar coverage described in the test plan.

**Out of scope / folded elsewhere:**

- Touch part selection (`part: "all"`, Ticket #327).
- Bottom-edge part mis-mapping (Ticket #791).
- Test-framework inability to catch this class of mapping error (Ticket #706, related but not blocking).

## 10. Recommended fix

Introduce one private accessor that returns the number of bar slots:

```ts
get #barSlots(): number {
  return (this.lilyData?.barCount ?? 139) + 1;
}
```

Then:

1. In `#calculateCanvasSize` (line 272):
   ```ts
   this.barWidth = (this.canvas.width - 2 * this.canvasPadding) / this.#barSlots;
   ```
2. In `#drawSelectionHighlight` (line 482):
   ```ts
   this.canvasPadding + this.#barSlots * this.barWidth - this.barWidth
   ```
   (equivalent to the current value but no longer hard-coded).
3. In `#projectToPosition` (line 730):
   ```ts
   bar: Math.min(
     this.lilyData!.barCount,
     Math.floor(((clampedX - cssPaddingX) * this.#barSlots) / drawableWidth)
   ),
   ```
4. In `#drawBarHighlight` (line 442):
   ```ts
   if (this.bar <= 0 || this.bar >= this.#barSlots) return;
   ```
5. In `#drawVoiceParts` (line 531):
   ```ts
   } else if (this.bar === 0 || this.bar >= barCount + 1) {
   ```
   (or use `this.#barSlots` if the method already has `this.lilyData` available).

This folds `refactor-MusicCanvas.ts.md` Tech Debt item 4 (hard-coded `140` in `#calculateCanvasSize` and `#drawSelectionHighlight`).

## 11. Test plan

Each new or corrected test must be seen failing before the fix.

1. **Hit-test centre.** With a mocked `getBoundingClientRect`, assert that a click at the visual centre of drawn bar 100 returns `bar: 100`. Repeat at widths 200, 800 and 1400. (Returns 99 today.)
2. **Last bar reachable.** Assert that a click inside the drawn body of bar 139, rather than at the extreme edge pixel, returns `bar: 139`. (Returns 138 today.)
3. **Round trip.** For every `b` in `0..139`, assert `xToBar(barToX(b)) === b`. This is the property that makes the class of defect unwritable.
4. **Playhead in the final bar.** With `bar = 139.5`, spy `ctx.moveTo` / `ctx.stroke` and assert the playhead is drawn. (It is not drawn today.)
5. **Colouring in the final bar.** With `bar = 139.5`, assert unselected rows take the dull branch, not the bar-0 branch.
6. **Correct existing assertions.** Update `canvas.test.ts:1061`, `:1204`, `:1748`, `:1759` and the short-viewport click/touch triples so the centre probe expects bar 70 (the bar actually drawn at the horizontal centre), not `floor(barCount / 2)` = 69.

## 12. Dependencies

- Related to #706 (the canvas tests cannot discriminate this class of mapping error). Not blocked by it.

## 13. Open questions

None.

## 14. Follow-up items

None.

## 15. Documentation impact

None expected. `doc/notes/canvas-colours.md` is unrelated; `Architecture-Canvas.md` already describes the coordinate system and will remain accurate once the fix lands.

---

- YASMIN

wainwright1000:
Fixes a hit-test off-by-one in the canvas overview: clicks and hovers across the right half of the overview resolved one bar too low, because the inverse map divided by the literal `140` while the forward map multiplied by `barCount` (139). The two maps now share one slot count, and the forward map is pinned by tests.

## What changed

- `MusicCanvas.ts`: the hit-test inverts through the same bar-slot count the drawing code uses, replacing the hard-coded `140`.
- `canvas.test.ts`: red-first regression tests pinning the forward map and the inverse round-trip at both edges of the bar range.

## Tests

- New regression tests fail on the old code and pass on the fix.
- Full suite green locally: 358 tests, plus the complete check gate (lint, types, format, dependency cruise).

## Note on the version line

This PR bumps the app version to 2.8.16. Open PR #834 bumps the same line to 2.8.15, so whichever of the two merges second will need its usual rebase; the version line is the only file the two share.

Fixes #790

### DIFF
--- a/packages/pwa/package.json
+++ b/packages/pwa/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@spem/pwa",
   "private": true,
-  "version": "2.8.14",
+  "version": "2.8.9",
   "description": "Helping singers learn Thomas Tallis Spem in alium",
   "type": "module",
   "dependencies": {},
--- a/packages/pwa/package.json
+++ b/packages/pwa/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@spem/pwa",
   "private": true,
-  "version": "2.8.9",
+  "version": "2.8.16",
   "description": "Helping singers learn Thomas Tallis Spem in alium",
   "type": "module",
   "dependencies": {},
--- a/packages/pwa/src/test/canvas.test.ts
+++ b/packages/pwa/src/test/canvas.test.ts
@@ -2399,7 +2399,7 @@ describe("MusicCanvas custom element", () => {
       // Branch 3, second disjunct: bar PAST the end of the piece takes the same
       // branch as bar 0. Deleting `|| this.bar > barCount` from the source would
       // send these bars to branch 4 (dull 38 instead of bright 45) and nothing
-      // else in the file would catch it. The pair also pins the STRICT `>`:
+      // else in the file would catch it. The pair also pins the `>=` boundary at 140:
       // barCount itself is still inside the piece and must take branch 4.
       canvas!.bar = barCount + 1;
       col = sole(styleSetsDuring(ctx(), "strokeStyle", () => canvas!.draw()));
@@ -2541,7 +2541,7 @@ describe("MusicCanvas custom element", () => {
           )
         ).toBe(true);
 
-        // bar === barCount: still drawn (the guard is bar > barCount).
+        // bar === barCount: still drawn (the guard is bar >= #barSlots, i.e. 140).
         canvas!.bar = canvas!.lilyData!.barCount;
         moveSpy.mockClear();
         canvas!.draw();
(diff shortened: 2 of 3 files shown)


wainwmr (approved):
OK

PR #858, opened 2026-07-15
fix: alert on a non-clean nightly PWA E2E run (#726)

### LINKED ISSUE #726: Alert on nightly PWA E2E cron failures (the one silent workflow)
### 1. Original title

CI cron jobs (pwa-ci.yml, pwa-e2e.yml) have no failure notification: silent-stop risk

### 2. Description

The nightly `PWA E2E` workflow (`pwa-e2e.yml`, cron-only, 02:00 UTC) gates nothing and notifies nobody. If it starts failing, the only trace is a Playwright report artefact that expires after 7 days and that nobody is prompted to look at. It is the one genuine silent-failure blind spot in the pipeline.

`pwa-ci.yml` is deliberately **out of scope**. It also runs on pull requests, so its failures are already visible at the point they matter. A second alert there would be noise.

Reclassified from `tooling bug` to `feature` (Andrew, 2026-07-09). This adds a new notification capability rather than repairing something broken.

### 3. Current state

`.github/workflows/pwa-e2e.yml` runs the `e2e` job on a nightly cron. It uploads the Playwright HTML report on failure and does nothing else. No notification, no gate.

### 4. Root cause

GitHub Actions scheduled runs notify nobody on failure unless the workflow explicitly sends a signal. The repository relies on artefacts and the Actions UI, neither of which is checked routinely.

### 5. Scope

In scope:

- A companion workflow (`.github/workflows/pwa-e2e-notify.yml`) triggered on `workflow_run` completion of `PWA E2E`, firing only when the conclusion is `failure`.
- A sender script (`packages/monitor/notify-workflow-failure.mjs`) reusing `packages/monitor`'s existing `sendTelegram`.
- Unit coverage in `packages/monitor/monitor-resources.test.mjs`.
- Doc updates to `doc/CI.md` and `doc/MONITOR.md`.

Out of scope:

- `pwa-ci.yml` (already visible via pull-request runs).
- Failure notifications for push/PR-triggered runs.
- Changes to the test commands or artefact uploads.

### 6. Recommended fix (as built)

A companion `workflow_run` workflow rather than an in-workflow notification step, and a Telegram alert rather than a tracking issue.

**Design change from the original spec, recorded here rather than left implicit.** The original spec called for notification steps inside *both* workflows, commenting on a persistent tracking issue, and explicitly ruled Telegram *out* of scope. The delivered branch (`andrew/726-cron-failure-notify`) does the opposite, and the branch's own header comment gives the reason: we already run a Telegram channel and `packages/monitor` already exposes `sendTelegram`, so reusing it is cheaper and lands where someone will actually see it, whereas a tracking issue is one more artefact nobody reads. A companion `workflow_run` workflow also needs no `issues: write` permission on the test workflow and keeps the notification concern out of the test job.

The branch is the better design. The spec was simply never updated to match it, which is the same spec-versus-branch drift that has left #758 stuck.

### 7. Test plan

- `packages/monitor/monitor-resources.test.mjs` covers the sender.
- Confirm the companion workflow fires only when `conclusion == 'failure'`.
- Confirm a successful nightly run sends nothing.

### 8. Dependencies

None.

### 9. Open questions

None.

### 10. Follow-up items

None.

### 11. Documentation impact

`doc/CI.md` and `doc/MONITOR.md` are updated on the branch.

- CLAUDE (title and spec corrected to match the delivered branch; the original spec described work that was not done)

wainwright1000:
The nightly PWA E2E workflow was the one silent workflow: a failed run notified nobody. This adds the alert path so a non-clean nightly run raises a notification, closing the gap where an E2E regression could sit unnoticed until someone happened to look.

## What changed

- `.github/workflows/pwa-e2e.yml` / `pwa-e2e-notify.yml`: failure notification wiring for the nightly run.
- `packages/monitor`: the notify helper and its tests.
- `doc/CI.md` and `doc/MONITOR.md`: updated to describe the alert path.

## Tests

- Full local suite green (363 tests) plus the complete check gate, on a branch rebased onto current main this evening.
- No app source is touched, so no version bump is owed.

Fixes #726


wainwmr (changes requested):
I don't think this is necessary at all.  Adds lots of extra complexity.  I get an email any time that E2E fails, so there is no chance of missing this.

I suggest you close this PR.

wainwright1000:
Fair points both: you do get the failure emails, and 546 diff lines is a real cost for an alert you personally never needed.

One fact that may change the weighing: those scheduled-run failure emails reach only you. GitHub sends them to the workflow's effective actor, so nothing reaches Andrew or the agents on our side; a nightly E2E failure is invisible over here until someone happens to look. That blindness (not your inbox) was the gap #726 was filed against; the ticket title's "silent" meant silent to us.

On the noise concern, if that is part of it: the alert's destination is just the chat id it is handed, so it can go to Andrew alone rather than the group. One value, no code change.

So a possible compromise: keep the alert, point it at Andrew only. If you would still rather not carry the extra surface in the repo at all, we are equally happy to close this and watch the run conclusions from our side over the API instead; the gap closes either way. Your call.

- TELE

PR #863, opened 2026-07-16
docs: correct scores-ci commit message and document the CI-loop guards (#751)

### LINKED ISSUE #751: doc/CI.md wrongly states the scores SVG commit uses [skip ci]; re-trigger guard is the actor check
**Type:** tooling bug
**Area:** Docs
**Difficulty:** XS

### 2. Description

`doc/CI.md` line 61 states the scores-ci SVG commit is made "with message `chore: regenerate SVGs [skip ci]`". That is wrong on two counts: the workflow commits `chore: regenerate SVGs` with no `[skip ci]`, and CI-loop prevention comes not from the tag but from an actor guard on the job. The doc therefore quotes a commit message that is not used and misdescribes the re-trigger mechanism.

The harm is not cosmetic. A contributor who "fixes" the workflow to match the doc would introduce a real defect: Netlify honours `[skip ci]`, so tagging the SVG commit would silently skip the production deploy of regenerated scores, and the site would serve stale sheet music until the next unrelated push to `main`.

### 4. Current state

- `doc/CI.md:61`: "Commit updated SVGs in `packages/pwa/src/scores/` if changes exist, with message `chore: regenerate SVGs [skip ci]`, then push."
- `.github/workflows/scores-ci.yml:62`: `git commit -m "chore: regenerate SVGs"` (no tag).
- `.github/workflows/scores-ci.yml:22`: the actual re-trigger guard, `if: github.actor != 'github-actions[bot]'`.
- `scores-ci.yml` triggers only on `packages/scores/**` (plus its own workflow file), disjoint from the commit's `packages/pwa/src/scores/` path — a second, structural loop guard.
- The commit path sits inside `pwa-ci.yml`'s push path filter (`packages/pwa/**`), so each regeneration triggers a `pwa-ci` verification run (free on this public repository) and a Netlify production deploy (required, to ship the new scores).

All verified against `origin/main` 2026-07-04.

### 5. Root cause

The doc line was written to the common convention (bot commits carry `[skip ci]`, as the monitor workflows' commits genuinely do) rather than to this workflow's actual, deliberate behaviour, and the omission's rationale was never documented anywhere a reader would find it.

### 6. Scope

In scope: the one bullet at `doc/CI.md:61`. Explicitly out of scope: any change to `scores-ci.yml` (its behaviour is correct), and the monitor workflows' `[skip ci]` documentation (accurate as written).

### 7. Recommended fix

Replace the bullet at `doc/CI.md:61` with wording to this effect:

> - Commit updated SVGs in `packages/pwa/src/scores/` if changes exist, with message `chore: regenerate SVGs`, then push. The commit deliberately carries no `[skip ci]`: it must trigger `pwa-ci` to verify the regenerated assets, and it must reach Netlify (which honours `[skip ci]`) so the new scores actually deploy. The workflow cannot re-trigger itself: its path filter (`packages/scores/**`) excludes the commit's path, and the job is guarded by `github.actor != 'github-actions[bot]'`.

Exact prose is the implementer's; the three facts (no tag, why, and the two loop guards) are the content that must land.

### 8. Test plan

Doc correction, so the check is conformance, not code: after editing, confirm the described mechanism matches `scores-ci.yml` (commit message at line 62, actor guard at line 22, `packages/scores/**` trigger paths), and run `npx markdownlint-cli2 doc/CI.md`. No unit tests.

### 9. Dependencies

None blocking. Parked branches for #726 and #728 also edit `doc/CI.md` (different lines, verified); whichever lands second rebases trivially.

### 10. Open questions

None.

### 11. Follow-up items

None.

### 12. Documentation impact

The fix is itself the doc change; no other `doc/*.md` is affected.

---

Original filing by Sauron (2026-07-04 09:46); history: mismatch first found 28 Jun during #727's review wrap-up, re-verified live 4 Jul.

- Tele

wainwright1000:
Corrects doc/CI.md where it wrongly stated the scores SVG commit uses [skip ci]: the actual re-trigger guard is the actor check, and the docs now describe the real CI-loop guards. Documentation only, no code or behaviour change, no version bump owed.

## Tests

- Full local suite green (367 tests) plus the complete check gate, on a branch rebased onto current main today.

Fixes #751

### DIFF
--- a/doc/CI.md
+++ b/doc/CI.md
@@ -171,7 +171,7 @@ Steps:
 - `pnpm run test:lilypond` — run the Lilypond-related test suite.
 - `bash packages/scores/build/install-lilypond.sh` — install LilyPond.
 - Set `PATH` to include LilyPond 2.26.0, then `pnpm run build:scores` — regenerate SVGs.
-- Commit updated SVGs in `packages/pwa/src/scores/` if changes exist, with message `chore: regenerate SVGs [skip ci]`, then push.
+- Commit updated SVGs in `packages/pwa/src/scores/` if changes exist, with message `chore: regenerate SVGs`, then push. The commit deliberately carries no `[skip ci]`: it must trigger `pwa-ci` to verify the regenerated assets, and it must reach Netlify (which honours `[skip ci]`) so the new scores actually deploy. The workflow cannot re-trigger itself: its path filter (`packages/scores/**`) excludes the commit's path, and the job is guarded by `if: github.actor != 'github-actions[bot]'`.
 
 If no SVGs changed, the job exits cleanly without committing.
 


wainwmr (approved):
Ok

